PR #1, opened 2021-10-14
bootstrap project

sergiusens:


cmatsuoka (approved):
No suprises in the basic project template. We can adjust it as we move forward.

PR #5, opened 2021-10-20
oci: add oci image manipulation support (CRAFT-579)

cmatsuoka:
Retrieve a base image from a docker registry, extract it to use as a
craft-parts overlay base image, repack the parts prime content as
a new layer and create an OCI archive file.

-----

### DIFF
--- a/rockcraft/lifecycle.py
+++ b/rockcraft/lifecycle.py
@@ -52,7 +52,7 @@ def pack():
     if not parts_data:
         raise CraftError("No parts specified.")
 
-    work_dir = Path("build").absolute()
+    work_dir = Path("work").absolute()
     image_dir = work_dir / "images"
     bundle_dir = work_dir / "bundles"
 


rockcraft/oci.py
```diff
@@ -0,0 +1,205 @@
...
+
+        image_dir.mkdir(parents=True, exist_ok=True)
+        image_target = image_dir / image_name
+        _copy_image(f"docker://{image_name}", f"oci:{image_target}")
+
+        ui.emit.message(f"Retrieved base {image_name}")
+
+        return cls(image_name=image_name, path=image_dir)
+
+    def copy_to(self, image_name: str, *, image_dir: Path) -> "Image":
```
sergiusens, line 64: This gets rid of no-self-use
```suggestion
    @staticmethod
    def copy_to(self, image_name: str, *, image_dir: Path) -> "Image":
```

Makefile
```diff
@@ -94,7 +94,7 @@ test-pydocstyle:
 .PHONY: test-pylint
 test-pylint:
 	pylint rockcraft
-	pylint tests --disable=missing-module-docstring,missing-function-docstring,redefined-outer-name
+	pylint tests --disable=missing-module-docstring,missing-function-docstring,redefined-outer-name,no-self-use
```
sergiusens, line 97: This should not be needed with my recommendation

rockcraft/lifecycle.py
```diff
@@ -36,23 +37,46 @@ def pack():
     except OSError as err:
         raise CraftError(err) from err
 
+    name = yaml_data.get("name")
+    if not name:
+        emit.message("Project name not specified.")
```
sergiusens, line 42: aren't these errors?

rockcraft/oci.py
```diff
@@ -0,0 +1,205 @@
...
+    @classmethod
+    def from_docker_registry(cls, image_name: str, *, image_dir: Path) -> "Image":
+        """Obtain an image from a docker registry.
+
+        :param image_name: The image to retrieve, in ``name:tag`` format.
+        :param image_dir: The directory to store local OCI images.
+
+        :returns: The downloaded image.
+        """
+        ui.emit.progress(f"Retrieving base {image_name}")
```
sergiusens, line 54: these emits might be better suited on the caller of this method, same for the one below

rockcraft/oci.py
```diff
@@ -0,0 +1,205 @@
...
+
+        ui.emit.progress(f"Exporting {name}:{tag} to local docker daemon")
+
+        src_path = self.path / f"{name}:{tag}"
+        _copy_image(f"oci:{str(src_path)}", f"docker-daemon:{name}:{tag}")
+
+        ui.emit.message(f"Exported {name}:{tag} to local docker daemon")
+
+    def to_oci_archive(self, tag: str) -> None:
+        """Export the current image to the local docker daemon.
```
sergiusens, line 154: this docstring is the same as the one above, what is the difference (can you add that to the doc?)

tools/freeze-requirements.sh
```diff
@@ -7,7 +7,7 @@ requirements_fixups() {
   sed -i '/^python-apt==/d' "$req_file"
 
   echo "git+https://github.com/canonical/craft-cli.git@616673f9312338c841e0da5de6417056aac88aa5#egg=craft_cli" >> "$req_file"
-  echo "git+https://github.com/canonical/craft-parts.git@47ce20b75a7074dff7f64dbda03b62d5740d5b60#egg=craft_parts" >> "$req_file"
+  echo "git+https://github.com/cmatsuoka/craft-parts.git@f0b13a842f97bb344fe4de468dfd55fa1bf5271b#egg=craft_parts" >> "$req_file"
```
sergiusens, line 10: let's get this merged

rockcraft/lifecycle.py
cmatsuoka: Yes. Will change the emit method accordingly.

rockcraft/oci.py
cmatsuoka: A result of the magic of copy and paste. Fixing.

tools/freeze-requirements.sh
cmatsuoka: Will prepare the PRs.

rockcraft/oci.py
cmatsuoka: Moved emits to caller.

rockcraft/lifecycle.py
```diff
@@ -36,23 +36,51 @@ def pack():
...
+    if not base:
+        raise CraftError("Base not specified.")
+
     parts_data = yaml_data.get("parts")
     if not parts_data:
-        emit.message("No parts specified.")
-        return
+        raise CraftError("No parts specified.")
+
+    work_dir = Path("build").absolute()
```
sergiusens, line 55: no change for now, but calling this build can be problematic as build is also a step for a part and has a directory. I cannot think of a better name?

sergiusens (approved):

PR #6, opened 2021-10-21
project: add basic rockcraft.yaml parsing and validation (CRAFT-584)

cmatsuoka:
Add a rockcraft.yaml parser that can also validate parts. This allows
the application to stop before running build providers if the parts
validation fails.

Also renamed the `tag` field to `version`, and changed the output image
file suffix to `.rock instead of `.oci.tar`.

Co-authored-by: Chris Patterson <chris.patterson@canonical.com>

-----

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -149,7 +149,6 @@ disable=print-statement,
         exception-escape,
         comprehension-escape,
         line-too-long,
-        too-few-public-methods,
         fixme,
         fixme-info
 
--- a/rockcraft/project.py
+++ b/rockcraft/project.py
@@ -37,7 +37,7 @@ class Project(pydantic.BaseModel):
     build_base: str
     parts: Dict[str, Any]
 
-    class Config:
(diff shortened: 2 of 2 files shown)


sergiusens (dismissed):

.pylintrc
```diff
@@ -149,6 +149,7 @@ disable=print-statement,
         exception-escape,
         comprehension-escape,
         line-too-long,
+        too-few-public-methods,
```
sergiusens, line 152: can you escape this in the method that needs it instead?

rockcraft/project.py
```diff
@@ -0,0 +1,164 @@
...
+class Project(pydantic.BaseModel):
+    """Rockcraft project definition."""
+
+    name: str
+    version: str
+    base: str
+    build_base: str
+    parts: Dict[str, Any]
+
+    class Config:
```
sergiusens, line 40:
```suggestion
    class Config:  # pylint: disable=R0903
```

PR #18, opened 2022-03-19
providers: fix id mapping in lxd project mount

cmatsuoka:
Map the directory owner when mounting project from host instead of
using the uid of the current user, otherwise file creation will fail in hosts
where the directory owner and user id are different (like in gce spread
tests).

-----


cmatsuoka:
This fixes OCI image generation. Spread tests still fail (is `rockcraft/tools/external/tools` missing?)

sergiusens (approved):
Looks good, going to fix spread in another PR

PR #27, opened 2022-04-13
project: add entrypoint, cmd and env properties

cmatsuoka:
Allow rockcraft projects to specify the entrypoint and cmd of the
generated OCI images, and define the container environment.

-----
CRAFT-811

### DIFF
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -1,0 +1,66 @@
+*********
+Reference
+*********
+
+Rockcraft.yaml format
+=====================
+
+A Rockcraft project is defined in a YAML file named ``rockcraft.yaml``
+at the root of the project tree in the filesystem.
+
+
+Format specification
+--------------------
+
+.. code-block:: yaml
+
+  # The name of the ROCK project.
+  name: <name>
+  
+  # The project version, used as the container tag.
+  version: <version>
+  
+  # The system image and version the application will be layered on.
+  base: ubuntu:18.04 | ubuntu:20.04
+  
+  # (Optional) The system and version on top of which the application
+  # will be built. Defaults to base.
+  build-base: <base>
+  
+  # (Optional) The container entry point.
+  entrypoint: [<path>, ...]
+  
+  # (Optional) The container command line, used as arguments for
+  # entrypoint. If the entrypoint is not defined, the first item in
+  # the cmd list the command to execute.
+  cmd: [<arg>, ...]
+  
+  # (Optional) A list of keys and values defining the container's
+  # runtime environment variables.
+  env:
+    - <var name>: <value>
+  
+  # The parts used to build the application.
+  parts:
+    <part name>:
+      ...
+  
+
+Example
+-------
+
+.. code-block:: yaml
+
+  name: hello 
+  version: latest
+  base: ubuntu:20.04
+  entrypoint: [/usr/bin/hello, -t]
+  env:
+    - VAR1: value
+    - VAR2: "other value"
+  
+  parts:
+    hello:
+      plugin: nil
+      overlay-packages:
+        - hello
--- a/rockcraft/oci.py
+++ b/rockcraft/oci.py
@@ -23,7 +23,7 @@ import subprocess
 import tarfile
 from dataclasses import dataclass
 from pathlib import Path
-from typing import List
+from typing import Dict, List
 
 from craft_cli import CraftError, emit
 
@@ -185,18 +185,23 @@ class Image:
         _config_image(image_path, params)
         emit.message(f"Cmd set to {cmd}", intermediate=True)
 
-    def set_env(self, env: List[str]) -> None:
+    def set_env(self, env: List[Dict[str, str]]) -> None:
         """Set the OCI image environment.
 
-        :param env: A list of environment variables in ``NAME=VALUE`` format.
+        :param env: A list of dictionaries mapping environment variables to
+            their values.
         """
         emit.progress("Configuring env...")
         image_path = self.path / self.image_name
         params = ["--clear=config.env"]
+        env_list = []
         for entry in env:
-            params.extend(["--config.env", entry])
+            for name, value in entry.items():
+                env_item = f"{name}={value}"
+                env_list.append(env_item)
+                params.extend(["--config.env", env_item])
         _config_image(image_path, params)
-        emit.message(f"Environment set to {env}", intermediate=True)
+        emit.message(f"Environment set to {env_list}", intermediate=True)
 
 
 def _copy_image(source: str, destination: str) -> None:
--- a/rockcraft/oci.py
+++ b/rockcraft/oci.py
@@ -16,7 +16,6 @@
 
 """OCI image manipulation helpers."""
 
-
 import logging
 import os
 import shutil
@@ -84,7 +83,7 @@ class Image:
 
         return bundle_path / "rootfs"
 
-    def add_layer(self, tag: str, layer_path: Path) -> None:
+    def add_layer(self, tag: str, layer_path: Path) -> "Image":
         """Add a layer to the image.
 
         :param tag: The tag of the image containing the new layer.
@@ -123,13 +122,15 @@ class Image:
         finally:
             temp_file.unlink(missing_ok=True)
 
+        name = self.image_name.split(":", 1)[0]
+        return self.__class__(image_name=f"{name}:{tag}", path=self.path)
+
     def to_docker_daemon(self, tag: str) -> None:
         """Export the current image to the local docker daemon.
 
         :param tag: The tag to export.
         """
-        parts = self.image_name.split(":", 1)
-        name = parts[0]
+        name = self.image_name.split(":", 1)[0]
         src_path = self.path / f"{name}:{tag}"
         _copy_image(f"oci:{str(src_path)}", f"docker-daemon:{name}:{tag}")
 
@@ -138,8 +139,7 @@ class Image:
 
         :param tag: The tag to export.
         """
-        parts = self.image_name.split(":", 1)
-        name = parts[0]
+        name = self.image_name.split(":", 1)[0]
         src_path = self.path / f"{name}:{tag}"
         _copy_image(f"oci:{str(src_path)}", f"oci-archive:{filename}:{tag}")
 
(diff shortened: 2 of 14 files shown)


tests/spread/general/environment/task.yaml
```diff
@@ -0,0 +1,13 @@
...
+  rockcraft pack
+
+  test -f environment-test_latest.rock
+  test ! -d work
+
+  # test container execution
+  docker images
+  sudo /snap/rockcraft/current/bin/skopeo --insecure-policy copy oci-archive:environment-test_latest.rock docker-daemon:environment-test:latest
+  docker images
+  docker run environment-test bash -c 'echo $X' | grep "ship it!"
```
sergiusens, line 13: if cmd is set to /bin/bash why add the "bash"?

tests/spread/general/entrypoint/rockcraft.yaml
```diff
@@ -0,0 +1,11 @@
+name: entrypoint-test
+version: latest
+base: ubuntu:20.04
+entrypoint: [/usr/bin/hello]
+cmd: [-g, "ship it!"]
```
sergiusens, line 5: is this really used like this? Shouldn't entrypoint be `[/usr/bin/hello,  -g, "ship it!"]`

cmatsuoka: Entrypoint and cmd are a bit confusing, but the idea is that if you define both values then `cmd` act as a complement to `entrypoint`. You can define only one of them, and if `entrypoint` is not defined the first `cmd` entry is the command to execute. `cmd` is also replaced by the command parameter if it's defined in the command-line.

tests/spread/general/environment/task.yaml
cmatsuoka: If a command is defined in the command line it replaces what's defined as the container `cmd`.

tests/spread/general/entrypoint/rockcraft.yaml
cmatsuoka: More information here: https://phoenixnap.com/kb/docker-cmd-vs-entrypoint

tests/spread/general/environment/task.yaml
cmatsuoka: I'll change the default cmd in the test to make the test more meaningful.

cmatsuoka:
Replaced the default command in the environment spread test to something not similar to the command-line override.

tests/spread/general/entrypoint/rockcraft.yaml
sergiusens: I wonder if we should get rid of the entrypoint and cmd commands and do something like we do in snaps where a `daemon: ` converts this to a service. In any case we have not solidified the spec on this.

cmatsuoka: The current entries map directly to the OCI spec and the concept should feel natural to dockerfile authors. We can change that to a different format, but having something that maps to `docker run --entrypoint` and a different thing that can be overriden using `docker run <command>` will probably be convenient/necessary.

cjdcordeiro: This is a good question to raise. Maybe I can shed some extra light on this.

### Meaning

The simplest way to look at it is: 
 - ENTRYPOINT is the image's executable. And even though it can be overridden at runtime, it shows what binary your container is going to run as PID 1 at startup
 - CMD provides default arguments to entrypoint

ENTRYPOINT always comes first, and depending on how they are set, the end result can behave very differently. This matrix summarizes their possible combinations: https://docs.docker.com/engine/reference/builder/#understand-how-cmd-and-entrypoint-interact

### Rule of thumb + examples

 1. "your image runs an executable which is not supposed to accept any arguments, i.e. it does one thing and one things only" -> `cmd: ["/usr/bin/hello", "-g", "ship it!"]`
 2. "your image runs an executable that by default prints "ship it!", but at runtime, users can print whatever they like" -> `entrypoint: ["/usr/bin/hello"]` + `cmd: ["-g", "ship it!"]`. Then at runtime, a user could do `docker run entrypoint-test -g "cool stuff"`, and "ship it" would no longer be printed
 3. "your image runs an executable which needs arguments at runtime, since its execution highly depends on the user configuration and underlying environment" -> `entrypoint: ["/usr/bin/hello"]` (no `cmd`)
 4. "your image runs an executable that always does one thing by default. It has one function. But this function can be extended with additional parameters if the users wants to" -> `entrypoint: ["/usr/bin/ls", "-l"]` (no `cmd`). Then at runtime, users could do something like `docker run entrypoint-test -h -a`
 5. "your image doesn't really have a default executable nor command..." -> not possible. At startup, the container runtime must at least have either an entrypoint or a default command to execute. This is the typical case of OS images like `ubuntu`. A common practice here is to set `cmd: ["bash"]`


 ### Formats

I personally don't like the way this was designed, but unfortunately, the current implementation of ENTRYPOINT and CMD has some entropy when it comes to how they are defined at build time (as you can see from the matrix at https://docs.docker.com/engine/reference/builder/#understand-how-cmd-and-entrypoint-interact). 

We either keep the same syntax and business logic as Dockerfiles (i.e. implicitly infer whether the user wants "exec form" or "shell form" based on whether the values are provided as a List[str] or str, respectively), **or**, we can try simplify this whilst providing the same functionality. I see 2 possible ways:
 1. **explicit declaration**: similar to Dockerfiles, but without any hidden assumptions. It would look something like this in the Dockerfile: 
 ```
 entrypoint:
    value: /usr/bin/hello
    format: exec  
 ```
 
2. **no format**: we'd let users literally write their ENTRYPOINT and CMD as they wish. So for example, if a user wants an entrypoint "/usr/bin/hello" to be executed in a shell, then he/she would have to write it like this in the rockcraft.yaml: `entrypoint: /bin/sh -c '/usr/bin/hello'`

rockcraft/oci.py
```diff
@@ -156,15 +156,62 @@ def digest(self) -> bytes:
...
 
 def _copy_image(source: str, destination: str) -> None:
     """Transfer images from source to destination."""
     _process_run(["skopeo", "--insecure-policy", "copy", source, destination])
 
 
-# XXX: update to use craft-cli output stream
+def _config_image(image_path: Path, params: List[str]) -> None:
+    """Configure the OCI image."""
+    _process_run(["umoci", "config", "--image", str(image_path)] + params)
```
cjdcordeiro, line 209: `umoci` creates new image config and manifest files, leaving the previous ones orphaned, but inside the same `blobs/sha256` directory. Are these being cleanup somewhere else?

rockcraft/oci.py
```diff
@@ -156,15 +156,62 @@ def digest(self) -> bytes:
...
+        """
+        emit.progress("Configuring cmd...")
+        image_path = self.path / self.image_name
+        params = ["--clear=config.cmd"]
+        for entry in cmd:
+            params.extend(["--config.cmd", entry])
+        _config_image(image_path, params)
+        emit.message(f"Cmd set to {cmd}", intermediate=True)
+
+    def set_env(self, env: List[str]) -> None:
```
cjdcordeiro, line 188: could `env` be a `dict` instead? eg.:

```
env:
  FOO: bar
  other_var: 42
```

cjdcordeiro (commented):
awesome. thanks for kicking this ball. 
I've left a couple of comments and a request. 

Hope these help ;)

jnsgruk:
Good to see this! Might be an idea to introduce a lightweight `rockcraft.yaml` spec in the README or docs somewhere, similar to how we do it for Pebble? https://github.com/canonical/pebble#layer-specification

rockcraft/oci.py
cmatsuoka: All the intermediate blobs exist inside the internal working directories, but when the final artifact (the `.rock` file) is exported it only contains the blobs corresponding to the final tag. The working directories are created inside the build container and are not exposed to the user.

Working directory content:
```
 root@rockcraft-hello-1987971:~# ls -l images/hello/blobs/sha256/
total 28168
-rw-r--r-- 1 root root      349 Apr 14 12:42 095ee187a14c5b2f970d09c1c31e3583791aba5290f78dbf587f6bbe6e3405a0
-rw-r--r-- 1 root root      577 Apr 14 12:42 14814774e98aa9a8be826f1155fc6349e2b750809513eeb5a2aeb98c55a88a33
-rw------- 1 root root      732 Apr 14 12:43 21c1b6182858be8dcc70ef3566ee80082147a7ef5224e38d3e78a435fec47c88
-rw------- 1 root root      864 Apr 14 12:43 2f70e21812fc84eb308586b9da197ce3b7dc0074ea460d29f28fc0e3fab54fc1
-rw------- 1 root root      507 Apr 14 12:43 4b14b089a90aa18e15a9cd1d0ce999d858169996d0eb38d7a57e92b25f03e49a
-rw------- 1 root root   239622 Apr 14 12:43 549607a70eeea9c42d2687ef8ababef8d36a5157727b016a364d220ac196ea47
-rw------- 1 root root      507 Apr 14 12:43 814d03fb0110de8d5fc618b0fc55ff7222e6fbe71290556436b9dd22efd245e8
-rw------- 1 root root      507 Apr 14 12:43 8f380bcd899f35a2d885a0639c3dea81f1695da965866e837e66387ecdcdc926
-rw-r--r-- 1 root root 28566292 Apr 14 12:42 e0b25ef516347a097d75f8aea6bc0f42a4e8e70b057e84d85098d51f96d458f9
-rw------- 1 root root      828 Apr 14 12:43 e177c7ea890f8a6eb35b4c9cc1d2491cb7b4f029ec73431ac7131f739254d84b
```

Final artifact:
```
$ tar tvf hello_1.0.rock
drwxr-xr-x root/root         0 2022-04-14 09:43 blobs/
drwxr-xr-x root/root         0 2022-04-14 09:43 blobs/sha256/
-rw-r--r-- root/root       864 2022-04-14 09:43 blobs/sha256/2f70e21812fc84eb308586b9da197ce3b7dc0074ea460d29f28fc0e3fab54fc1
-rw-r--r-- root/root    239622 2022-04-14 09:43 blobs/sha256/549607a70eeea9c42d2687ef8ababef8d36a5157727b016a364d220ac196ea47
-rw-r--r-- root/root       507 2022-04-14 09:43 blobs/sha256/814d03fb0110de8d5fc618b0fc55ff7222e6fbe71290556436b9dd22efd245e8
-rw-r--r-- root/root  28566292 2022-04-14 09:43 blobs/sha256/e0b25ef516347a097d75f8aea6bc0f42a4e8e70b057e84d85098d51f96d458f9
-rw-r--r-- root/root       244 2022-04-14 09:43 index.json
-rw-r--r-- root/root        31 2022-04-14 09:43 oci-layout
```

rockcraft/oci.py
cmatsuoka: We could express it as a dict with a few caveats: the YAML spec doesn't guarantee order (but our implementation will preserve order in Python 3.7+), and as a dictionary it won't allow constructs such as:
```
env:
   FOO: bar
   FOO: $FOO:baz
```
(unsure if docker allows this type of construct however).

This was addressed in Snapcraft by using a list of single-entry dictionaries, such as:
```
env:
  - FOO: bar
  - FOO: $FOO:baz
```
Using a similar construct in rockcraft would also make it consistent with `build-environment` in parts, which is a good thing IMHO.

cmatsuoka:
Will add a yaml spec to the documentation.

tests/spread/general/entrypoint/rockcraft.yaml
sergiusens: Wouldn't this clean things up a bit from an experience point of view?
```yaml
run:  # optional
    type: [entrypoint|cmd]  # defaults to entrypoint
    command: [<cmd>...]
    args: [<args>...]  # optional
```

I used `run` to avoid use of `command` or `entrypoint` as the main keyword to avoid confusion of intent.

rockcraft/oci.py
cjdcordeiro: a list is fine too

cmatsuoka:
Added some initial format documentation, which should be improved as specifications mature and new features are added.

rockcraft/oci.py
cmatsuoka: Changed to use the same format used in snapcraft (list of dictionaries).

sergiusens: We probably want the duplicate keys detection we added to the YAML loader for Snapcraft for this too.

sergiusens:
I am merging this on the premise the the presentation layer can change in the upcoming weeks and that this is enough to get things tested more.

docs/reference.rst
```diff
@@ -0,0 +1,66 @@
...
+*********
+
+Rockcraft.yaml format
+=====================
+
+A Rockcraft project is defined in a YAML file named ``rockcraft.yaml``
+at the root of the project tree in the filesystem.
+
+
+Format specification
```
sergiusens, line 12: We can probably leverage the pydantic model here for some autodocumentation of the mode.

sergiusens (approved):

PR #31, opened 2022-04-29
project: kickstart population of metadata

cjdcordeiro:
Sets annotations prior to exporting the OCI image, based on the rockcraft.yaml values and the build ISO date.
 
-----


cjdcordeiro:
@cmatsuoka @sergiusens this is ready for review

rockcraft/oci.py
```diff
@@ -203,6 +183,58 @@ def set_env(self, env: List[Dict[str, str]]) -> None:
         _config_image(image_path, params)
         emit.message(f"Environment set to {env_list}", intermediate=True)
 
+    def set_control_data(self, metadata: Dict[str, Any]) -> None:
+        """Create and populate the ROCK's control data folder.
+
+        :param metadata: content for the ROCK's metadata YAML file
+        :type metadata: Dict[str, Any]
+        """
+        control_data_destination = "/.rock"
```
sergiusens, line 192: is this mostly to not show the temp path? I would argue that the emit.progress' where this is used would probably go through so fast they would not be read. In any case I would remove this variable and make progress not print out /.rock and instead emit.progress("Writing ROCK metadata") and the closing message is probably a verbose or debug one.

tests/unit/test_oci.py
```diff
@@ -228,3 +228,109 @@ def test_set_env(self, mocker):
                 universal_newlines=True,
             )
         ]
+
+    @patch("subprocess.run")
+    @patch("tempfile.mkdtemp")
+    @patch("pathlib.Path.mkdir")
+    @patch("shutil.rmtree")
+    @patch("rockcraft.oci._compress_layer")
```
sergiusens, line 236: can you use the mocker fixture instead of these?

something like https://github.com/canonical/rockcraft/pull/31/files#diff-43d03eb32b3769225898764d036e6e7fad2649ce712b0d45135d084631789acfL29 ?

sergiusens: the process_run fixture is probably satisfying the need to patch subprocess.run too

rockcraft/oci.py
cjdcordeiro: I have a few doubts about your comment:

> is this mostly to not show the temp path? 

Which temp path? `/.rock` is the final path inside the ROCK. 
Are you suggesting we don't print it? (just trying to understand the problem)

cjdcordeiro: got it. This was leftover code, cause in the past I was using `umoci insert` instead of `umoci raw add-layer`. Fixed in https://github.com/canonical/rockcraft/pull/31/commits/83945867836ac918ddf148de47b63b34d577a8e2

tests/unit/test_oci.py
cjdcordeiro: done in https://github.com/canonical/rockcraft/pull/31/commits/5f58b87e3733ec797b8db80c2adeb3c5f2bdb92a

rockcraft/oci.py
```diff
@@ -203,6 +183,57 @@ def set_env(self, env: List[Dict[str, str]]) -> None:
...
+        emit.progress("Setting the ROCK's Control Data")
+        local_control_data_path = Path(tempfile.mkdtemp())
+
+        # the ROCK control data structure starts with the folder ".rock"
+        control_data_rock_folder = local_control_data_path / Path(".rock")
+        control_data_rock_folder.mkdir(parents=True)
+
+        rock_metadata_file = control_data_rock_folder / Path("metadata.yaml")
+        with open(rock_metadata_file, "w", encoding="utf-8") as rock_meta:
+            yaml.dump(metadata, rock_meta)
```
sergiusens, line 201: We shouldn't need parents in the mkdir as mkdtemp would already create the path up to that point (we should consider cleaning up mkdtemp, I can look into that in a follow up PR)

A string appended to a path, is a Path; we can also use pathlib.Path's open method instead

```suggestion
        # the ROCK control data structure starts with the folder ".rock"
        control_data_rock_folder = local_control_data_path / ".rock"
        control_data_rock_folder.mkdir()

        rock_metadata_file = control_data_rock_folder / "metadata.yaml"
        with rock_metadata_file.open("w", encoding="utf-8") as rock_meta:
            yaml.dump(metadata, rock_meta)
```

sergiusens (approved):
looks good, one comment

PR #36, opened 2022-05-18
project: support build-base and allow 'bare' bases

cjdcordeiro:
Adds support for the `build-base` property in the rockcraft.yaml input file, allowing to have a ROCK base (including `bare`) which can be different from the environment used to build the ROCK's content

`ubuntu:22.04` is also supported now. 

-----


cjdcordeiro:
ping @valentincanonical

valentincanonical:
The 'bare' base feature introduces the bug described in https://github.com/canonical/rockcraft/issues/39

cjdcordeiro:
> The 'bare' base feature introduces the bug described in #39

fixed in https://github.com/canonical/rockcraft/pull/36/commits/d1a5f04cd7ad477e8d5e056228360b97afcea28b

cjdcordeiro:
@cmatsuoka @sergiusens this is ready for review

tests/spread/general/bare-base/task.yaml
```diff
@@ -0,0 +1,13 @@
+summary: bare base build test
+
+execute: |
```
sergiusens, line 3: there is no restore here which would cleanup everything.

Is there a general docker command we could run that would remove everything? We could run that in a more general restore (we seem to be running out of space). While I would prefer proper cleanups, we can request larger images with the "storage" key in spread.yaml, example of a different project for reference:

```yaml
    systems:
      - ubuntu-18.04-64:
          workers: 18
          image: ubuntu-1804-64
      - ubuntu-20.04-64:
          workers: 6
          image: ubuntu-2004-64
          storage: 40G
      - ubuntu-22.04-64:
          workers: 6
          image: ubuntu-2204-64
```

rockcraft/lifecycle.py
```diff
@@ -49,7 +49,10 @@ def pack():
     # Obtain base image and extract it to use as our overlay base
     # TODO: check if image was already downloaded, etc.
     emit.progress(f"Retrieving base {project.base}")
-    base_image = oci.Image.from_docker_registry(project.base, image_dir=image_dir)
+    if project.base == "bare":
+        base_image = oci.Image.new_oci_image(image_dir=image_dir)
```
sergiusens, line 53: to keep oci.Image, I would say project.base should also be passed in here, is tagging required in the called to logic?

tests/spread/general/bare-base/task.yaml
cjdcordeiro: we could use `docker image prune -a` but this is a bit aggressive. if these tests are run on a user's env, they'll cleanup all the images on that host.

the alternative is to cleanup per task, just the things we've created -> in this case, the .rock file + respective docker image.

wdyt?

rockcraft/lifecycle.py
cjdcordeiro: yes I think the tag plays a role, cause certain `umoci` calls do expect it. what do u propose?

rockcraft/lifecycle.py
```diff
@@ -50,7 +50,10 @@ def pack():
...
```
sergiusens, line 54:
```suggestion
        base_image = oci.Image.new_oci_image(f"{project.base}:latest}", image_dir=image_dir)
```

sergiusens (approved):

Issue #23, opened 2022-04-08
Provide a simple, declarative way to provision users in the resulting OCI image


jnsgruk:
Right now, one has to do something like the following:

```yaml
parts:
  non-root-user:
    plugin: nil
    overlay-script: |
      chroot $CRAFT_OVERLAY
      useradd -M -r my-user
```

Such a feature would likely need to provide some sort of language for describing UID, system user or not, home directory if required, shell, etc. An MVP might just be the ability to create system users with a given UID, since that is a very common use case.

jnsgruk:
This is now essentially complete with the introduction of the daemon user :)

PR #40, opened 2022-05-20
lifecycle: define new and default users for the ROCK

cjdcordeiro:
Adds the properties `users` and `default_user` to rockcraft.yaml, allowing for the build-time creation of new users, as well as for the configuration of the default runtime user for the container image.

-----
fixes https://github.com/canonical/rockcraft/issues/23

### DIFF
--- a/rockcraft/errors.py
+++ b/rockcraft/errors.py
@@ -1,0 +1,44 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright 2022 Canonical Ltd.
+#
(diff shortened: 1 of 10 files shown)


cjdcordeiro:
ping @valentincanonical

gruyaume:
Would this feature work on the `bare` base as well?

cjdcordeiro:
> Would this feature work on the `bare` base as well?

this PR will most likely be dropped. The most recent discussions have us looking into setting the default OCI user, only, from a well-known list of users. The actual setting of private users will be postponed.

sergiusens:
I am closing this for now to avoid confusion

PR #46, opened 2022-07-15
Issues/45/add base metadata

### LINKED ISSUE #45: fix: missing base information in the .rock metadata file
https://github.com/canonical/rockcraft/blob/1209dd9a14d3fb43e2b2931ab27a0e7c8388a6b9/rockcraft/project.py#L124

we're missing both the `base-digest` (and respective annotation) and `base`

cjdcordeiro:
From the project base, takes the base name and digest, and adds:
 - the base name to the ROCK metadata file
 - the base digest to both the ROCK metadata file and labels&annotations

**note:** for `bare` bases, the digest is being taken into consideration, even though in these cases, the base is created from nothing...should digest be `None` in these cases?

fixes #45 
-----


sergiusens (approved):

cjdcordeiro:
@sergiusens I had to do a few changes to the original commit due to `skopeo --preserver-digests` not working as I expected. Should be ok now.

also pls take note of my remark about
> note: for bare bases, the digest is being taken into consideration, even though in these cases, the base is created from nothing...should digest be None in these cases?

atm, a digest is always being taken. If that makes sense to you, then it should be ok to merge

PR #47, opened 2022-07-22
simplify confirm_with_user

SamirPS:

-----
Remove the cast of str in the input() on confirm with user and simplify the if


SamirPS:
@cjdcordeiro

sergiusens:
Hi @SamirPS, as I mentioned in your other PR, this PR has issues which can be seen on the CI run output https://github.com/canonical/rockcraft/runs/7610878501?check_suite_focus=true

SamirPS:
> Hi @SamirPS, as I mentioned in your other PR, this PR has issues which can be seen on the CI run output https://github.com/canonical/rockcraft/runs/7610878501?check_suite_focus=true

What command I need to launch to reformat automatically ?

cjdcordeiro:
yes, it is enough for you to run `black rockcraft/utils.py` to fix that file's syntax.

if you want to run the liniting tests locally, before pushing a new change, you can have a look at https://github.com/canonical/rockcraft/blob/main/Makefile#L59

The CI is not running all of those though, so if you just want to make the CI pass, you can run these ones: https://github.com/canonical/rockcraft/blob/main/.github/workflows/tests.yaml#L24

SamirPS:
> yes, it is enough for you to run `black rockcraft/utils.py` to fix that file's syntax.
> 
> if you want to run the liniting tests locally, before pushing a new change, you can have a look at https://github.com/canonical/rockcraft/blob/main/Makefile#L59
> 
> The CI is not running all of those though, so if you just want to make the CI pass, you can run these ones: https://github.com/canonical/rockcraft/blob/main/.github/workflows/tests.yaml#L24

Thanks i add a commit to pass the test

SamirPS:
Hello,

i don't understand the error from the Spread / snap-tests (pull_request)

SamirPS:
I just commit a fix to pass the macos test

rockcraft/utils.py
```diff
@@ -77,11 +77,7 @@ def confirm_with_user(prompt, default=False) -> bool:
...
-    if reply and reply[0] == "y":
-        return True
+    reply = input(prompt + choices).lower().strip()
+    possible_answers = {" y ": True, " n ": False}
 
-    if reply and reply[0] == "n":
-        return False
-
-    return default
+    return possible_answers.get(reply[0].lower().strip(), default) if reply else default
```
sergiusens, line 83: why .lower here and on line 80? same for strip

SamirPS: Lower : Because of the commentary of the function "True if answer starts with [yY], False if answer starts with [nN],
        otherwise the default." so i lower to get only have the key y or n in my dict
Strip : i will change it it's because of black i need to add space in the get

cjdcordeiro (changes requested):
what about:

```python
reply = input(prompt + choices).lower().strip()
return reply[0] == "y" if reply else default
```

cjdcordeiro:
> Normally it's work because if reply[0] is "n" or other thing it's will return false . For me it's good too

then pls update, and let's wait for @sergiusens for give the green light

SamirPS:
> > Normally it's work because if reply[0] is "n" or other thing it's will return false . For me it's good too
> 
> then pls update, and let's wait for @sergiusens for give the green light

Okay, i changed the file

sergiusens (approved):

PR #51, opened 2022-08-16
cleanup: don't write test files in cwd

tigarmo:
Use "tmp_path" to isolate test data and keep the result of "make
test-units" clean.

I want to do some more cleaning-up but I figured I'd get this 'easy' fix out of the way

### DIFF
--- a/tests/unit/test_project.py
+++ b/tests/unit/test_project.py
@@ -17,7 +17,6 @@
 import datetime
 import json
 import textwrap
-from pathlib import Path
 
 import pytest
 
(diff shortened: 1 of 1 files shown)


tests/unit/test_project.py
```diff
@@ -164,13 +163,14 @@ def test_project_parts_validation(yaml_data):
...
-    Path("rockcraft.yaml").write_text(
+def test_project_load(yaml_data, tmp_path):
+    rockcraft_file = tmp_path / "rockcraft.yaml"
+    rockcraft_file.write_text(
         textwrap.dedent(json.dumps(yaml_data)),
         encoding="utf-8",
     )
 
-    project = load_project("rockcraft.yaml")
+    project = load_project(str(rockcraft_file))
```
cmatsuoka, line 173: We can refactor `load_project` later to accept a Path argument.

cmatsuoka (approved):
Good catch!

tests/unit/test_project.py
tigarmo: Yeap it's a good idea, I think the function already works with `Paths` because it just forwards the argument to `open`, but I thought I'd try to minimize the number of touched files on this first PR

PR #52, opened 2022-08-16
feat: add support for the "platforms" primitive

### LINKED ISSUE #49: support architecture-specific builds
The goal is to support the syntax for a Rockstar to specify a list of “build” and “target” architectures, which would, in turn, result in one ROCK artefact per combination.

All of these ROCK artefacts combined would, later on, compose the multi-arch ROCK (each artefact is an OCI manifest).

---

## Spec

```yaml
# takes a map of labels. Each label produces one ROCK artefact
# dict[str, dict[str, obj]] (req)
platforms:
  # labels are free text, used as a default for build-on/build-for when needed
  # labels are included in the final ROCK artefact name, eg. my-rock_1.0_<label>.rock.
  # There must be at least one label within platforms
  amd64:
    # will build-on amd64, and build-for amd64
  foo:
    # same as the previous label "amd64" above, just a different ROCK artefact
    
    # the build environment architecture where Rockcraft CAN build the ROCK artefact
    # list (req if `build-for` is present): compatible with $GOARCH, 
    build-on: [amd64]
    # the target architecture(s) the ROCK will be built for.
    # both `base` and `build-base` must respect the provided architecture
    # list (opt): one of $GOARCH
    build-for: amd64
  arm:
    # `build-for` defaults to the valid architecture defined by the label "arm"
    build-on: [amd64, arm, arm64] 
  i386: 
    # ERROR: `build-on` must be present 
    build-for: i386   
  another-label:
    # build-for i386
    build-on: [amd64]
    build-for: i386:
  amd64:
    # ERROR: the label-defined arch differs from the one defined by `build-for`
    build-on: [amd64]
    build-for: arm64
```

cjdcordeiro:
Fixes #49 

Implements the multi-arch support syntax detailed in #49 .

-----

### DIFF
--- a/rockcraft/lifecycle.py
+++ b/rockcraft/lifecycle.py
@@ -39,6 +39,18 @@ def pack():
         pack_in_provider(project)
         return
 
+    for platform_entry, platform in project.platforms.items():
+        build_for = (
+            platform["build_for"][0] if platform.get("build_for") else platform_entry
+        )
+        build_for_variant = platform.get("build_for_variant")
+        pack_for_platform(build_for, build_for_variant, project, managed_mode)
+
+
+def pack_for_platform(
+    build_for: str, build_for_variant: str, project: Project, managed_mode: bool
+) -> None:
+    """Packs a ROCK for a given target architecture."""
     if managed_mode:
         work_dir = utils.get_managed_environment_home_path()
     else:
@@ -52,11 +64,14 @@ def pack():
     emit.progress(f"Retrieving base {project.base}")
     if project.base == "bare":
         base_image, source_image = oci.Image.new_oci_image(
-            f"{project.base}:latest", image_dir=image_dir
+            f"{project.base}:latest",
+            image_dir=image_dir,
+            arch=build_for,
+            variant=build_for_variant,
         )
     else:
         base_image, source_image = oci.Image.from_docker_registry(
-            project.base, image_dir=image_dir
+            project.base, image_dir=image_dir, arch=build_for, variant=build_for_variant
         )
     emit.message(f"Retrieved base {project.base}", intermediate=True)
 
@@ -102,12 +117,16 @@ def pack():
     oci_annotations, rock_metadata = project.generate_metadata(
         packing_time, base_digest
     )
+    rock_metadata["architecture"] = build_for
+    # TODO: add variant to rock_metadata too
+    # if build_for_variant:
+    #     rock_metadata["variant"] = build_for_variant
     new_image.set_annotations(oci_annotations)
     new_image.set_control_data(rock_metadata)
     emit.progress("Metadata added")
 
     emit.progress("Exporting to OCI archive")
-    archive_name = f"{project.name}_{project.version}.rock"
+    archive_name = f"{project.name}_{project.version}_{build_for}.rock"
     new_image.to_oci_archive(tag=project.version, filename=archive_name)
     emit.message(f"Exported to OCI archive '{archive_name}'", intermediate=True)
 
--- a/rockcraft/oci.py
+++ b/rockcraft/oci.py
@@ -16,6 +16,8 @@
 
 """OCI image manipulation helpers."""
 
+import hashlib
+import json
 import logging
 import os
 import shutil
@@ -45,12 +47,15 @@ class Image:
 
     @classmethod
     def from_docker_registry(
-        cls, image_name: str, *, image_dir: Path
+        cls, image_name: str, *, image_dir: Path, arch: str, variant: str = None
     ) -> Tuple["Image", str]:
         """Obtain an image from a docker registry.
 
         :param image_name: The image to retrieve, in ``name:tag`` format.
         :param image_dir: The directory to store local OCI images.
+        :param arch: The architecture of the Docker image to fetch.
+        :param variant: The variant, if any, of the Docker image to fetch.
+
 
         :returns: The downloaded image and it's corresponding source image
         """
@@ -58,16 +63,23 @@ class Image:
         image_target = image_dir / image_name
 
         source_image = f"docker://{image_name}"
-        _copy_image(source_image, f"oci:{image_target}")
+        platform_params = ["--override-arch", arch]
+        if variant:
+            platform_params += ["--override-variant", variant]
+        _copy_image(source_image, f"oci:{image_target}", *platform_params)
 
         return cls(image_name=image_name, path=image_dir), source_image
 
     @classmethod
-    def new_oci_image(cls, image_name: str, image_dir: Path) -> Tuple["Image", str]:
+    def new_oci_image(
+        cls, image_name: str, image_dir: Path, arch: str, variant: str
+    ) -> Tuple["Image", str]:
         """Create a new OCI image out of thin air.
 
         :param image_name: The image to initiate, in ``name:tag`` format.
         :param image_dir: The directory to store the local OCI image.
+        :param arch: The architecture of the OCI image to create.
+        :param variant: The variant, if any, of the OCI image to create.
 
         :returns: The new image object and it's corresponding source image
         """
@@ -77,6 +89,13 @@ class Image:
         shutil.rmtree(image_target_no_tag, ignore_errors=True)
         _process_run(["umoci", "init", "--layout", image_target_no_tag])
         _process_run(["umoci", "new", "--image", str(image_target)])
+
+        # Unfortunately, umoci does not allow initializing an image
+        # with arch and variant. We can configure the arch via
+        # umoci config, but not the variant. Need to do it manually
+        _config_image(image_target, ["--architecture", arch, "--no-history"])
+        if variant:
+            _inject_architecture_variant(Path(image_target_no_tag), variant)
 
         # for new OCI images, the source image corresponds to the newly generated image
         return cls(image_name=image_name, path=image_dir), f"oci:{str(image_target)}"
@@ -260,12 +279,15 @@ class Image:
         emit.message(f"Labels and annotations set to {labels_list}", intermediate=True)
 
 
-def _copy_image(source: str, destination: str) -> None:
+def _copy_image(source: str, destination: str, *system_params) -> None:
     """Transfer images from source to destination."""
     _process_run(
         [
             "skopeo",
             "--insecure-policy",
+        ]
+        + list(system_params)
+        + [
             "copy",
             source,
             destination,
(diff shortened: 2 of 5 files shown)


cjdcordeiro:
@sergiusens @cmatsuoka -> PR is ready for review

rockcraft/lifecycle.py
```diff
@@ -39,6 +39,24 @@ def pack():
         pack_in_provider(project)
         return
 
+    for platform_entry, platform in project.platforms.items():
```
tigarmo, line 42: Hi! What happens if `project.platforms` is empty? Or more specifically: is this new primitive mandatory?

rockcraft/oci.py
```diff
@@ -45,29 +47,39 @@ class Image:
...
-    def new_oci_image(cls, image_name: str, image_dir: Path) -> Tuple["Image", str]:
+    def new_oci_image(
+        cls, image_name: str, image_dir: Path, arch: str, variant: str
+    ) -> Tuple["Image", str]:
         """Create a new OCI image out of thin air.
 
         :param image_name: The image to initiate, in ``name:tag`` format.
         :param image_dir: The directory to store the local OCI image.
+        :param arch: The architecture of the OCI image to create.
+        :param variant: The variant, if any, of the OCI image to create.
```
tigarmo, line 82: > `The variant, if any,`

So should this be an `Optional[str]`? Might be a bit more ergonomic than having to explicitly pass `""`

rockcraft/project.py
```diff
@@ -34,6 +37,105 @@ class ProjectValidationError(CraftError):
...
+        """Vectorise target architecture if needed."""
+        if isinstance(val, str):
+            val = [val]
+        return val
+
+    @pydantic.root_validator(skip_on_failure=True)
+    @classmethod
+    def _validate_platform_set(cls, values):
+        """Validate the build_on build_for combination."""
+        build_for = values["build_for"] if values.get("build_for") else []
```
tigarmo, line 117: (minor)
```suggestion
        build_for = values.get("build_for", [])
```
Same for `build_on`

rockcraft/project.py
```diff
@@ -34,6 +37,105 @@ class ProjectValidationError(CraftError):
...
+                    f"Trying to build a ROCK for {build_for} "
+                    "but multiple target architectures are not "
+                    "currently supported. Please specify only 1 value."
+                )
+            )
+
+        # If build_for is provided, then build_on must also be
+        if not build_on and build_for:
+            raise ProjectValidationError(
+                "'build_for' expects 'build_on' to also be provided."
```
tigarmo, line 133: (minor) Maybe add the platform name to the message? To facilitate yaml debugging
I see that there's a similar idea in line 216, but I don't think this failure will get that prefix

tigarmo (commented):
Thanks for adding me to the review, happy to be here :)

rockcraft/project.py
cjdcordeiro: unfortunately this won't work, because the YAML parser will serialize the values into a JSON like this: `{"build_for": None}`, so `values.get("build_for",[])` will in fact return `None` instead of []

rockcraft/lifecycle.py
cjdcordeiro: hey!

indeed it is mandatory. it cannot be empty. The project expected a Dict to be passed (https://github.com/canonical/rockcraft/pull/52/files#diff-0ea7d1a44d76c7792a852505490ca9e3fb738ec2a7814b051ebc828efb8ceceeR148)

rockcraft/oci.py
cjdcordeiro: yep good catch
https://github.com/canonical/rockcraft/pull/52/commits/9811b0d8ced59a6b80dc451e549cb50b2c76f2fe

rockcraft/project.py
cjdcordeiro: The `Platform` class doesn't have access to the platform name. 
But I've instead wrapped each platform validation such that I can prefix that platform name when it fails: https://github.com/canonical/rockcraft/pull/52/commits/7808f17ca989f5f67f98c33731b5b80feadffbd2

rockcraft/project.py
tigarmo: Yes you are right!

rockcraft/lifecycle.py
tigarmo: If it's mandatory, can you please the tutorial in `doc/tutorials.rst`? It's short but I believe both the `rockcraft.yaml` sample and the archive name are incorrect now.

Either that or we should add a task to update the tutorial soon-ish (I would also be fine with this)

docs/reference.rst
```diff
@@ -53,6 +53,18 @@ Format specification
   # runtime environment variables.
   env:
     - <var name>: <value>
+
+  # List of architecture-specific ROCKs to be built.
+  # Supported architectures are: amd64, arm64, arm, i386, ppc64le, riscv64 and s390x.
+  platforms:
+    # If matching a valid architecture name, it must be the same as "build-for".
+    <entry>:
```
tigarmo, line 61: I think I'm confused on the definition of the entities here: can you please clarify?
- `Supported architectures` refers to the `build-on` and `build-for` values, right? That is, the `<arch>` entries;
- If this is true, what exactly *is* each `<entry>` platform? Can you tell me a use-case for a platform that would not match a valid architecture name?

rockcraft/project.py
```diff
@@ -122,7 +331,7 @@ def unmarshal(cls, data: Dict[str, Any]) -> "Project":
         :raise TypeError: If data is not a dictionary.
         """
         if not isinstance(data, dict):
-            raise TypeError("part data is not a dictionary")
+            raise TypeError("project data is not a dictionary")
```
tigarmo, line 334: :star2:

rockcraft/lifecycle.py
cjdcordeiro: y missed that one.
fixed now: https://github.com/canonical/rockcraft/pull/52/commits/58db8095731dd627a27224180a84afe77fae27b1

thanks for spotting it.

This should actually be automated. The doc's tutorial should fetch the YAML from https://github.com/canonical/rockcraft/pull/52/files#diff-9acc5f73b478b10cf158341e7a1ba82274b3912ed6f9e040fd04372acf75abdc, which is actually being tested by the CI. Otherwise, we start having duplicated and untested content in the tutorials. This is a good issue to have in the backlog: https://github.com/canonical/rockcraft/issues/54

docs/reference.rst
cjdcordeiro: yes, you're right.

The extra bit of information you can add to your reasoning is: "if <entry> matches a `supported architecture`, then its name is taken into account for the business logic.

here's an  example: https://github.com/canonical/rockcraft/pull/52/files#diff-31c6bafe300f15ab74a13b7445a1e801360baa32ce1b99c0c025c1edb27bd67f

tests/unit/test_project.py
```diff
@@ -76,139 +112,278 @@ def test_project_unmarshal(yaml_data):
...
     assert str(err.value) == (
         "Bad rockcraft.yaml content:\n"
         "- extra field 'invalid' not permitted in 'parts.foo' configuration"
     )
 
 
-def test_project_load(yaml_data, tmp_path):
-    rockcraft_file = tmp_path / "rockcraft.yaml"
+def test_project_load(yaml_data, yaml_loaded_data):
+    rockcraft_file = Path("rockcraft.yaml")
```
tigarmo, line 328: Hmm why did you revert this? Did we overwrite each other?
The reason I added the `tmp_path` is that the test was dumping the yaml file on `cwd`, which was visible when e.g. running via `make test-units`.

tests/unit/test_project.py
```diff
@@ -76,139 +112,278 @@ def test_project_unmarshal(yaml_data):
...
+                for platform in project.__getattribute__(attr).values()
+            )
+            assert all(
+                "build_for" in platform
+                for platform in project.__getattribute__(attr).values()
+            )
+            continue
         assert project.__getattribute__(attr) == v
 
+    rockcraft_file.unlink()
```
tigarmo, line 355: I think we should either wrap the whole test in a `try/finally` to ensure this cleanup, or maybe use `request.addfinalizer()` 
(or with the `tmp_path` fixture I guess we wouldn't have to worry about this test file).

tigarmo (changes requested):

tests/unit/test_project.py
cjdcordeiro: this should have been a merge conflict...I'd need to dig deeper into the Git history to understand why there was no conflict

cjdcordeiro: re-added

tests/unit/test_project.py
cjdcordeiro: `tmp_path` is back in there again

tigarmo (approved):

tigarmo:
LGTM, should probably get a more seasoned pair of eyes on this before merging

Makefile
```diff
@@ -57,7 +57,7 @@ install: clean ## Install python package.
 	python setup.py install
 
 .PHONY: lint
-lint: test-black test-codespell test-flake8 test-isort test-mypy test-pydocstyle test-pylint test-pyright ## Run all linting tests.
+lint: test-black test-codespell test-flake8 test-isort test-mypy test-pydocstyle test-pyright test-pylint ## Run all linting tests.
```
sergiusens, line 60: is there a reason for changing this?

tigarmo: Possibly when @cjdcordeiro was working on this the `pylint` step was broken (I think I fixed it in the meantime)

cjdcordeiro: yes. the reason is to be consistent with the CI.

the `linters` job (https://github.com/canonical/rockcraft/blob/main/.github/workflows/tests.yaml#L10) isn't (yet) calling for pylint. Locally, I like to run `make tests` before pushing new changes to ensure the CI will pass. With the previous order, if pylint complains, `make tests` will fail before getting to `pyright`, so you won't get an accurate prediction of the CI (unless you then manually run the `pyright` by hand. 

Another alternative would be to have a linting pipeline that doesn't fail-fast, always going through all tests, so that you can run all linters in one go, and get all issues that need fixing

sergiusens (approved):

cjdcordeiro:
@sergiusens fixed conflicts, merged and passing tests (ready to merge)

PR #53, opened 2022-08-17
cleanup: Clean pylint config and refactor a bit

tigarmo:
Modifications so that `make lint` finishes successfully. This moves the pylint config to `pyproject.toml` and does some minor refactoring to clear the linting errors that are worth supporting (such as duplicate-code).


tests/unit/providers/__init__.py
```diff
@@ -0,0 +1,55 @@
+from abc import ABC, abstractmethod
+
+import pytest
+
+from rockcraft.providers import Provider
+
+
+class BaseProviderTest(ABC):
```
tigarmo, line 8: This is the "major" change in the PR: I wanted to reused those two tests in both `test_lxd` and `test_multipass` without getting too "magical" with the fixtures. Eager to hear if there's a cleaner way to do this, though.

sergiusens: If this is controversial or not it does not matter as much as this code is moving into craft-providers; correct me if I am wrong @mr-cal

pyproject.toml
```diff
@@ -5,3 +5,19 @@ force_grid_wrap = 0
...
+
+[tool.pylint.format]
+max-line-length = "88"
+max-locals = 16
+
+[tool.pylint.MASTER]
+extension-pkg-whitelist = [
+    "pydantic"
+]
+load-plugins = "pylint_pytest"
```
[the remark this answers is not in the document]
tigarmo: :+1:

sergiusens (approved):

PR #59, opened 2022-08-24
Install pebble via implicit part

cjdcordeiro:
Since Pebble will be the entrypoint for all ROCKs, we must have it installed for all ROCKs too.

Similar to what Kerncraft is doing at https://github.com/canonical/kerncraft/pull/22/files#diff-a5aa2c8752af7f6a4943ee8761dd913a6a7662e7d5e7b4c80602e225070617d2, also in Rockcraft we must add a part for building and installing Pebble into the ROCK.

The difference between this PR and the Kerncraft implementation is that 1) the Pebble part is implicit (always there), and 2) there's no need for a custom Pebble plugin, since Pebble is just a Go application, so we can just use the already existing Go plugin from craft-cli.

-----


cjdcordeiro:
@sergiusens @cmatsuoka @tigarmo pls note the comment in https://github.com/canonical/rockcraft/pull/59/files#diff-0ea7d1a44d76c7792a852505490ca9e3fb738ec2a7814b051ebc828efb8ceceeR108 (it isn't a serious problem since we have unit tests which would flag an invalid part spec, but still, maybe there is a better place for this)

rockcraft/project.py
```diff
@@ -101,6 +101,29 @@ def _validate_build_base(cls, build_base, values):
...
+        # since the user hasn't specified such part.
+        # Is there another logical place where this insertion could happen?
+        pebble_part_spec = {
+            "pebble": {
+                "plugin": "go",
+                "source": "https://github.com/canonical/pebble.git",
+                "build-environment": [
+                    {"CGO_ENABLED": "0"},
+                ],
+                "build-snaps": ["go"],
```
sergiusens, line 120: lock to a channel for this not to unexpectedly break

rockcraft/project.py
```diff
@@ -101,6 +101,29 @@ def _validate_build_base(cls, build_base, values):
             build_base = values.get("base")
         return build_base
 
+    @pydantic.validator("parts", pre=True)
+    @classmethod
+    def _add_pebble_part(cls, parts):
```
sergiusens, line 106: I am conflicted with this, on one side it is nice to see that a user could do something like:

```yaml
parts:
    pebble:
       source-branch: experimental-branch
```

and for the right thing to happen. This is what we do in Snapcraft which adds an external part from something defined elsewhere https://github.com/snapcore/snapcraft/blob/main/snapcraft/parts/lifecycle.py#L105 ... a similar strategy could be used here where if pebble is not defined we add it (it should not fail validation at this point as we control the input).

@cmatsuoka what do you prefer?

tests/unit/test_project.py
```diff
@@ -68,10 +83,13 @@ def test_project_unmarshal(yaml_data):
     assert project.cmd == ["world"]
     assert project.env == [{"NAME": "VALUE"}]
     assert project.parts == {
-        "foo": {
-            "plugin": "nil",
-            "overlay-script": "ls",
-        }
+        **pebble_part,
```
sergiusens, line 86: might be good to see a test where we override one value only too, what happens if a user declares "plugin: make"?

rockcraft/project.py
cjdcordeiro: I thought about this.
The problem here is that we must be very strict about how Pebble gets into the ROCK, since it is our entrypoint. At most, I do agree it would be nice to allow users to pinpoint a certain Pebble version, but I think this requires some additional discussions to consider:
 - what is the "special" part name that we allow users to define, such that we take that into account when injecting our implicit Pebble part?
 - do we actually need to do this via a part? For now, the only thing users should have a saying about, is what version of Pebble to have installed...isn't a "part" an overkill for such a simple setting?

this sounds like something that requires a spec, cause we would be introducing a naming convention for a special part, that would be taken as an input for our implicit Pebble part. I suggest moving this to a different PR, after we've aligned on the right approach

tests/unit/test_project.py
cjdcordeiro: I'm not sure I understand. can you elaborate pls? 

> what happens if a user declares "plugin: make"?

from which part?

rockcraft/project.py
cjdcordeiro: good idea. done

rockcraft/project.py
tigarmo: Out of curiosity: do we expect users to want to override/configure the ROCK's pebble? Isn't that similar to, I don't know, configuring the base core22 somehow?

cjdcordeiro: I think that's the question we need to tackle. My first instinct would be to say that users could, at most, have a say about the Pebble version that is being installed in their ROCK. 

But we must enforce other things like:
 - how and where Pebble is installed
 - the OCI pebble entrypoint
 - the Pebble env variables
(and maybe more)

sergiusens: I believe these latter restrictions should come in the form of linters as the system is flexible enough to allow you to

```
override-prime: |
    craftctl prime
    mv|rm <pebble-path> ...
```

cjdcordeiro: yep agreed. thus one more reason why such flexibility should probably be addressed separately, in addition to this PR, as an improvement. resolving this thread in favor of https://github.com/canonical/rockcraft/issues/65

cjdcordeiro:
the CI error could be an intermittent one. could you please re-run the workflow?

cjdcordeiro:
I'm not sure what the error is about. I can't reproduce it locally. I've run `spread` and also directly `rockcraft`, and couldn't reproduce.

tigarmo:
Weird, it failed again:
```
subprocess.CalledProcessError: Command '['snap', 'install', 'go', '--channel', '1.19/stable', '--classic']' returned non-zero exit status 1.
```

tigarmo:
I think the other tests use an older version of go; maybe we should use latest/stable?

cjdcordeiro:
oh but now the error was different. So the go installation failed due to lack of disk space

tigarmo:
Very curious, these machines have 40G of storage I think. @sergiusens any insight into why this would fail?

tigarmo:
Hi @cjdcordeiro, I updated the PR with a few changes:
- Merged `main`, fixed a bunch of conflicts (nothing of note there);
- Added two commits: [one updating the pebble part](https://github.com/canonical/rockcraft/pull/59/commits/13cb8d66b730821e8683abd3267791d9b6217d91) because of issue #102, and [one refactoring the code a bit](https://github.com/canonical/rockcraft/pull/59/commits/99ce1d7edbd4fca4a127fa6cdf1f488bfeeff24d) to support a pre-existing, user-set pebble part. I also moved the code to a separate `_add_pebble_data()` function to have an place to look at when we do further pebble changes.

Please take a look at the changes, and if you +1 them I'll go ahead and approve/merge the PR. Thanks!

tests/unit/test_project.py
tigarmo: After some (offline) discussion with @sergiusens, the latest version of the code supports a pre-existing (user-provided) "pebble" part. If the part exists, we assume it's complete (so no trying to "merge" the two parts together). I think that's an OK compromise for now, and since we can revisit this in the future after pebble+rockcraft has matured a bit I'll resolve this. Thanks!

cjdcordeiro:
Thanks @tigarmo ! Nice work.

It all lgtm except https://github.com/canonical/rockcraft/pull/59/files#diff-0ea7d1a44d76c7792a852505490ca9e3fb738ec2a7814b051ebc828efb8ceceeR493. This is not according to our spec, because it allows users to create a ROCK without Pebble as the entrypoint. 

I understand that the objective is to align with https://github.com/canonical/rockcraft/issues/65, but we don't need to do this now. My fear is that we are introducing a capability which later on must be removed ("being able of not having pebble").

cjdcordeiro:
> > This is not according to our spec, because it allows users to create a ROCK without Pebble as the entrypoint.
> 
> How would the users create a ROCK without Pebble as the entrypoint? Do you mean if they have a part called `pebble` that actually has unrelated things?

exactly. something like

```
parts:
  pebble:
    plugin: nil
```

Which could be there either by mistake or by an ill-intended desire to avoid Pebble

cjdcordeiro:
perfect

tigarmo (approved):

Issue #44, opened 2022-07-12
File permissions/ownership not preserved


jnsgruk:
Hey :wave: 

I've been building a ROCK for Traefik. I want to include a config file and have it owned by a specific user. Take the following excerpt as an example (full file [here](https://github.com/jnsgruk/traefik-rock/blob/2.8.1/rockcraft.yaml)):

```yaml
  default-config:
    plugin: dump
    source: files
    organize:
      traefik.yaml: etc/traefik/traefik.yaml
      001-default.yaml: var/lib/pebble/default/layers/001-default.yaml
    stage:
      - etc/traefik/traefik.yaml
      - var/lib/pebble/default/layers/001-default.yaml
    override-prime: |
      craftctl default
      ls -l etc/
      chown -R 1000:1000 etc/traefik
      ls -l etc/

  non-root-user:
    plugin: nil
    after: [default-config]
    overlay-script: |
      # Create a user in the $CRAFT_OVERLAY chroot
      useradd -R $CRAFT_OVERLAY -M -r -u 1000 -U traefik
```

I've added the `ls -l` as an example - the first shows the following:

```# bash
# ...
2022-07-12 07:50:57.423 :: -rw-r----- 3 root shadow 523 Jul 12 07:48 shadow
2022-07-12 07:50:57.423 :: -rw-r----- 3 root shadow 501 May 31 15:43 shadow-
2022-07-12 07:50:57.423 :: drwxr-xr-x 1 root root    24 Jul 12 07:50 traefik
```

The second then shows that the `chown` is working as expected:

```bash
# ...
2022-07-12 07:50:57.426 :: -rw-r----- 3 root shadow 523 Jul 12 07:48 shadow
2022-07-12 07:50:57.426 :: -rw-r----- 3 root shadow 501 May 31 15:43 shadow-
2022-07-12 07:50:57.426 :: drwxr-xr-x 1 1000   1000  24 Jul 12 07:50 traefik
```

However, when the image is built, if I get a shell and list the directory:

```bash
❯ docker run --rm --entrypoint pebble --entrypoint /bin/bash -it traefik:2.8.1
root@ea2f942c09f4:/# ls -l /etc | grep trae
drw------- 1 root root      24 Jul 12 07:49 traefik
root@ea2f942c09f4:/#
```

I also attempted to `chown` the directory in the `non-root-user` stage after I'd created the user (within the chroot), but it can't see the `/etc/traefik` directory created in the previous stage from within the `chroot`.

[...]

PR #62, opened 2022-08-25
spread: add a "big" spread test

tigarmo:
This new test can be used to "house" multiple different checks without
having to "setup" and "teardown" the whole machinery multiple times.
This initial version tests bug #44, regarding permissions of added
directories.

-----

@sergiusens @cmatsuoka this is an attempt at adding a spread test that we can augment over time, as discussed last week. Thanksss

### DIFF
--- a/tests/spread/general/big/rockcraft.yaml
+++ b/tests/spread/general/big/rockcraft.yaml
@@ -1,0 +1,24 @@
+name: big
+version: latest
+summary: A big ROCK to test many features
+description: |
+  A big ROCK whose purpose is to test many features while only paying the "setup"
+  and "teardown" price once. Feel free to add to this file and to "task.yaml", 
+  adding references to issues/PRs where appropriate.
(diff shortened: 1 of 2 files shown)


tests/spread/general/big/rockcraft.yaml
```diff
@@ -0,0 +1,24 @@
...
+    organize:
+      a.txt: etc/newfiles/a.txt
+      b.txt: etc/newfiles/b.txt
+    stage:
+      - etc/newfiles/a.txt
+      - etc/newfiles/b.txt
+    override-prime: |
+      craftctl default
+      chown -R 9999:9999 etc/newfiles
+      chown 3333:3333 etc/newfiles/b.txt
```
sergiusens, line 24: My only comment :-)

```suggestion
      chown 3333:3333 etc/newfiles/b.txt

```

sergiusens (approved):

PR #64, opened 2022-09-02
commands: add lifecycle step commands

cmatsuoka:
Implement rockcraft commands for the parts lifecycle steps pull,
overlay, build, stage, and prime. Originally implemented to debug
problems related to OCI whiteout files in part layers.

-----

### DIFF
--- a/rockcraft/commands/lifecycle.py
+++ b/rockcraft/commands/lifecycle.py
@@ -58,7 +58,7 @@ class _LifecycleStepCommand(_LifecycleCommand):
 
 
 class PullCommand(_LifecycleStepCommand):
-    """The "pull" command."""
+    """Command to pull parts."""
 
     name = "pull"
     help_msg = "Download or retrieve artifacts defined for a part"
@@ -72,7 +72,7 @@ class PullCommand(_LifecycleStepCommand):
 
 
 class OverlayCommand(_LifecycleStepCommand):
-    """The "overlay" command."""
+    """Command to overlay parts."""
 
     name = "overlay"
     help_msg = "Create part layers over the base filesystem."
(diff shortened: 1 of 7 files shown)


tigarmo (approved):
thanks!

rockcraft/commands/lifecycle.py
```diff
@@ -18,16 +18,112 @@
...
+        """
+        Download or retrieve artifacts defined for a part. If part names
+        are specified only those parts will be pulled, otherwise all parts
+        will be pulled.
+        """
+    )
+
+
+class OverlayCommand(_LifecycleStepCommand):
+    """Run the lifecycle up to the build step."""
```
sergiusens, line 75:
```suggestion
    """Run the lifecycle up to the overlay step."""
```

rockcraft/commands/lifecycle.py
```diff
@@ -18,16 +18,112 @@
...
+        """Run the command."""
+        if not self.name:
+            raise RuntimeError("command name not specified")
+
+        emit.trace(f"lifecycle command: {self.name!r}, arguments: {parsed_args!r}")
+        lifecycle.run(self.name, parsed_args)
+
+
+class _LifecycleStepCommand(_LifecycleCommand):
+    """Run lifecycle step commands."""
```
sergiusens, line 46: nit, but the class does not _Run_, it is the command itself, right?

rockcraft/commands/lifecycle.py
cmatsuoka: Copy-and-paste leftover fixed.

rockcraft/commands/lifecycle.py
cmatsuoka: Indeed, removed the "run" word.

PR #68, opened 2022-09-06
commands: add shell and shell after parameters

cmatsuoka:
Implement parameters `--shell` and `--shell-after` for lifecycle step
commands to shell into the build instance during the parts lifecycle
execution.

-----


tests/unit/test_parts.py
```diff
@@ -64,6 +66,88 @@ def test_parts_lifecycle_run(new_dir):
     assert Path(lifecycle.prime_dir, "foo.txt").is_file()
 
 
+@pytest.mark.parametrize("step_name", ["pull", "overlay", "build", "stage", "prime"])
```
sergiusens, line 69: something like this avoids the if block below
```suggestion
@pytest.mark.parametrize("step_name,last_step", [("pull", None), ("overlay",steps.PULL), ("build", steps.OVERLAY), ("stage", steps.BUILD), ("prime", steps.STAGE)])
```

sergiusens (commented):
looks good, but a minor improvement to the tests is suggested

tests/unit/test_parts.py
cmatsuoka: Much better, will also change this in the original test in Snapcraft.

sergiusens (approved):

PR #70, opened 2022-09-09
parts: use part validator from craft-parts

cmatsuoka:
Use the common part validator implemented in craft-parts to validate part data against part and plugin models, and drop the local implementation.

-----

### DIFF
--- a/rockcraft/parts.py
+++ b/rockcraft/parts.py
@@ -22,8 +22,7 @@ from typing import Any, Dict, List, Optional
 
 import craft_parts
 from craft_cli import emit
-from craft_parts import ActionType, Step, plugins
-from craft_parts.parts import PartSpec
+from craft_parts import ActionType, Step
 from xdg import BaseDirectory  # type: ignore
 
 from rockcraft.errors import PartsLifecycleError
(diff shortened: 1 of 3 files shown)


rockcraft/parts.py
```diff
@@ -191,21 +190,4 @@ def validate_part(data: Dict[str, Any]) -> None:
...
-
-    plugin_class = plugins.get_plugin_class(plugin_name)
-
-    # validate plugin properties
-    plugin_class.properties_class.unmarshal(spec)
-
-    # validate common part properties
-    part_spec = plugins.extract_part_properties(spec, plugin_name=plugin_name)
-    PartSpec(**part_spec)
+    craft_parts.validate_part(data)
```
sergiusens, line 193: are you doing it this way to provide some form of abstraction?

tests/unit/test_project.py
```diff
@@ -327,14 +327,12 @@ def test_project_load(yaml_data, yaml_loaded_data, tmp_path):
             attr = "rock_license"
         if attr == "platforms":
             # platforms get mutated at validation time
-            assert project.__getattribute__(attr).keys() == v.keys()
+            assert getattr(project, attr).keys() == v.keys()
```
sergiusens, line 330: can this be separate?

rockcraft/parts.py
cmatsuoka: I'm accessing craft-parts only from the parts module. This tries to mirror snapcraft which accesses craft-parts only in the parts package.

tests/unit/test_project.py
cmatsuoka: Ah, this shouldn't be in this PR at all, it was done locally to be able to run linters. Removing.

cmatsuoka:
Rebased on main.

sergiusens (approved):

PR #71, opened 2022-09-12
providers: install rockcraft snap via craft-providers API

mr-cal:

-----

Use the new `craft-providers` API to install snaps inside a provider.  No functional changes.

(CRAFT-1347)


sergiusens (approved):
looks good, had to dig into the craft-providers code to verify we weren't over removing and it seems like we are not

PR #79, opened 2022-09-20
cli: capture logs in logfile

mr-cal:

-----

When `rockcraft` is running in managed mode (i.e. inside an instance), then save the log file.

Referenced from [snapcraft](https://github.com/snapcore/snapcraft/blob/9e10fd1b658df80c35e1996cd92a077e831a8d51/snapcraft/cli.py#L166).


facundobatista (approved):
+1 awesome

PR #80, opened 2022-09-21
providers: move `get_instance_name()` call to lifecycle.py

mr-cal:

-----
Another bite-sized chunk of separating application-specific logic out of the `Provider` classes.

`get_instance_name()` is a rockcraft-specific function, so it is now called from `lifecycle.py`.

(CRAFT-1379)

### DIFF
--- a/rockcraft/providers/_provider.py
+++ b/rockcraft/providers/_provider.py
@@ -71,6 +71,8 @@ class Provider(ABC):
         :returns: True if installed.
         """
 
+    # TODO: migrate `create_environment()` from snapcraft or charmcraft
+
     @abstractmethod
     @contextlib.contextmanager
     def launched_environment(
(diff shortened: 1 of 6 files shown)


rockcraft/providers/_provider.py
```diff
@@ -79,10 +81,16 @@ def launched_environment(
...
-        """Launch environment for specified base.
+        """Configure and launch environment for specified base.
+
+        When this method loses context, all directories are unmounted and the
+        environment is stopped. For more control of environment setup and teardown,
+        use `create_environment()` instead.
 
-        :param project_name: Name of the project.
-        :param project_path: Path to the project.
+        :param project_name: Name of project.
```
facundobatista, line 92: this English language detail is quite counterintuitive for me :/

facundobatista (approved):
Awesome!!!

PR #81, opened 2022-09-21
providers: move base configuration logic to providers.py

mr-cal:

-----

Configuring a base is specific to rockcraft, so it is:
1. encapsulated in a function and moved to `providers.py` 
2. called by `lifecycle.py` instead of the Provider classes `lxd.py` and `multipass.py`

(CRAFT-1380)

### DIFF
--- a/rockcraft/providers/providers.py
+++ b/rockcraft/providers/providers.py
@@ -95,7 +95,7 @@ def capture_logs_from_instance(instance: executor.Executor) -> None:
 
 def get_base_configuration(
     *, alias: bases.BuilddBaseAlias, project_name: str, project_path: Path
-):
+) -> bases.BuilddBase:
     """Create a BuilddBase configuration for rockcraft."""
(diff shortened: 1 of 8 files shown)


rockcraft/providers/providers.py
```diff
@@ -87,3 +91,33 @@ def capture_logs_from_instance(instance: executor.Executor) -> None:
             emit.trace(
                 f"Could not find log file {source_log_path.as_posix()} in instance."
             )
+
+
+def get_base_configuration(
+    *, alias: bases.BuilddBaseAlias, project_name: str, project_path: Path
+):
```
cmatsuoka, line 98: We can add a return type annotation here.

cmatsuoka (approved):
LGTM, added a comment about a return type annotation.

rockcraft/providers/providers.py
mr-cal: Good catch!  Added.

PR #84, opened 2022-09-22
Add chisel to rockcraft

tigarmo:
This PR lets rockcraft support chisel slices in stage-packages. 
Notes:

- "chisel" is added directly on rockcraft's snap, just like skopeo and 
  umoci, as it's only an "image-build-time" dependency;
- We currently pin `craft-parts` to a specific commit on `main` because
  we don't yet have a release of craft-parts with chisel support, but 
  that should come soon;
- The added spread test uses the dotnet-runtime-6.0_libs slice, and 
  currently overrides `prime` to create symlinks because the internal 
  directory structure of the slice changes with new versions. This will
  be addressed/discussed in the future.

CRAFT-1333

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -100,3 +100,14 @@ parts:
         - libbtrfs-dev
         - libdevmapper-dev
         - pkg-config
+
+  chisel:
+    plugin: nil
+    source: https://github.com/canonical/chisel.git
+    build-snaps:
+      - go/1.18/stable
+    override-build: |
+      go mod download
+      CGO_ENABLED=0 GOOS=linux go build -o chisel ./cmd/chisel
+      mkdir "$SNAPCRAFT_PART_INSTALL"/bin
+      install -m755 chisel "$SNAPCRAFT_PART_INSTALL"/bin/chisel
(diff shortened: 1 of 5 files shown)


snap/snapcraft.yaml
```diff
@@ -102,11 +102,22 @@ parts:
...
+    plugin: nil
     source: https://github.com/canonical/chisel.git
     #source-commit: f0bff5a30dfdcb400b
     build-snaps:
       - go/1.18/stable
     override-build: |
       go mod download
-      go install -p "1" ./...
\ No newline at end of file
+      CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -o chisel ./cmd/chisel
```
cjdcordeiro, line 112: btw, I don't think GOARCH should be hardcoded, but rather left to default into the snapcraft's `architecture.build-for`

snap/snapcraft.yaml
```diff
@@ -100,3 +100,15 @@ parts:
         - libbtrfs-dev
         - libdevmapper-dev
         - pkg-config
+
+  chisel:
+    plugin: nil
+    source: https://github.com/canonical/chisel.git
+    #source-commit: f0bff5a30dfdcb400b
```
sergiusens, line 107: leftover?

sergiusens (approved):

PR #86, opened 2022-09-22
providers: use ProviderError from craft-providers

mr-cal:

-----

Use `ProviderError` from craft-providers.

### DIFF
--- a/rockcraft/cli.py
+++ b/rockcraft/cli.py
@@ -21,6 +21,7 @@ import sys
 
 import craft_cli
 from craft_cli import ArgumentParsingError, EmitterMode, ProvideHelpException, emit
+from craft_providers import ProviderError
 
 from rockcraft import __version__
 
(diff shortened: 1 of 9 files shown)


sergiusens (changes requested):
this is now missing an explicit catch here https://github.com/canonical/rockcraft/blob/main/rockcraft/cli.py#L94

rockcraft/cli.py
```diff
@@ -94,3 +95,9 @@ def run():
     except craft_cli.CraftError as err:  # TODO: define RockcraftError
         emit.error(err)
         sys.exit(1)
+    except ProviderError as err:
+        emit.error(craft_cli.CraftError(f"craft-providers error: {err}"))
+        sys.exit(1)
+    except Exception as err:  # pylint: disable=broad-except
```
mr-cal, line 101: I saw this broad exception in charmcraft and added it while I was here.

mr-cal:
Added a catch and a battery of tests.

sergiusens (approved):

PR #91, opened 2022-09-26
providers: move `confirm_with_user()` call to providers.py

mr-cal:

-----
Two changes:
1. `confirm_with_user()` is a rockcraft-specific call, so it is moved out of the craft-providers interface.
    - The user confirmation logic was split out to `providers.py`
    - `lxd.py` and `multipass.py` still handle the installation and ensure the provider is available
3. Rename `is_provider_available()` to `is_provider_installed()`

(CRAFT-1387)

### DIFF
--- a/rockcraft/providers/_lxd.py
+++ b/rockcraft/providers/_lxd.py
@@ -23,8 +23,6 @@ from typing import Generator, List
 
 from craft_providers import Executor, ProviderError, base, bases, lxd
 
-from rockcraft.utils import confirm_with_user
-
 from ._provider import Provider
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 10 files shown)


rockcraft/providers/_lxd.py
```diff
@@ -23,8 +23,6 @@
 
 from craft_providers import Executor, ProviderError, base, bases, lxd
 
-from rockcraft.utils import confirm_with_user
```
mr-cal, line 26: :tada: :tada: - The last `rockcraft` imports have been pulled out of `_lxd.py`, `multipass.py`, and `_provider.py`!

facundobatista (approved):
+1, great work. I wonder why `is_provider_installed` is a `classmethod`, but we can leave that conversation outside this PR. Thanks!

PR #93, opened 2022-09-27
commands: add clean command

mr-cal:

-----
#### Overview

Add command `rockcraft clean` with stubbed out TODOs.  This lays down groundwork needed to refactor `clean_project_environments()`.

There are 3 ways the `clean` command is used:
#### 1. `rockcraft clean <part>` 
- cleans a specific part
- added as a TODO to be implemented later

#### 2. `rockcraft clean --destructive-mode`
- cleans the local build environment
- added as a TODO, but users shouldn't encouter this, since `destructive mode` is also unsupported and labeled with a TODO

#### 3. `rockcraft clean`
- cleans a provider environment
- added as a TODO, but will be supported in my next PR

(CRAFT-1394)

### DIFF
--- a/tests/unit/test_lifecycle.py
+++ b/tests/unit/test_lifecycle.py
@@ -19,7 +19,7 @@ from pathlib import Path
 from unittest.mock import Mock, patch
 
 import pytest
-from craft_cli import EmitterMode, emit
+from craft_cli import CraftError, EmitterMode, emit
 from craft_providers.bases.buildd import BuilddBaseAlias
 
 from rockcraft import lifecycle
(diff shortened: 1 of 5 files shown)


tests/unit/test_lifecycle.py
```diff
@@ -41,6 +47,76 @@ def mock_provider(mocker, mock_instance, fake_provider):
     yield _mock_provider
 
 
+@pytest.fixture()
+def mock_get_instance_name(mocker):
+    with patch(
```
tigarmo, line 52: is there a specific reason to use `with patch()` here since you're already using `mocker`?

tigarmo (approved):
thankss

tests/unit/test_lifecycle.py
mr-cal: Whoops! I switched these to fixtures to `mocker`.

PR #94, opened 2022-09-27
spread: add 'global' clean-up calls

tigarmo:
The commands in `restore-each` are run after each task, which lets us cleanup docker images/lxc containers/etc without having to "remember" to do it per-task.

-----


mr-cal (approved):
Looks great, nice work!

PR #107, opened 2022-09-29
spread: restrict the files we package

tigarmo:
The spread run only needs the 'test/' and 'tools/' dir. This change has little impact on CI runs because they start from a 'clean' checkout, but for local runs the behavior of adding everything can add hundreds of MB of local 'junk' (specially from the virtual environment). So this speeds up local runs a bit.

-----


mr-cal (approved):
+1, shrinks my payload from 384MB to 99MB

Issue #112, opened 2022-09-30
bug: `init` is advertised as a lifecycle command


cjdcordeiro:
https://github.com/canonical/rockcraft/blob/f72beac0b8f6404217c2a8a4cc9608fdaa7c8fcd/rockcraft/cli.py#L42

`init` is not longer a lifecycle command since https://github.com/canonical/rockcraft/pull/35/commits/908f04cebddd2d2050ec0ade874155096fa107d0

PR #113, opened 2022-09-30
fix: #112 and help messages

cjdcordeiro:

-----

closes https://github.com/canonical/rockcraft/issues/112

Move `init` out of the lifecycle commands and fixes the help messages that were inherited from snapcraft.

### DIFF
--- a/rockcraft/cli.py
+++ b/rockcraft/cli.py
@@ -39,9 +39,14 @@ COMMAND_GROUPS = [
             commands.StageCommand,
             commands.PrimeCommand,
             commands.PackCommand,
+        ],
+    ),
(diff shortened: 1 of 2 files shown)


rockcraft/cli.py
```diff
@@ -39,9 +39,14 @@
             commands.StageCommand,
             commands.PrimeCommand,
             commands.PackCommand,
+        ],
+    ),
+    craft_cli.CommandGroup(
+        "Other",
```
tigarmo, line 45: nice catch hehe

tigarmo (approved):
thanks for catching this so soon

PR #118, opened 2022-10-04
tests: use tmp_path instead of mock_path

mr-cal:

-----

Tiny change to a unit test.

The test fixture `providers/conftest.py::mock_path` is leaving rockcraft, so I removed the only reference to it inside rockcraft's test suite.


tigarmo (approved):
:+1: thanks

PR #126, opened 2022-10-13
Docs improvements

evildmp:
Made some documentation improvements

- [* ] Have you signed the [CLA](http://www.ubuntu.com/legal/contributors/)?

-----

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,8 +42,8 @@ coverage: ## Run pytest with coverage report.
 docs: ## Generate documentation.
 	rm -f docs/rockcraft.rst
 	rm -f docs/modules.rst
-	$(MAKE) -C docs clean
-	$(MAKE) -C docs html
+	. $(DOCSVENV); $(MAKE) -C docs clean
+	. $(DOCSVENV); $(MAKE) -C docs html
 
 .PHONY: rundocs
 rundocs: ## start a documentation runserver
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,6 @@
+DOCSVENV = docs/env/bin/activate
+PORT = 8080
+
 .PHONY: help
 help: ## Show this help.
 	@printf "%-40s %s\n" "Target" "Description"
@@ -41,6 +44,11 @@ docs: ## Generate documentation.
 	rm -f docs/modules.rst
 	$(MAKE) -C docs clean
 	$(MAKE) -C docs html
+
+.PHONY: rundocs
+rundocs: ## start a documentation runserver
+	. $(DOCSVENV); sphinx-autobuild $(ALLSPHINXOPTS) --ignore ".git/*" --ignore "*.scss" ./docs -b dirhtml -a docs/_build/html --host 0.0.0.0 --port $(PORT)
+
 
 .PHONY: dist
 dist: clean ## Build python package.
--- a/docs/how-to/build-docs.rst
+++ b/docs/how-to/build-docs.rst
@@ -8,7 +8,7 @@ Create a virtual environment and activate it::
 
 Install the documentation requirements::
 
-    pip install -r requirements-doc.txt
+    pip install -r requirements-focal.txt -r requirements-doc.txt
 
 Once the requirements are installed, you can use the provided ``Makefile`` to build the documentation::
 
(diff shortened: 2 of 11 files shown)


requirements-doc.txt
```diff
@@ -42,3 +43,4 @@ types-Deprecated==1.2.9
 typing_extensions==4.3.0
 urllib3==1.26.12
 wrapt==1.14.1
+sphinx-autobuild
```
tigarmo, line 46: how come `sphinx-autobuild` is listed twice?

Makefile
```diff
@@ -42,6 +45,11 @@ docs: ## Generate documentation.
 	$(MAKE) -C docs clean
 	$(MAKE) -C docs html
 
+.PHONY: rundocs
+rundocs: ## start a documentation runserver
+	. $(DOCSVENV); sphinx-autobuild $(ALLSPHINXOPTS) --ignore ".git/*" --ignore "*.scss" ./docs -b dirhtml -a docs/_build/html --host 0.0.0.0 --port $(PORT)
```
tigarmo, line 50: Is activating the docs venv necessary here? The instructions have a step to activate it "manually", right?
`make docs` doesn't activate the env (that I can see), and the other makefile commands don't either, so maybe we should skip it

tigarmo:
Hi @evildmp, thanks for the fixes! It's weird that the cla-check is failing, as far as I can tell you're a member of the Canonical org. Have you had this happen before?

sergiusens:
For CLA, the first check sees that evildmp is not in the list of folks who have signed the CLA; which leads to the second check (for the author of the commit) and verifies that daniele@vurt.org is not on that list.

fwiw, for canonical, authoring with an @canonical.com address solves that issue

Makefile
evildmp: I have updated the instructions to reflect this, but yes, I believe that providing the simplest possible path to having getting the docs up and running locally is important to making working on documentation pleasant. I'd have done something similar for `make docs`, but I didn't want to mess with Makefile entries I didn't understand.

requirements-doc.txt
evildmp: Mistake; fixed.

Makefile
tigarmo: Alright, then I think we should update `make docs` too for consistency. I've tested locally that the following should work:

```
diff --git a/Makefile b/Makefile
index df0f013..a542d6f 100644
--- a/Makefile
+++ b/Makefile
@@ -42,8 +42,8 @@ coverage: ## Run pytest with coverage report.
 docs: ## Generate documentation.
        rm -f docs/rockcraft.rst
        rm -f docs/modules.rst
-       $(MAKE) -C docs clean
-       $(MAKE) -C docs html
+       . $(DOCSVENV); $(MAKE) -C docs clean
+       . $(DOCSVENV); $(MAKE) -C docs html
```
(If you want I can add those changes on your branch too, I just didn't want to do it without discussing first)

docs/how-to/build-docs.rst
```diff
@@ -0,0 +1,22 @@
...
+******************************
+
+Create a virtual environment and activate it::
+
+    python3 -m venv docs/env
+    source docs/env/bin/activate
+
+Install the documentation requirements::
+
+    pip install -r requirements-doc.txt
```
tigarmo, line 11: I tried to follow this locally but I got an error with `python-apt`; the command that worked for me is `pip install -r requirements-focal.txt -r requirements-doc.txt`

Makefile
evildmp: Go for it @tigarmo thanks.

docs/how-to/build-docs.rst
evildmp: I can't comment on that, the plain pip install -r requirements-doc.txt worked for me, but it could be that I have some global Python components or environment conditions that make it work.

tigarmo (approved):
thanks!

tigarmo:
@evildmp the build will probably fail because of an unrelated bug that has since been fixed in `main`. If you're done with the changes I can update/rebase from `main` in and schedule the final merge

evildmp:
Please do @tigarmo

PR #131, opened 2022-10-20
Craft 1415 rockcraft permissions

tigarmo:

-----

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -13,7 +13,7 @@ codespell==2.2.1
 commonmark==0.9.1
 coverage==6.4.4
 craft-cli==1.2.0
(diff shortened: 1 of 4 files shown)


mr-cal (approved):

cjdcordeiro:
thanks for the PR. I see it is still in draft mode, but given our recent conversations about `owner` vs `user-id`, I'm wondering: has this version also been upgraded in Snapcraft already?

tigarmo:
I don't think Snapcraft has been released with this new feature yet, but it won't take too long. 
But I've been thinking about the `user-id` issue and I think we might be able to keep the `craft-parts` implementation as-is and then add `user-id` "support" in `rockcraft`. We'll chat!

PR #139, opened 2022-11-08
parts: compile Pebble as a static binary

tigarmo:

-----

### DIFF
--- a/rockcraft/project.py
+++ b/rockcraft/project.py
@@ -501,7 +501,7 @@ def _add_pebble_data(yaml_data: Dict[str, Any]) -> None:
         "override-build": (
             "go mod download\n"
             'CGO_ENABLED=0 go build -ldflags="-w -s" -o pebble ./cmd/pebble\n'
-            'install -D -m755 pebble "$CRAFT_PART_INSTALL"/bin/pebble\n'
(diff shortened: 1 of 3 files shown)


rockcraft/project.py
```diff
@@ -498,5 +498,10 @@ def _add_pebble_data(yaml_data: Dict[str, Any]) -> None:
         "plugin": "go",
         "source": "https://github.com/canonical/pebble.git",
         "build-snaps": ["go/1.19/stable"],
+        "override-build": (
+            "go mod download\n"
+            'CGO_ENABLED=0 go build -ldflags="-w -s" -o pebble ./cmd/pebble\n'
+            'install -D -m755 pebble "$CRAFT_PART_INSTALL"/bin/pebble\n'
```
cjdcordeiro, line 504: should we put it in `/usr/bin/` instead? At least while a solution for preserving symlinks is being worked on...

tigarmo: sure, updated

cjdcordeiro (approved):

rockcraft/project.py
sergiusens: Putting in /usr/bin is interesting, as if the base is bare it is a strange directory structure to have. After some thought, I believe we need a rocks owned directory (we already do).

This is good enough for now, but let's make sure we talk about this in one of our live conversations

sergiusens (approved):

Issue #128, opened 2022-10-13
The `go` build-snap doesn't install the requested go version


gruyaume:
**Description**

Using the `go` build-snap doesn't allow for using the specified go version. The go version installed is always 1.19.

**Example**

rockcraft.yaml:
```yaml
  golang:
    plugin: nil
    build-snaps:
      - go/1.13/stable
    override-build: |
      set -x
      go version
      craftctl default
      go version
```

Execution output:

```
guillaume@thinkpad:~/PycharmProjects/magma-controller-rock$ rockcraft pack -v
Starting Rockcraft 0.0.1.dev1                                                                                                                                                                                                                          
Logging execution to '/home/guillaume/.cache/rockcraft/log/rockcraft-20221013-170003.209576.log'                                                                                                                                                       
Launching instance...                                                                                                                                                                                                                                  
Starting Rockcraft 0.0.1.dev1                                                                                                                                                                                                                          
Logging execution to '/tmp/rockcraft.log'                                                                                                                                                                                                              
Retrieving base ubuntu:18.04                                                                                                                                                                                                                           
Retrieved base ubuntu:18.04 for amd64                                                                                                                                                                                                                  
Extracting ubuntu:18.04                                                                                                                                                                                                                                
Extracted ubuntu:18.04                                                                                                                                                                                                                                 
Initializing parts lifecycle                                                                                                                                                                                                                           
Executing parts lifecycle                                                                                                                                                                                                                              
Executing parts lifecycle: pull golang                                                                                                                                                                                                                 
Executing action                                                                                                                                                                                                                                       
Executed: pull golang                                                                                                                                                                                                                                  
Executing parts lifecycle: skip pull pebble (already ran)                                                                                                                                                                                              
Executing action                                                                                                                                                                                                                                       
Executed: skip pull pebble (already ran)                                                                                                                                                                                                               
Executing parts lifecycle: overlay golang                                                                                                                                                                                                              
Executing action                                                                                                                                                                                                                                       
Executed: overlay golang                                                                                                                                                                                                                               
Executing parts lifecycle: reapply pebble (previous layer changed)                                                                                                                                                                                     
Executing action                                                                                                                                                                                                                                       
Executed: reapply pebble (previous layer changed)                                                                                                                                                                                                      
Executing parts lifecycle: build golang                                                                                                                                                                                                                
Executing action                                                                                                                                                                                                                                       
:: + set -x                                                                                                                                                                                                                                            
[...]

PR #144, opened 2022-11-17
project: use isolated go to build pebble

sergiusens:
This allows a rockcrafter to install whatever version of go they would required.

-----

### DIFF
--- a/tests/unit/test_project.py
+++ b/tests/unit/test_project.py
@@ -84,12 +84,16 @@ def yaml_loaded_data():
 def pebble_part():
     return {
         "pebble": {
-            "plugin": "go",
+            "plugin": "nil",
             "source": "https://github.com/canonical/pebble.git",
-            "build-snaps": ["go/1.19/stable"],
+            "override-pull": (
+                "craftctl default\n"
+                "snap set system experimental.parallel-instances=true\n"
+                "snap install go_pebble --channel 1.19/stable --classic\n"
+            ),
             "override-build": (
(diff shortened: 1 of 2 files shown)


sergiusens:
This is a PoC to fix #128

cjdcordeiro:
interesting

tests/unit/test_project.py
```diff
@@ -84,12 +84,16 @@ def yaml_loaded_data():
...
     return {
         "pebble": {
-            "plugin": "go",
+            "plugin": "nil",
             "source": "https://github.com/canonical/pebble.git",
-            "build-snaps": ["go/1.19/stable"],
+            "override-pull": (
+                "craftctl default\n"
+                "snap set system experimental.parallel-instances=true\n"
+                "snap install go_pebble --channel 1.19/stable --classic\n"
```
tigarmo, line 92: I see, so `go_pebble` means an instance of the `go` snap called `pebble`?
I found this reference, for whoever else was unaware for parallel instances: https://snapcraft.io/docs/parallel-installs

tigarmo (approved):
I know this is a draft but I say we merge it and let it play out a little bit, to see if any issues crop up

sergiusens:
I would like for parallel installs to be natively supported by craft parts but also have some for of detection to not need to declare the equivalent of CC and CXX for go

sergiusens:
none of what I mentioned is super important for this _hidden_ part until we start thinking about manifests

PR #146, opened 2022-11-21
project: install pebble's go without aliases setup

sergiusens:
This avoids a conflict when the user's go is installed.

-----

### DIFF
--- a/rockcraft/project.py
+++ b/rockcraft/project.py
@@ -500,7 +500,7 @@ def _add_pebble_data(yaml_data: Dict[str, Any]) -> None:
         "override-pull": (
             "craftctl default\n"
             "snap set system experimental.parallel-instances=true\n"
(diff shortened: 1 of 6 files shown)


rockcraft/project.py
```diff
@@ -500,7 +500,7 @@ def _add_pebble_data(yaml_data: Dict[str, Any]) -> None:
         "override-pull": (
             "craftctl default\n"
             "snap set system experimental.parallel-instances=true\n"
-            "snap install go_pebble --channel 1.19/stable --classic\n"
+            "snap install --unaliased go_pebble --channel 1.19/stable --classic\n"
```
tigarmo, line 503: should we maybe add a spread test that uses `plugin: go` to see what else breaks?

tigarmo (approved):
thanks!

PR #154, opened 2022-11-30
first pass at handling usrmerge

tigarmo:
This commit is a first step in gracefully handling existing layers when
packing the parts' contents. Specifically, it fixes the existing
breakage where priming a file in "/lib" will break a resulting
Ubuntu-based rock because on the base "/lib" is a symlink to "/usr/lib",
and if that link is broken everything breaks.

The updates to support this are:

- Enhance oci.Image.add_layer() to take the existing base into account,
  via a parameter that points to the extracted base (see below);
- Update lifecycle code to pass the extracted base forward when packing;
- Add a jammy runner to CI, because the new integrated tests need umoci
  which is only available on jammy.

The update to add_layer() is to look for a file's parent dir in the
existing base. If it *does* exist, and is a symlink to another dir, add
the file as belonging to the symlink target instead. For example:

- The prime dir contains "bin/file.txt";
- The base is ubuntu:22.04, which means that there "/bin/" is a symlink
  to "/usr/bin";
- Instead of adding the file as "bin/file.txt", add it as
  "usr/bin/file.txt".

Note that the original idea of just adding "bin/file.txt" but *not* adding
the "bin/" itself seems to be flaky; it works on *some* configurations of
the relevant tools but not others (more investigation necessary).


tests/integration/test_oci.py
```diff
@@ -0,0 +1,128 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
```
sergiusens, line 1: you can remove test_stub.py now :-)

rockcraft/oci.py
```diff
@@ -320,15 +337,42 @@ def _add_layer_into_image(image_path: Path, archived_content: Path, **kwargs) ->
...
+        base below this new layer. <usrmerge>
     """
+    # with tarfile.open(temp_tar_file, mode="w") as tar_file:
+    #     # With `arcname="."` the contents are added relative to `layer_path`.
+    #     tar_file.add(layer_path, arcname=".", filter=filter_func)
     with tarfile.open(temp_tar_file, mode="w") as tar_file:
-        # With `arcname="."` the contents are added relative to `layer_path`.
-        tar_file.add(layer_path, arcname=".")
+        for dirpath, subdirs, filenames in os.walk(layer_path):
+            # Sort `subdirs` in-place, to ensure that `os.walk()` iterates on
```
sergiusens, line 355: nice comment!

tests/integration/test_oci.py
tigarmo: this PR is very much a work-in-progress, don't blame me if you look at the code and turn into stone!

rockcraft/oci.py
```diff
@@ -320,15 +339,62 @@ def _add_layer_into_image(image_path: Path, archived_content: Path, **kwargs) ->
...
 
 
-def _archive_layer(layer_path: Path, temp_tar_file: Path) -> None:
+def _archive_layer(
+    layer_path: Path, temp_tar_file: Path, lower_rootfs: Optional[Path] = None
+) -> None:
     """Prepare new OCI layer by archiving its content into tar file.
 
     :param layer_path: path to the content to be archived into a layer
     :param temp_tar_file: path to the temporary tar fail holding the archived content
```
lengau, line 348: I know you didn't touch this line but I saw a typo anyway

```suggestion
    :param temp_tar_file: path to the temporary tar file holding the archived content
```

rockcraft/lifecycle.py
```diff
@@ -135,11 +136,17 @@ def _pack(
     base_digest: bytes,
     rock_suffix: str,
     build_for: str,
+    base_layer_dir: Path,
 ):
-    """Create the rock image for a given architecture."""
+    """Create the rock image for a given architecture.
+
+    :param base_layer_dir: The directory where the rock's base image was extracted.
```
cjdcordeiro, line 143: since you're at it, might as well add the other args

rockcraft/oci.py
```diff
@@ -119,39 +119,58 @@ def copy_to(self, image_name: str, *, image_dir: Path) -> "Image":
 
         return Image(image_name=image_name, path=image_dir)
 
-    def extract_to(self, bundle_dir: Path) -> Path:
+    def extract_to(self, bundle_dir: Path, *, rootless=False) -> Path:
```
cjdcordeiro, line 122:
```suggestion
    def extract_to(self, bundle_dir: Path, *, rootless: bool = False) -> Path:
```

tests/spread/general/big/task.yaml
```diff
@@ -50,3 +50,8 @@ execute: |
   docker cp big-container:/usr/bin/pebble - | tar xf -
   file pebble | grep "statically linked"
   docker rm -f big-container
+
+  ############################################################################################
+  # check that the /bin symlink to /usr/bin wasn't broken by the /bin/new_bin_file addition
+  ############################################################################################
+  docker run --rm big ls /bin/bash /bin/new_bin_file
```
cjdcordeiro, line 57: isn't this file always going to be there? I guess you should check instead that this **isn't** the only file, and that all other binaries are symlinks instead of just this one...?

cjdcordeiro (changes requested):

rockcraft/lifecycle.py
```diff
@@ -135,11 +136,17 @@ def _pack(
...
+    """
     emit.progress("Creating new layer")
     new_image = project_base_image.add_layer(
-        tag=project.version, layer_path=lifecycle.prime_dir
+        tag=project.version,
+        layer_path=lifecycle.prime_dir,
+        lower_rootfs=base_layer_dir,
```
cmatsuoka, line 149: Maybe we could use `base_layer_path` as the argument name to keep variable names consistent with `base_layer_...` and also with `layer_path`? (or `layer_dir` and `base_layer_dir`)

rockcraft/oci.py
```diff
@@ -320,15 +339,62 @@ def _add_layer_into_image(image_path: Path, archived_content: Path, **kwargs) ->
...
+                        f"Skipping {upper_subpath} because it exists as a symlink on the lower layer"
+                    )
+                    add_dir = not lower_path.is_symlink()
+                    if lower_path.is_symlink():
+                        lower_symlink_target = Path(os.readlink(lower_path))
+
+                if add_dir:
+                    emit.trace(f"Adding to layer: {upper_subpath}")
+                    tar_file.add(
+                        upper_subpath, arcname=f"./{relative_path}", recursive=False
```
cmatsuoka, line 382: Is `./` required?

rockcraft/oci.py
```diff
@@ -320,15 +339,62 @@ def _add_layer_into_image(image_path: Path, archived_content: Path, **kwargs) ->
...
+            # - The directory is not the root (to skip a spurious "." entry);
+            # - The directory's name does not exist on ``lower_rootfs`` as a symlink
+            #   to another directory (like in usrmerge).
+            if upper_subpath != layer_path:
+                add_dir = True
+                relative_path = upper_subpath.relative_to(layer_path)
+
+                if lower_rootfs is not None:
+                    lower_path = lower_rootfs / relative_path
+                    emit.trace(
```
cmatsuoka, line 372: This should be `emit.debug()`.

rockcraft/oci.py
```diff
@@ -320,15 +339,62 @@ def _add_layer_into_image(image_path: Path, archived_content: Path, **kwargs) ->
...
+
+            upper_subpath = Path(dirpath)
+            lower_symlink_target: Optional[Path] = None
+
+            # Handle adding an entry for the directory. We only do that if:
+            # - The directory is not the root (to skip a spurious "." entry);
+            # - The directory's name does not exist on ``lower_rootfs`` as a symlink
+            #   to another directory (like in usrmerge).
+            if upper_subpath != layer_path:
+                add_dir = True
```
cmatsuoka, line 367: Maybe use `is_symlink` instead and set to False?

rockcraft/oci.py
```diff
@@ -320,15 +339,62 @@ def _add_layer_into_image(image_path: Path, archived_content: Path, **kwargs) ->
...
+                relative_path = upper_subpath.relative_to(layer_path)
+
+                if lower_rootfs is not None:
+                    lower_path = lower_rootfs / relative_path
+                    emit.trace(
+                        f"Skipping {upper_subpath} because it exists as a symlink on the lower layer"
+                    )
+                    add_dir = not lower_path.is_symlink()
+                    if lower_path.is_symlink():
```
cmatsuoka, line 376: Then we could set `is_symlink` to `lower_path.is_symlink()`, and test the newly set variable.

rockcraft/oci.py
```diff
@@ -320,15 +339,62 @@ def _add_layer_into_image(image_path: Path, archived_content: Path, **kwargs) ->
...
+                        lower_symlink_target = Path(os.readlink(lower_path))
+
+                if add_dir:
```
cmatsuoka, line 379: And this happens if it's not a symlink.

rockcraft/oci.py
```diff
@@ -320,15 +339,62 @@ def _add_layer_into_image(image_path: Path, archived_content: Path, **kwargs) ->
...
+                    )
+
+            # Handle adding each file in the directory.
+            for name in filenames:
+                path = upper_subpath / name
+                if lower_symlink_target is not None:
+                    # If the directory containing `name` is a symlink on the lower
+                    # layer, add `name` into the symlink's target instead.
+                    # For example: if the file is `bin/file.txt`, and `bin` is a
+                    # symlink to `usr/bin`, add the file as `usr/bin/file.txt`.
```
cmatsuoka, line 392: If we implement a separate utility function to resolve the symlink based on the base filesystem we could later improve it to resolve symlinks to symlinks without changing the caller code.

rockcraft/oci.py
```diff
@@ -320,15 +339,62 @@ def _add_layer_into_image(image_path: Path, archived_content: Path, **kwargs) ->
...
+                    relative_path = lower_symlink_target / name
+                else:
+                    relative_path = Path(path).relative_to(layer_path)
+                emit.trace(f"Adding to layer: {path}")
+                tar_file.add(path, arcname=f"./{relative_path}")
```
cmatsuoka, line 397: Is `./` required?

tests/spread/general/big/task.yaml
lengau: If I understand correctly, the files in /usr/bin won't appear as symlinks here.

The best way to test this is probably to put a file into /usr/bin and then check for the existence of that file in /bin

tigarmo: Hmm I'm not sure I see the issue here. The command `ls /bin/bash /bin/new_bin_file` will only succeed if both `/bin/bash` and `/bin/new_bin_file` exist:

```bash
$ ls /bin/bash /bin/sbrubbles
ls: cannot access '/bin/sbrubbles': No such file or directory
/bin/bash
$ echo $?
2
```

Since `bash` is actually `/usr/bin/bash`, the command will only succeed if the `/bin` symlink to `/usr/bin` was successfully preserved, right? Note that the files themselves aren't symlinks, only `/bin` is.

lengau: Are we guaranteed that the bash file will be `/usr/bin/bash` in future? (E.g. might that get swapped around and `/usr/bin` symlinks to `/bin` at some point?) 

Controlling both files being tested might be more robust to unexpected changes upstream, but it may also be overkill.

rockcraft/oci.py
tigarmo: I thought it was weird that mypy didn't catch this, so I added #159 . Thanks for noticing this

tests/spread/general/big/task.yaml
cjdcordeiro: Yes indeed that's a more coherent test. In fact, any of the following tests seems sane enough:
 - len(/usr/bin) = len(/bin)+1
 - /bin -> /usr/bin
 - [ls /bin] = [ls /usr/bin , /bin/new_bin_file]

tigarmo: wait but the first and third checks aren't possible right? Since `/bin` is just a symlink to `/usr/bin`, there is no way for something to be reachable via `/bin` but not via `/usr/bin`

cjdcordeiro: yes you're right...it doesn't work as is. It would only make (some) sense if you could compare before and after adding the new_bin_file.

nevermind ;)

rockcraft/oci.py
tigarmo: I did some refactoring on this whole function, please take a look

cjdcordeiro (approved):
LGTM, thanks

cmatsuoka (approved):
Looks good, thank you.

lengau (approved):

lengau (approved):

PR #161, opened 2022-12-06
lint: disallow untyped defs in rockcraft code

tigarmo:

-----


mr-cal (approved):

rockcraft/project.py
```diff
@@ -96,15 +110,15 @@ class Platform(pydantic.BaseModel):
 
     @pydantic.validator("build_for", pre=True)
     @classmethod
-    def _vectorise_build_for(cls, val):
+    def _vectorise_build_for(cls, val: Any) -> Any:
```
lengau, line 113: Are we expecting to handle anything less strict than:

```suggestion
    def _vectorise_build_for(cls, val: Union[str, List[str]]) -> List[str]:
```

rockcraft/project.py
```diff
@@ -96,15 +110,15 @@ class Platform(pydantic.BaseModel):
...
         """Vectorise target architecture if needed."""
         if isinstance(val, str):
             val = [val]
         return val
 
     @pydantic.root_validator(skip_on_failure=True)
     @classmethod
-    def _validate_platform_set(cls, values):
+    def _validate_platform_set(cls, values: Any) -> Any:
```
lengau, line 121:
```suggestion
    def _validate_platform_set(cls, values: Mapping[str, Any]) -> Mapping[str, Any]:
```

rockcraft/project.py
```diff
@@ -175,15 +189,15 @@ def _validate_license(cls, rock_license):
 
     @pydantic.validator("title", always=True)
     @classmethod
-    def _validate_title(cls, title, values):
-        """If title is not provided, it default to the provided ROCK name."""
+    def _validate_title(cls, title: Optional[str], values: Any) -> str:
```
lengau, line 192:
```suggestion
    def _validate_title(cls, title: Optional[str], values: Mapping[str, Any]) -> str:
```

Or possibly `Mapping[str, str]`?

rockcraft/project.py
```diff
@@ -305,7 +319,7 @@ def _validate_all_platforms(cls, platforms):
 
     @pydantic.validator("parts", each_item=True)
     @classmethod
-    def _validate_parts(cls, item):
+    def _validate_parts(cls, item: Any) -> Any:
```
lengau, line 322: It can help the type checker here to make a `TypeVar` so it knows we're returning the same type as we're  accepting, especially if this method becomes more complex in future.

lengau (approved):
I've suggested some spots where there might be space for stricter type checking, but I'm approving regardless because this seems fine and you may well have had good reasons not to use stricter type annotations there.

rockcraft/project.py
tigarmo: I was undecided on doing that because this (specific) validator runs before the typical Pydantic validation of the field types, so `val` here is going to be whatever the used put in the `build_for` field (also I guess I got a bit lazy)
But then that's the case for every type annotation for function parameters in Python, so I changed it as you suggested, thanks!

rockcraft/project.py
tigarmo: In this case I updated it to `Dict[str, Any]` because this `_validate_parts()` function just forwards the parameter to `rockcraft.parts.validate_part()` which has that annotation. Thanks

tigarmo:
Thanks for the suggestions, I did not in fact have good reasons to not use stricter types, heh

cmatsuoka (approved):
Looks nice!

PR #163, opened 2022-12-08
oci: handle opaque whiteouts in new layer (usrmerge) 

tigarmo:
Now that add_layer() looks at the base, improve its logic to support
*explicit* removals/hiding of base layer symlinks. With this change,
regular directories created by lifecycle's "build" step will respect
symlinks in the base layer, but removals via overlay-scripts will
hide those symlinks.

### DIFF
--- a/rockcraft/oci.py
+++ b/rockcraft/oci.py
@@ -30,6 +30,7 @@ from typing import Any, Dict, List, Optional, Tuple
 
 import yaml
 from craft_cli import CraftError, emit
+from craft_parts.overlays import overlays
 
 logger = logging.getLogger(__name__)
 
@@ -368,16 +369,18 @@ def _archive_layer(
             # `relative_path` corresponds to a symlink in the base layer.
             archive_subdir = relative_path
 
-            # Handle adding an entry for the directory. We only do that if:
-            # - The directory is not the root (to skip a spurious "." entry);
-            # - The directory's name does not exist on ``base_layer_dir`` as a symlink
-            #   to another directory (like in usrmerge).
+            # Handle adding an entry for the directory. We skip this IF:
+            # - The directory is the root (to skip a spurious "." entry), OR
+            # - The directory is NOT an opaque OCI entry AND
+            # - The directory's exists on ``base_layer_dir`` as a symlink to another
+            #   directory (like in usrmerge).
             if upper_subpath != new_layer_dir:
+                is_not_opaque_dir = not overlays.is_oci_opaque_dir(upper_subpath)
                 lower_symlink_target = _symlink_target_in_base_layer(
                     relative_path, base_layer_dir
                 )
 
-                if lower_symlink_target is not None:
+                if is_not_opaque_dir and lower_symlink_target is not None:
                     emit.debug(
                         f"Skipping {upper_subpath} because it exists as a symlink on the lower layer"
                     )
--- a/rockcraft/oci.py
+++ b/rockcraft/oci.py
@@ -26,7 +26,7 @@ import tarfile
 import tempfile
 from dataclasses import dataclass
 from pathlib import Path
-from typing import Any, Dict, List, Optional, Tuple
+from typing import Any, Dict, List, Optional, Tuple, cast
 
 import yaml
 from craft_cli import CraftError, emit
@@ -375,16 +375,17 @@ def _archive_layer(
             # - The directory's exists on ``base_layer_dir`` as a symlink to another
             #   directory (like in usrmerge).
             if upper_subpath != new_layer_dir:
-                is_not_opaque_dir = not overlays.is_oci_opaque_dir(upper_subpath)
+                upper_is_not_opaque_dir = not overlays.is_oci_opaque_dir(upper_subpath)
                 lower_symlink_target = _symlink_target_in_base_layer(
                     relative_path, base_layer_dir
                 )
+                lower_is_symlink = lower_symlink_target is not None
 
-                if is_not_opaque_dir and lower_symlink_target is not None:
+                if upper_is_not_opaque_dir and lower_is_symlink:
                     emit.debug(
                         f"Skipping {upper_subpath} because it exists as a symlink on the lower layer"
                     )
-                    archive_subdir = lower_symlink_target
+                    archive_subdir = cast(Path, lower_symlink_target)
                 else:
                     emit.debug(f"Adding to layer: {upper_subpath}")
                     tar_file.add(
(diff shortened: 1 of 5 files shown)


tests/spread/general/big/rockcraft.yaml
```diff
@@ -38,5 +38,12 @@ parts:
   usrmerge-part:
     plugin: nil
     override-build: |
+      # This build script adds a file in bin/ - the symlink in the base should be preserved.
       mkdir ${CRAFT_PART_INSTALL}/bin
       touch ${CRAFT_PART_INSTALL}/bin/new_bin_file
+    overlay-script: |
+      # explicitly remove a base layer symlink and create a dir instead
+      cd ${CRAFT_OVERLAY}
+      rm lib32
```
cjdcordeiro, line 47: it would be interesting to do this in separate parts, i.e. `partOne` -> `rm /lib32`, and then `partTwo` -> `mkdir`.

rockcraft/oci.py
```diff
@@ -368,16 +369,18 @@ def _archive_layer(
...
             if upper_subpath != new_layer_dir:
+                is_not_opaque_dir = not overlays.is_oci_opaque_dir(upper_subpath)
                 lower_symlink_target = _symlink_target_in_base_layer(
                     relative_path, base_layer_dir
                 )
 
-                if lower_symlink_target is not None:
+                if is_not_opaque_dir and lower_symlink_target is not None:
                     emit.debug(
                         f"Skipping {upper_subpath} because it exists as a symlink on the lower layer"
```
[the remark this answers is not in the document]
tigarmo: I guess it's confusing but the two conditions on the `if` refer to two different paths: `is_not_opaque_dir` refers to the path in the "upper" (new) layer, and `lower_symlink_target` refers to the path in the "lower" (base) layer.
Would it be clearer if it said something like `if upper_is_not_opaque and lower_is_symlink:` instead?

tests/spread/general/big/rockcraft.yaml
tigarmo: Do you mean interesting in a testing sense? We can do that. I don't think that would be the typical usecase though

cjdcordeiro: Yes, in a test.

Indeed it is not the most common case, but I believe it is a more "stressful" case to test your changes. 
It is true that all parts are squashed into a single layer, so in theory this would still work well.

Maybe the real stress test is to create a `.rock` dir, since we always add a 2nd layer into the ROCK, for this path. Given that these are different layers, it is a true test for the whiteout and opaque features.

rockcraft/oci.py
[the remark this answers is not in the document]
cjdcordeiro: ok I see. ty.

y, why not, the more explicit the better.

tests/integration/test_oci.py
```diff
@@ -72,26 +119,68 @@ def test_add_layer_with_symlink_in_base(new_dir):
...
+                unlink bin
+                mkdir bin
+                touch bin/file_from_overlay_script
+                """
+            ),
+        }
+    }
+    work_dir = Path("work")
+
+    # Mock os.geteuid() because currently craft-parts doesn't allow overlays
```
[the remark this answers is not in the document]
cmatsuoka: The superuser test was added when we were using kernel overlayfs, maybe we can just remove the requirement now that we're using fuse.

cmatsuoka (approved):

tests/spread/general/big/rockcraft.yaml
tigarmo: > Maybe the real stress test is to create a .rock dir, since we always add a 2nd layer into the ROCK, for this path. Given that these are different layers, it is a true test for the whiteout and opaque features.

They are in different layers, but there wouldn't be a whiteout/opaque dir because nothing was "removed", right? Did you have a specific test in mind?

Anyway, I tried to do the original check of a part removing `/libx32` and then another part (re)creating `/libx32` and it didn't work: https://github.com/canonical/rockcraft/actions/runs/3697829487/jobs/6263399207
I expected the overlay mount to "persist" between parts but from the log it looks like it's mounted and unmounted for each part, so the original suggestion is not possible to express today. Is this something that we would want to support at some point?

tigarmo: It's interesting though that the spread test only failed on ubuntu-18.04 (so it passed 20.04, 22.04, fedora, etc)

cjdcordeiro: > a part removing /libx32 and then another part (re)creating /libx32 and it didn't work

interesting. I feel this should be possible, after all that's how people expect `overlay-*` to work -> you're making changes on the final overlay system (a la Dockerfile style).

wrt to testing this with different layers, you could try having a part that creates `/.rock` as a symlink to something, and then see what happens to the 2nd layer (it should preserve the symlink `/.rock` and still put the metadata file inside)

tigarmo: For the first point (one part removes/other part re-adds) I also feel like it should be possible. I suggest we create a separate issue to address this in the future

As for the second point, your suggestion is useful because [the test failed](https://github.com/canonical/rockcraft/actions/runs/3703751156/jobs/6275668885) and I realize that I haven't made this clear enough: the current implementation only looks at the base layer (ubuntu or bare), and not at the "current" layer stack. So the `.rock` symlink that the part added was hidden by the `.rock` dir in the metadata layer...
My suggestion for this is to *also* create an issue to support this in the future, which we'll definitely need if/when we start having per-part layers. In the meantime I can update the new test to document the fact that this *doesn't* work. Is this acceptable?

cjdcordeiro: > For the first point (one part removes/other part re-adds) I also feel like it should be possible. I suggest we create a separate issue to address this in the future

ok. but please make it a priority (maybe for the next sprint @sergiusens ?) cause it's clearly a problem.

> My suggestion for this is to also create an issue to support this in the future, which we'll definitely need if/when we start having per-part layers. 

ok. same as above then. with less priority though, since we don't yet support multi-layers (the .rock is a special case, implicitly caused by us)

Ok. sounds like a plan.

cjdcordeiro (approved):
good enough for now, considering that the limitations addressed in https://github.com/canonical/rockcraft/pull/163#discussion_r1049612263 should be fixed asap

tests/spread/general/big/rockcraft.yaml
tigarmo: Added #166 and #167 , thanks!

PR #168, opened 2022-12-15
project: move from spdx-license-list to spdx-lookup

tigarmo:
It looks like the spdx-license-list package was wholesale removed from PyPI, which is a pretty big breakage for us. The spdx-lookup isn't specially well updated, but it looks like it has the (very few) things we need from it.

-----


setup.cfg
```diff
@@ -32,7 +32,7 @@ install_requires =
     craft-parts
     craft-providers
     overrides
-    spdx-license-list
+    spdx-lookup
```
lengau, line 35: This project has [one release on pypi](https://pypi.org/project/spdx-lookup/#files) from 2020 and is basically just a wrapper around [another module](https://pypi.org/project/spdx/) also last updated in 2020.

It might be easier and more reliable to just create a list of license IDs ourselves with:
```python3
license_ids = [lic['licenseId'].lower() for lic in requests.get('https://github.com/spdx/license-list-data/raw/main/json/licenses.json').json()['licenses']]
```

or including the latest `licenses.json` in the wheel file at packaging time.

tigarmo: I agree that the package isn't in great shape, but I tried to do the "quickest" thing that would un-break `main` for now. The other solutions would need a bit more effort (and I'm not sure we should add a runtime dependency to github.com), I just didn't want rockcraft to stay broken over the holidays ;)

I added #169 to review this early next year.

mr-cal: I'm +1 for unbreaking main for the holidays.  

That package is so small and infrequently updated, that forking and maintaining may be an option :thinking: 
Or keeping a hardcopy of https://github.com/spdx/license-list-data/raw/main/json/licenses.json in this repo.

lengau: The other option is to use SPDX's [`spdx-tools`](https://github.com/spdx/tools-python) package, but that has problems of its own (including also having an outdated package file).

I agree that making this a January problem is probably the best way to go, so I'll put opinions in #169

lengau (approved):

mr-cal (approved):

cmatsuoka (approved):

PR #170, opened 2022-12-19
cli: don't expose the managed instance logpath

cmatsuoka:
The log path inside the managed instance is not directly accessible from the outer environment, so don't display it when ending the execution with an error message.

Co-authored-by: Facundo Batista <facundo@taniquetil.com.ar>

-----

### DIFF
--- a/rockcraft/cli.py
+++ b/rockcraft/cli.py
@@ -18,6 +18,7 @@
 
 import logging
 import sys
+from typing import Optional
 
 import craft_cli
 from craft_cli import ArgumentParsingError, EmitterMode, ProvideHelpException, emit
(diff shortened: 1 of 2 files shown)


mr-cal (approved):

rockcraft/cli.py
```diff
@@ -57,6 +58,20 @@
...
 
+def _emit_error(error: craft_cli.CraftError, cause: Optional[Exception] = None) -> None:
+    """Emit the error in a centralized way so we can alter it consistently."""
+    # set the cause, if any
+    if cause is not None:
+        error.__cause__ = cause
+
+    # if running inside a managed instance, do not report the internal logpath
+    if is_managed_mode():
+        error.logpath_report = False
```
SamirPS, line 69: Can we just do :

error.logpath_report = (not is_managed_mode()) ?

rockcraft/cli.py
```diff
@@ -57,6 +58,20 @@
 ]
 
 
+def _emit_error(error: craft_cli.CraftError, cause: Optional[Exception] = None) -> None:
```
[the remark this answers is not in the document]
cmatsuoka: It's not used now, but the option is available. This is a copy of @facundobatista's implementation in Charmcraft and has the same signature.

rockcraft/cli.py
cmatsuoka: In the hypothetical case where the variable already comes disabled for whatever reason, we may not want to turn it back on.

PR #174, opened 2023-01-05
tests: check before handling rockcraft instances

lengau:
The previous code always tried to stop/delete rockcraft-related lxc instances because all spread tests created them. Now we have spread tests that don't, so wrap the cleanup in an `if`.

-----

### DIFF
--- a/spread.yaml
+++ b/spread.yaml
@@ -80,8 +80,11 @@ prepare: |
 restore-each: |
   # Cleanup after each task.
   docker system prune -a -f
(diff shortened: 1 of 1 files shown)


lengau:
Created this because we got agreement in standup that completing #173 sooner rather than later is the right approach given this repository's inconsistent 2-reviewer requirement.

mr-cal (approved):

tigarmo (approved):
thanksss

PR #177, opened 2023-01-16
project: use pebble snap instead of building

tigarmo:
Now that pebble has been published as a snap we can use it directly instead of building the binary each time. Prime only the binary and "organize" it as /usr/bin/pebble just like the previous part did.

-----

(CRAFT-1566)

### DIFF
--- a/rockcraft/project.py
+++ b/rockcraft/project.py
@@ -508,16 +508,8 @@ def _add_pebble_data(yaml_data: Dict[str, Any]) -> None:
 
     pebble_part_spec = {
         "plugin": "nil",
-        "source": "https://github.com/canonical/pebble.git",
-        "override-pull": (
-            "craftctl default\n"
-            "snap set system experimental.parallel-instances=true\n"
-            "snap install --unaliased go_pebble --channel 1.19/stable --classic\n"
-        ),
-        "override-build": (
-            "go_pebble mod download\n"
-            'CGO_ENABLED=0 go_pebble build -ldflags="-w -s" -o pebble ./cmd/pebble\n'
-            'install -D -m755 pebble "$CRAFT_PART_INSTALL"/usr/bin/pebble\n'
-        ),
+        "stage-snaps": ["pebble/latest/edge"],
+        "organize": {"bin/pebble": "usr/bin/pebble"},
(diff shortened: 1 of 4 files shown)


tigarmo:
The spread failure is due to this bug: https://bugs.launchpad.net/snapd/+bug/2002835
All spread tests that run `rockcraft` in Fedora are currently failing

lengau (approved):
I presume we've got a reminder somewhere to change this to a more stable pebble snap once it's released to a channel other than edge

rockcraft/project.py
```diff
@@ -508,16 +508,8 @@ def _add_pebble_data(yaml_data: Dict[str, Any]) -> None:
...
-            "snap set system experimental.parallel-instances=true\n"
-            "snap install --unaliased go_pebble --channel 1.19/stable --classic\n"
-        ),
-        "override-build": (
-            "go_pebble mod download\n"
-            'CGO_ENABLED=0 go_pebble build -ldflags="-w -s" -o pebble ./cmd/pebble\n'
-            'install -D -m755 pebble "$CRAFT_PART_INSTALL"/usr/bin/pebble\n'
-        ),
+        "stage-snaps": ["pebble/latest/edge"],
+        "organize": {"bin/pebble": "usr/bin/pebble"},
```
cjdcordeiro, line 512: thanks for the PR. The binary should actually be placed in `/bin/pebble`, according to the spec (RK010)

cjdcordeiro (changes requested):

rockcraft/project.py
sergiusens: I suspect this is related to usrmerge

cjdcordeiro: sounds good. please re-request the review whenever ready ;)

tigarmo: g2g, the failing spread is because Fedora 35 is EoL (fixed in #179)

cjdcordeiro (approved):
LGTM, thanks

sergiusens:
Out of curiosity, what is the speed improvement here?

tigarmo:
@sergiusens I compared the spread runs with and without this change and it looks like we went from around ~32min to around ~26min (total). Pretty nice actually

PR #179, opened 2023-01-17
retry rockcraft commands in fedora

tigarmo:
Because of a snapd 2.58 bug(1), rockcraft may fail when running "snapd
set ..." commands inside the instance. This only happens once, and the
command succeeds the second time. Add a wrapper around rockcraft calls
to retry when in Fedora while the bug is not fixed.

Also update to Fedora 36 because 35 is EOL and broken already.

(1): (https://bugs.launchpad.net/snapd/+bug/2002835

-----

### DIFF
--- a/spread.yaml
+++ b/spread.yaml
@@ -100,6 +100,9 @@ debug-each: |
 suites:
   docs/tutorials/code/:
     summary: tests basic tutorials from the docs
+    # Don't run tutorial code in Fedora because of a snapd 2.58 bug
+    # (https://bugs.launchpad.net/snapd/+bug/2002835)
+    systems: [-fedora-36-64]
 
   docs/how-to/code/:
     summary: tests how-to guides from the docs


spread.yaml
```diff
@@ -100,6 +100,9 @@ debug-each: |
 suites:
   docs/tutorials/code/:
     summary: tests basic tutorials from the docs
+    # Don't run tutorial code in Fedora because of a snapd 2.58 bug
+    # (https://bugs.launchpad.net/snapd/+bug/2002835)
+    systems: [-fedora-36-64]
```
tigarmo, line 105: not sure what else we could do in this case; I don't want to use `run_rockcraft` in doc code because that'll end up in the, well, docs

sergiusens (approved):

spread.yaml
cjdcordeiro: ah, true. well, what about a different approach? Instead of creating this wrapper `run_rockcraft`, what about running the spread tests twice? 
1. run all spread tests, on all system **except** Fedora
2. run all spread tests, **only** on Fedora, **but** with the Spread [flag](https://github.com/snapcore/spread#repeating-tasks) `-repeat 2`?

tigarmo: Oh thanks for the suggestion, I didn't know about `-repeat`. The doc was a little vague so I tried it out and it looks like it's meant to help catch flakiness, because it stops as soon as a test fails:
```
$ spread -repeat 2 google:fedora-36-64:docs/tutorials/code/chisel
2023-01-19 09:39:17 Project content is packed for delivery (112.54MB).
2023-01-19 09:39:17 If killed, discard servers with: spread -reuse-pid=26713 -discard
2023-01-19 09:39:17 Allocating google:fedora-36-64...
2023-01-19 09:39:28 Waiting for google:fedora-36-64 (jan191239-148658) to boot at 34.75.53.67...
2023-01-19 09:40:05 Allocated google:fedora-36-64 (jan191239-148658).
2023-01-19 09:40:05 Connecting to google:fedora-36-64 (jan191239-148658)...
2023-01-19 09:40:07 Connected to google:fedora-36-64 (jan191239-148658) at 34.75.53.67.
2023-01-19 09:40:07 Sending project content to google:fedora-36-64 (jan191239-148658)...
2023-01-19 09:45:08 WARNING: google:fedora-36-64 (jan191239-148658) running late. Output still empty.
2023-01-19 09:47:38 Preparing google:fedora-36-64 (jan191239-148658)...
2023-01-19 09:49:52 Executing google:fedora-36-64:docs/tutorials/code/chisel (jan191239-148658) (1/1)...
2023-01-19 09:52:41 Error executing google:fedora-36-64:docs/tutorials/code/chisel (jan191239-148658) : 
-----
... <snip>
2023-01-19 09:52:43 Restoring google:fedora-36-64:docs/tutorials/code/chisel (jan191239-148658)...
2023-01-19 09:52:45 Discarding google:fedora-36-64 (jan191239-148658)...
2023-01-19 09:52:47 Successful tasks: 0
2023-01-19 09:52:47 Aborted tasks: 0
2023-01-19 09:52:47 Failed tasks: 1
    - google:fedora-36-64:docs/tutorials/code/chisel
error: unsuccessful run
```
so it looks like that doesn't help us unfortunately :/

cjdcordeiro: that's unfortunate. ok then, ty for trying

PR #197, opened 2023-02-17
oci: use alternative registry

tigarmo:
Docker Hub is fast but the Ubuntu images are updated infrequently and
the registry itself has pull limits for anonymous users. This limit is
annoying because currently rockcraft does not caching of base images.
    
The caching is coming, but in the meantime switch to Amazon ECR. The
registry has newer images and is only a little bit slower (from testing)
than Docker Hub).

### DIFF
--- a/rockcraft/oci.py
+++ b/rockcraft/oci.py
@@ -36,6 +36,12 @@ from rockcraft import errors
 
 logger = logging.getLogger(__name__)
 
+# Public Amazon Elastic Container Registry has images that are updated more frequently
+# and with less severe pull-rate limits than Docker Hub.
+ECR_URL = "public.ecr.aws/ubuntu"
+
+REGISTRY_URL = ECR_URL
+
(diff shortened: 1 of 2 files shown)


rockcraft/oci.py
```diff
@@ -36,6 +36,12 @@
 
 logger = logging.getLogger(__name__)
 
+# Public Amazon Elastic Container Registry has images that are updated more frequently
+# and with less severe pull-rate limits than Docker Hub.
+ECR_URL = "public.ecr.aws/ubuntu"
```
SamirPS, line 41: Why did you add this variable and not just do REGISTRY_URL = "public.ecr.aws/ubuntu" I don't see the value of this variable.

cjdcordeiro (approved):
lgtm, thanks

cmatsuoka (approved):
Looks good.

PR #200, opened 2023-02-17
oci: fix handling of subdirs in usrmerge

tigarmo:
The scenario that was broken was: a payload like "bin/dir1/dir2" where "bin" is a symlink to somewhere on the lower (base) layer. The subdirs "dir1" and "dir2" were being incorrectly placed in "bin" instead of in the symlink target.

This broke quite a few plugins that place file trees in places like "lib/".

-----

CRAFT-1605

### DIFF
--- a/rockcraft/oci.py
+++ b/rockcraft/oci.py
@@ -344,6 +344,7 @@ def _add_layer_into_image(
     _process_run(cmd + ["--history.created_by", " ".join(cmd)])
 
 
+# pylint: disable-next=too-many-locals
 def _archive_layer(
     new_layer_dir: Path, temp_tar_file: Path, base_layer_dir: Optional[Path] = None
 ) -> None:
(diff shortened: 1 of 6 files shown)


tests/spread/general/big/rockcraft.yaml
```diff
@@ -35,19 +35,6 @@ parts:
       - path: dir2/3.txt
         mode: "755"
 
-  usrmerge-part:
```
tigarmo, line 38: I moved this usrmerge check to its own spread test because I think it's "important" enough. Also it messes with common places like `${CRAFT_PART_INSTALL}/bin` so I feared these parts might start interfering with each other

rockcraft/oci.py
```diff
@@ -355,7 +356,41 @@ def _archive_layer(
         base below this new layer. Used to preserve lower-level directory symlinks,
         like the ones from Debian/Ubuntu's usrmerge.
     """
+
+    class LayerLinker:
```
cmatsuoka, line 360: This could even be a top level class.

cmatsuoka (approved):
I think it works! (And the implementation is quite simple)

sergiusens (approved):
nice work

cjdcordeiro (commented):
apologies for the late review. it does look good. thanks for the work

Issue #203, opened 2023-02-24
Error processing tar file(duplicates of file paths not supported)


gboutry:
Hi,

Trying to pack a ROCK image for the first time and encountering issues.

I'm trying to pack a Keystone image with the following `rockcraft.yaml`:

```yaml
name: keystone
summary: Openstack Keystone
license: Apache-2.0
description: |
  Ubuntu distribution of OpenStack Keystone
version: antelope

base: ubuntu:22.04
platforms:
  amd64:

parts:
  cloud-archive:
    plugin: nil
    build-packages:
      - ubuntu-cloud-keyring
      - software-properties-common
    override-pull: |
      craftctl default
      add-apt-repository cloud-archive:antelope
      apt update

  keystone:
    after: [cloud-archive]
    plugin: nil
    stage-packages:
      - apache2
      - keystone
```

When I try to import the generated ROCK image using `skopeo`:
```
skopeo copy --insecure-policy oci-archive:keystone_antelope_amd64.rock docker-daemon:keystone:antelope
Getting image source signatures
Copying blob 677076032cca done  
Copying blob 4af93843b91c done  
Copying blob 4bfd57fdda04 done  
Copying config 05d3bd154c done  
Writing manifest to image destination
Storing signatures
```

It's missing from docker registry:
```bash
docker images
REPOSITORY    TAG        IMAGE ID       CREATED         SIZE
hello-world   latest     feb5d9fea6a5   17 months ago   13.3kB
```

When uploading to local registry with `skopeo` then downloading with `docker`:
```bash
docker pull localhost:32000/keystone:antelope                                                                                            
antelope: Pulling from keystone
677076032cca: Already exists 
4af93843b91c: Extracting [==================================================>]  9.225MB/9.225MB
4bfd57fdda04: Download complete 
failed to register layer: Error processing tar file(duplicates of file paths not supported):
```

When following the procedure [here](https://github.com/canonical/rockcraft/issues/194):
```bash
skopeo copy --insecure-policy oci-archive:keystone_antelope_amd64.rock oci:keystone
❯ cd keystone 
❯ manifest=$(jq -r '.manifests[].digest' < index.json | cut -d ':' -f 2)
❯ cd blobs/sha256 
❯ blob_name=$(jq -r '.layers[1].digest' < $manifest | cut -d ':' -f 2)
❯ tar -tvf $blob_name | awk -F' ' '{print $6}' | sort | uniq -d       
usr/lib/x86_64-linux-gnu/
```

[...]

PR #204, opened 2023-02-28
oci: handle duplicate directories because of usrmerge 

tigarmo:
Reviewing note: the first commit restructures the code a bit without functional changes; the second commit has the actual bugfix.

=================
The situation is this: the primed payload has two distinct directories
(e.g. bin/dir1 and usr/bin/dir1) that, because of the usrmerge handling,
"point" to the same location (in this case, /usr/bin/dir1). We check
that the two directories have the same attributes (ownership and mode)
and then add a single entry to the layer (either one, they are all
equivalent).

All other cases (dirs with different attributes, conflicts between
files) raise errors. The reason for this is that we haven't seen this
come up "in the wild" yet, and it's hard to choose a behavior when we
don't know the circunstances.

(CRAFT-1634)

### DIFF
--- a/tests/unit/test_oci.py
+++ b/tests/unit/test_oci.py
@@ -20,14 +20,14 @@ import json
 import os
 import tarfile
 from pathlib import Path
-from typing import List
+from typing import List, Tuple
 from unittest.mock import ANY, call, mock_open, patch
 
 import pytest
 from craft_parts.overlays import overlays
 
 import tests
-from rockcraft import oci
+from rockcraft import errors, oci
 
 
 @pytest.fixture
@@ -297,7 +297,7 @@ class TestImage:
         rootfs_dir.mkdir()
 
         (rootfs_dir / "first").mkdir()
-        os.symlink("first", rootfs_dir / "second")
+        (rootfs_dir / "second").symlink_to("first")
 
         assert len(temp_tar_contents) == 0
         image.add_layer("tag", layer_dir, base_layer_dir=rootfs_dir)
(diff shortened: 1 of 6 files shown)


gboutry:
Just tested this, and it does solve #203, thank you for you work!

tigarmo:
Ah thanks for testing this @gboutry ! I had also tested your project but it's good to get confirmation from the person who knows what the ROCK is about, heh.

tests/unit/test_oci.py
```diff
@@ -402,6 +402,112 @@ def test_add_layer_with_base_layer_subdirs(self, tmp_path, temp_tar_contents):
...
+        (layer_dir / "bin/dir1").mkdir(parents=True)
+        (layer_dir / "bin/dir1/a.txt").touch()
+        (layer_dir / "usr/bin/dir1").mkdir(parents=True)
+        (layer_dir / "usr/bin/dir1/b.txt").touch()
+
+        rootfs_dir = tmp_path / "rootfs"
+        rootfs_dir.mkdir()
+        # In the base layer "bin" is a symlink to "usr/bin"
+        (rootfs_dir / "usr/bin").mkdir(parents=True)
+        os.symlink("usr/bin", rootfs_dir / "bin")
```
sergiusens, line 432: fyi, `pathlib.Path` has the amazing `symlink_to`

sergiusens (approved):
looks really good, one very small nit (ignorable); please use "rebase and merge" for this one.

tests/unit/test_oci.py
tigarmo: thanks! I have a strong memory of looking at `symlink_to` and not being able to use it for "some reason", but I updated the code now and it seems to work fine? weird

Issue #65, opened 2022-09-05
allow user configuration of the Pebble entrypoint binary


cjdcordeiro:
Users can define the Pebble services to be executed, but they should also be able to configure certain parameters about the Pebble software which is being installed in the ROCK.

The most obvious parameter is the Pebble version, i.e., users should be able to have a say on the version of Pebble that is being used as the ROCK entrypoint.

See https://github.com/canonical/rockcraft/pull/59#discussion_r954192413 for more details

PR #205, opened 2023-03-02
WIP: Add support for the Pebble entrypoint

cjdcordeiro:
Needs https://github.com/canonical/pebble/pull/191 and https://github.com/canonical/pebble/pull/188.

**TODO:**
Once the above PRs are merged:
 - re-run the tests
 - update the Tutorials and How-to guides

In this PR:
 - the `services` field is added to the Rockcraft project, with a lightweight validation according to the spec from https://github.com/canonical/pebble#layer-specification
 - the OCI entrypoint is fixed to always be `pebble enter`
 - the OCI cmd is `null`
 - the implicit `pebble` part now also creates the `$PEBBLE` path, where the Pebble layer is later be written
 - `entrypoint` and `cmd` are kept, but marked as deprecated, printing an error message and exiting the build whenever they are used
 - the Reference docs are updated

-----


rockcraft/lifecycle.py
```diff
@@ -159,13 +159,18 @@ def _pack(
     )
     emit.progress("Created new layer", permanent=True)
 
-    if project.entrypoint:
-        new_image.set_entrypoint(project.entrypoint)
-        if not project.cmd:
-            new_image.set_cmd([])
+    emit.progress("Adding Pebble entrypoint")
+    if project.services:
```
tigarmo, line 163: isn't this field mandatory now?

rockcraft/oci.py
```diff
@@ -215,34 +219,86 @@ def to_oci_archive(self, tag: str, filename: str) -> None:
...
+            "description": description,
+            "services": services,
+        }
+
+        # To infer the right filename for this Pebble layer, we should check
+        # if there are existing Pebble layers in the base root filesystem,
+        # and increment on that.
+        # NOTE: the layer's filename prefix will always be "001-" when using
+        # "bare" and "ubuntu" bases
+        pebble_layers_path_in_base = f"{base_layer_dir}/{self._PEBBLE_LAYERS_PATH}"
```
tigarmo, line 267: I feel like this method is doing too much work and only some of it is related to the OCI layer that we're adding. What do you think about the following suggestion:
Add a `pebble` module like the other comment (from me, heh) suggested. Then add a function like `add_pebble_services(target_dir, services, ...)` that will do the meat of this method. Then this method can be something like:

```python
def set_pebble_services(...):
    with tempfile.TemporaryDirectory() as td:
        tmpfs = Path(td)
        pebble.add_pebble_services(tmpfs, services, base_layer_dir)
        self.add_layer(tag, tmpfs)
```

tigarmo (commented):
thanks for the work! I only reviewed the code in `rockcraft/` so far; once we agree on the points and the tests and docs are updated I'll review them

rockcraft/project.py
```diff
@@ -143,6 +144,34 @@ def _validate_platform_set(cls, values: Mapping[str, Any]) -> Mapping[str, Any]:
...
+    user_id: Optional[int]
+    group: Optional[str]
+    group_id: Optional[int]
+    on_success: Optional[Literal["restart", "shutdown", "ignore"]]
+    on_failure: Optional[Literal["restart", "shutdown", "ignore"]]
+    on_check_failure: Optional[Dict[str, Literal["restart", "shutdown", "ignore"]]]
+    backoff_delay: Optional[str]
+    backoff_factor: Optional[float]
+    backoff_limit: Optional[str]
+
```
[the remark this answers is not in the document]
cjdcordeiro: not on purpose. it was added recently I think. I will add it

rockcraft/project.py
```diff
@@ -172,6 +202,19 @@ class Config:  # pylint: disable=too-few-public-methods
...
+    def _check_deprecated_options(cls, values: List) -> str:
+        """Before validation, check if deprecated fields exist. Exit if so."""
+        # pylint: disable=unused-argument
+        deprecation_msg = str(
+            "The fields 'entrypoint', 'cmd' and 'env are not supported in "
+            "Rockcraft. All ROCKs have Pebble as their entrypoint, so you must "
+            "use 'services' to define your container application and "
+            "respective environment."
+        )
+        raise ProjectValidationError(deprecation_msg)
```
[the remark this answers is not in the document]
tigarmo: You can have a validator with `pre=True` to run before other validation. The reason that I prefer this approach is that then those "invalid" attributes won't be part of the model at all, so there's no chance of a Rockcraft dev in the future trying to use e.g. `project.env` to check the project's environment. It's a maintenance benefit mostly, the end result for users is the same.

rockcraft/oci.py
```diff
@@ -215,34 +219,86 @@ def to_oci_archive(self, tag: str, filename: str) -> None:
...
+
+        new_layer_prefix = f"{(int(prefixes[-1]) + 1):03}" if prefixes else "001"
+        # TODO: we need proper name validation as Pebble is quite unforgiving
+        # about the layer's label. We also don't want to allow any character
+        # in the ROCK's name. For now, just replacing the common "_" with "-".
+        new_layer_name = f"{new_layer_prefix}-{name.replace('_', '-')}.yaml"
+
+        emit.progress(f"Writing new Pebble layer file {new_layer_name}")
+
+        tmpfs = Path(tempfile.mkdtemp())
```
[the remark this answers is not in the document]
cjdcordeiro: done

rockcraft/oci.py
```diff
@@ -215,34 +219,86 @@ def to_oci_archive(self, tag: str, filename: str) -> None:
...
+        pebble_layer_content = {
+            "summary": summary,
+            "description": description,
+            "services": services,
+        }
+
+        # To infer the right filename for this Pebble layer, we should check
+        # if there are existing Pebble layers in the base root filesystem,
+        # and increment on that.
+        # NOTE: the layer's filename prefix will always be "001-" when using
```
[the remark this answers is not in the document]
cjdcordeiro: it is a note for future self, if and when we decide to support other ROCKs as bases

rockcraft/oci.py
```diff
@@ -215,34 +219,86 @@ def to_oci_archive(self, tag: str, filename: str) -> None:
...
-            entrypoint is not defined, the first entry of cmd is the executable to
-            run when the container starts.
+    def set_pebble_services(
+        self,
+        services: Dict[str, Any],
+        name: str,
+        tag: str,
+        summary: str,
+        description: str,
+        base_layer_dir: Path,
```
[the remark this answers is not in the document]
cjdcordeiro: It is the base's filesystem, yes. 

Users could try to do that indeed. I propose we address that in a separate issue/PR, cause I believe that belongs to the broader story of "supporting user-defined Pebble". 

This is one of the items I have in the roadmap for us to discuss, so I'd rather push it for later instead of making a rush throw-away implementation right now. I've edited https://github.com/canonical/rockcraft/issues/65 to reference this discussion

rockcraft/lifecycle.py
cjdcordeiro: good point

cjdcordeiro: fixed

cjdcordeiro: oh actually no. `services` is optional, cause if you don't have it, the entrypoint will simply be `pebble enter`

rockcraft/oci.py
cjdcordeiro: done

rockcraft/project.py
```diff
@@ -509,6 +552,8 @@ def _add_pebble_data(yaml_data: Dict[str, Any]) -> None:
     pebble_part_spec = {
         "plugin": "nil",
         "stage-snaps": ["pebble/latest/edge"],
-        "stage": ["bin/pebble"],
+        "stage": [Image._PEBBLE_BINARY_PATH],  # pylint: disable=protected-access
```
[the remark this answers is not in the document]
cjdcordeiro: done

rockcraft/project.py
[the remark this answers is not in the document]
cjdcordeiro: done

rockcraft/project.py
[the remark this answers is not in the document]
cjdcordeiro: done

rockcraft/oci.py
```diff
@@ -215,34 +216,56 @@ def to_oci_archive(self, tag: str, filename: str) -> None:
...
+
+        pebble = Pebble()
+        with tempfile.TemporaryDirectory() as tmpfs:
+            tmpfs_path = Path(tmpfs)
+            pebble.define_pebble_layer(
+                tmpfs_path, base_layer_dir, pebble_layer_content, name
+            )
+
+        emit.progress("Writing new Pebble layer file")
+        self.add_layer(tag, tmpfs_path)
```
tigarmo, line 267: something is weird here; the `TemporaryDirectory()` context manager is supposed to remove the directory on cleanup, so how are we still able to access it here??
(imo these two lines should be inside the context manager, and the `shutil.rmtree()` call can be removed.

rockcraft/pebble.py
```diff
@@ -0,0 +1,84 @@
...
+    """Handles the Pebble metadata information and setup operations."""
+
+    PEBBLE_PATH = "var/lib/pebble/default"
+    PEBBLE_LAYERS_PATH = f"{PEBBLE_PATH}/layers"
+    PEBBLE_BINARY_PATH = "bin/pebble"
+    PEBBLE_PART_SPEC = {
+        "plugin": "nil",
+        "stage-snaps": ["pebble/latest/edge"],
+        "stage": [PEBBLE_BINARY_PATH],
+        "override-prime": f"craftctl default\nmkdir -p {PEBBLE_LAYERS_PATH}",
```
tigarmo, line 36: So if the actual pebble layer file is written outside of the parts lifecycle, in `define_pebble_layer()`, is there a need for this part to create this path? Doesn't `define_pebble_layer()` do it too, on line 73?

rockcraft/pebble.py
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    def define_pebble_layer(
+        self, target_dir: Path, ref_fs: Path, layer_content: dict, rock_name: str
```
tigarmo, line 40:
```suggestion
        self, target_dir: Path, ref_fs: Path, layer_content: Dict[str, Any], rock_name: str
```

rockcraft/project.py
```diff
@@ -205,18 +209,22 @@ class Config:  # pylint: disable=too-few-public-methods
         allow_population_by_field_name = True
         alias_generator = lambda s: s.replace("_", "-")  # noqa: E731
 
-    @pydantic.validator("cmd", "entrypoint", "env", pre=True)
+    @pydantic.root_validator(pre=True)
     @classmethod
-    def _check_deprecated_options(cls, values: List) -> str:
-        """Before validation, check if deprecated fields exist. Exit if so."""
+    def _check_unsupported_options(cls, values: Mapping[str, Any]) -> Mapping[str, Any]:
```
tigarmo, line 214: :+1:

tests/unit/test_pebble.py
```diff
@@ -0,0 +1,176 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright 2021-2022 Canonical Ltd.
```
tigarmo, line 3:
```suggestion
# Copyright 2023 Canonical Ltd.
```

tests/unit/test_project.py
```diff
@@ -50,13 +57,7 @@
 summary: "example for unit tests"
 description: "this is an example of a rockcraft.yaml for the purpose of testing rockcraft"
 license: Apache-2.0
-entrypoint: ["/bin/hello"]
-cmd: ["world"]
-env:
-    - NAME: "VALUE"
-package-repositories:
```
tigarmo, line 57: wait why was the `package-repositories` entry removed?

tests/unit/test_project.py
```diff
@@ -106,50 +113,28 @@ def test_project_unmarshal(check, yaml_loaded_data):
...
-
-
-def test_unmarshal_undefined_repositories(yaml_loaded_data):
-    del yaml_loaded_data["package-repositories"]
-
-    project = Project.unmarshal(yaml_loaded_data)
-
-    assert project.package_repositories is None
-
+    assert project.rock_license == "Apache-2.0"
```
tigarmo: please keep the `check.equal()` calls (unless there was a specific reason to remove them?)

tests/unit/test_project.py
```diff
@@ -106,50 +113,28 @@ def test_project_unmarshal(check, yaml_loaded_data):
...
+        }
+    }
 
-def test_unmarshal_invalid_repositories(yaml_loaded_data):
-    yaml_loaded_data["package-repositories"] = [{}]
 
-    with pytest.raises(ProjectValidationError) as error:
-        Project.unmarshal(yaml_loaded_data)
+@pytest.mark.parametrize(
+    "deprecated_field",
```
tigarmo, line 130: can we refer to those fields as "removed" instead of "deprecated"?

tests/unit/test_pebble.py
```diff
@@ -0,0 +1,176 @@
...
+        mocked_data = {"writes": ""}
+
+        def mock_write(s):
+            mocked_data["writes"] += s
+
+        m = mock_open()
+        pebble_obj = Pebble()
+        with patch("builtins.open", m) as f:
+            with patch("pathlib.Path.mkdir") as local_mock_mkdir:
+                m.return_value.write = mock_write
```
tigarmo, line 163: I understand that opinions on this diverge, but I think this is too much mocking. Can't we just call the method on a temp path and check the contents afterwards?

docs/tutorials/code/chisel/task.yaml
```diff
@@ -23,11 +23,11 @@ execute: |
   # [docs:skopeo-copy-end]
 
   # [docs:docker-run]
-  docker run --rm  chisel-openssl
+  docker run --rm chisel-openssl exec openssl
   # [docs:docker-run-end]
 
   # [docs:docker-run-with-args]
-  docker run --rm chisel-openssl s_client -connect ubuntu.com:443 -brief
+  docker run --rm chisel-openssl exec --env=SSL_CERT_FILE=/etc/ssl/certs/ca-certificates.crt openssl s_client -connect ubuntu.com:443 -brief
```
tigarmo, line 30: Hmm. Since this is a tutorial, I'm wondering about the best way to do this. Instead of `exec --env...`, wouldn't it be preferable to define an `openssl` pebble service with the `SSL_CERT_FILE` env var?
Is the issue passing the remaining parameters to the `openssl` service?

docs/tutorials/code/hello-world/task.yaml
```diff
@@ -31,10 +31,10 @@ execute: |
   # [docs:skopeo-copy-end]
 
   rm hello_1.0_amd64.rock
-  docker images
+  docker images hello:1.0
 
   # [docs:docker-run]
-  docker run --rm hello:1.0
+  docker run --rm hello:1.0 exec hello -t
```
tigarmo, line 37: this one could be a service right?

tigarmo (changes requested):
I did another pass at the code in rockcraft/, and also reviewed the tests and docs. Many thanks for all this work meu amigo ;)

tigarmo:
Here's a question for @sergiusens and @cjdcordeiro ; when we merge this PR, everyone's rockcraft.yaml files will be broken. Do we want to pre-warn the users, via I don't know Mattermost, a few days in advance?  It's not like they can do anything until the update drops, but at least they won't be caught completely by surprise.

cjdcordeiro:
> Here's a question for @sergiusens and @cjdcordeiro ; when we merge this PR, everyone's rockcraft.yaml files will be broken. Do we want to pre-warn the users, via I don't know Mattermost, a few days in advance? It's not like they can do anything until the update drops, but at least they won't be caught completely by surprise.

I'd say so. More than that, it would probably be worth creating a `legacy` track for those who absolutely need to keep using the old version for a bit longer. @sergiusens wdyt?

rockcraft/oci.py
cjdcordeiro: hum...true.

I guess the ongoing spread tests will tell us whether that is the case or not. 

I'm applying those changes anyway

rockcraft/pebble.py
cjdcordeiro: that sounds accurate since line 74 is using `parents=True`. Changing it

tests/unit/test_project.py
cjdcordeiro: yes that has been updated already. please re-check

tests/unit/test_project.py
cjdcordeiro: yep

docs/tutorials/code/chisel/task.yaml
cjdcordeiro: yes you can't do this today. It is a Pebble limitation. whether it is by design or not, we'll have to take it internally for further discussions

docs/tutorials/code/hello-world/task.yaml
cjdcordeiro: yep it could. It was re-written this way because Pebble doesn't handle well services with a very short lifetime, so if that was a service, the container would keep running after the printout

tests/unit/test_project.py
cjdcordeiro: jeez...I've had a lot of conflicts when rebasing with main. Hopefully the last commit does it

rockcraft/pebble.py
tigarmo: so how come this stayed? Was there a problem?

cjdcordeiro: Yes, I forgot (again lol) that `services` is optional. So if there are no `services`, the `set_pebble_services` will not run, and this path won't exist, which will make `pebble enter` complain about it. we could set `--create-dirs` within the OCI entrypoint, but this is cleaner.

tigarmo (approved):

rockcraft/project.py
```diff
@@ -144,6 +145,41 @@ def _validate_platform_set(cls, values: Mapping[str, Any]) -> Mapping[str, Any]:
...
+    command: str
+    summary: Optional[str]
+    description: Optional[str]
+    startup: Optional[Literal["enabled", "disabled"]]
+    after: Optional[List[str]]
+    before: Optional[List[str]]
+    requires: Optional[List[str]]
+    environment: Optional[Dict[str, str]]
+    user: Optional[str]
+    user_id: Optional[int] = pydantic.Field(alias="user-id")
```
cmatsuoka, line 165: A model-wide alias can be defined to replace underscores instead of doing it in each field (see `alias_generator`).

cmatsuoka (approved):
Lgtm, added a comment about the Pydantic alias generator.

PR #221, opened 2023-03-29
spread: don't remove the base instance in cleanup

tigarmo:
The original `restore-each` removes all instances, including the base one! With this fix we should be able to reuse the base instance between compatible tests.

### DIFF
--- a/rockcraft/oci.py
+++ b/rockcraft/oci.py
@@ -43,6 +43,9 @@ ECR_URL = "public.ecr.aws/ubuntu"
 
 REGISTRY_URL = ECR_URL
 
+# The number of times to try downloading an image from `REGISTRY_URL`.
+MAX_DOWNLOAD_RETRIES = 5
+
 
 @dataclass(frozen=True)
 class Image:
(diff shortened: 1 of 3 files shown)


tigarmo:
Ran this five times after adding the `--retry-times` option in the `skopeo copy` call, all passed. Runtime went from about 55 minutes to about 37, so pretty nice! I can split this PR into two if you feel the changes are too unrelated
(cc @cjdcordeiro FYI)

rockcraft/oci.py
```diff
@@ -313,8 +325,18 @@ def set_annotations(self, annotations: Dict[str, Any]) -> None:
...
+    destination: str,
+    *system_params: str,
+    copy_params: Optional[List[str]] = None,
+) -> None:
+    """Transfer images from source to destination.
+
+    :param copy_params: Extra parameters to pass specifically to the "copy"
+        skopeo command (and not to skopeo itself).
+    """
+    copy_extra = copy_params if copy_params else []
```
mr-cal, line 339: Would you be able to get rid of this line if you changed the parameter from
`copy_params: Optional[List[str]] = None` to `copy_params: List[str] = []`?

mr-cal (approved):
Nice work!!

rockcraft/oci.py
tigarmo: Yes, but I thought our "standard" regarding optional parameters was to have `None` instead (I'm not sure on that one actually)

mr-cal: Interesting, I hadn't thought about that.  Now I had to look it up:

Occurrences of `<param>: Optional[List...] = None` in snapcraft, charmcraft, and rockcraft = 14
Occurrences of `<param>: List[...] = []` in snapcraft, charmcraft, and rockcraft = 45

I'm fine with either choice.  If you have a preference, I will add it to my list of team standards, which is probably ready to be published somewhere for the whole team.

sergiusens: the potential problems with mutables exist when assigning to a paramater. So None is the safe choice

- https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

sergiusens (approved):

Issue #194, opened 2023-02-14
Unable to import rock into the docker daemon when packing with python plugin and staging python3-dev


arturo-seijas:
Hi!

I've just noticed `skopeo` fails to import the resulting rock when packing with the python plugin:
```
name: indico
summary: Indico rock
description: Indico OCI image for the Indico charm
version: "1.0"
base: ubuntu:22.04
build-base: ubuntu:22.04
license: Apache-2.0
platforms:
  amd64:
parts:
  indico:
    stage-packages:
      - python3-dev
      - python3-venv
    source: plugins
    plugin: python
```
`skopeo --insecure-policy copy oci-archive:indico_1.0_amd64.rock docker-daemon:localhost:32000/indico:latest` seems to execute successfully but when I run `docker image ls` there are no results. However, it works if I change the plugin:
```
name: indico
summary: Indico rock
description: Indico OCI image for the Indico charm
version: "1.0"
base: ubuntu:22.04
build-base: ubuntu:22.04
license: Apache-2.0
platforms:
  amd64:
parts:
  indico:
    stage-packages:
      - python3-dev
      - python3-venv
    source: plugins
    plugin: nil
```
After some changes in my rockcraft.yaml to provide a narrower use case, I noticed the following does work with the python plugin, so this likely has to do with the staged packages, either python3-dev or pythone-venv:
```
name: indico
summary: Indico rock
description: Indico OCI image for the Indico charm
version: "1.0"
base: ubuntu:22.04
build-base: ubuntu:22.04
license: Apache-2.0
platforms:
  amd64:
parts:
  indico:
    stage-packages:
      - build-essential
    source: plugins
    plugin: python
```
Let me know if there's anything else you'd like me to provide in order to pinpoint the issue.

Thank you

tigarmo:
The issue here looks like a bad interaction between our `usrmerge` handling and craft-parts' Python plugin. The plugin creates a virtual env where things like `python-packages` and the python code get installed; this creates a `bin/python3` symlink pointing to `/usr/bin/python3.10`. 

However, in the ubuntu base `/bin` is a symlink to `/usr/bin`, so `/bin/python3` becomes `/usr/bin/python3` which already exists on the base. There's clearly a bug there where rockcraft isn't taking into account the fact that `/usr/bin/python3` already exists on the base, but you can see the original well-intentioned handling on the log:

```
[...]

PR #222, opened 2023-03-29
plugins: add rockcraft specific python plugin

tigarmo:
The specialized plugin has logic to support bare and ubuntu bases.

The diff is quite big but most of that is from test updates.

(CRAFT-1664)


tigarmo:
Closes #194

sergiusens (approved):
Looks really good

Issue #228, opened 2023-04-06
Conflicting files when packing a ROCK with overlay-packages on an ubuntu base


tigarmo:
[This commit](https://github.com/canonical/rockcraft/commit/c3382fc2408e3e2bae964e0622e50798c46a2e2f) added handling for multiple paths ending up as the same, "final" one when the usrmerge is taken into account. Currently we accept conflicting _directories_, but not _files_ because we didn't know a "live" use-case for conflicting files. The following is a reduced example of a real rockcraft.yaml:

```
parts:
  overlay-pkgs:
    plugin: nil
    overlay-packages: [libexpat1]

  regular-pkgs:
    plugin: nil
    stage-packages: [libexpat1]
```

Error message:

```
Creating new layer                                                                                                                                                                          
Conflicting paths pointing to 'usr/lib/x86_64-linux-gnu/libexpat.so.1': /root/prime/lib/x86_64-linux-gnu/libexpat.so.1, /root/prime/usr/lib/x86_64-linux-gnu/libexpat.so.1                  
craft-providers error: Failed to execute pack in instance
```

The `libexpat1` package provides `lib/x86_64-linux-gnu/libexpat.so.1`. My initial reasoning for the root cause of the bug is that, because the overlay is on top of an ubuntu base, `lib/x86_64-linux-gnu/libexpat.so.1` becomes `usr/lib/x86_64-linux-gnu/libexpat.so.1` (because of the usrmerge), which then conflicts with the `lib/x86_64-linux-gnu/libexpat.so.1` file provided by the `regular-pkgs` part.

The files are identical (as indicated by `md5sum`), so in this case it seems reasonable to accept conflicts _if_ all paths are identical files.

(cc @jnsgruk)

tigarmo:
Fixed by #231

PR #231, opened 2023-04-17
oci: support file collisions if the files match

tigarmo:
In c3382fc2408 we added support for directory collisions caused by the usrmerge when creating the payload layer; now we add support for file collisions if the files match in attributes and contents. This scenario can happen if an Ubuntu package is used both as an overlay-package and as a stage-package.

-----


cmatsuoka (approved):

mr-cal (approved):

tigarmo:
This fixes #228

PR #233, opened 2023-04-18
feat: add run-user

### LINKED ISSUE #235: [bug] Pebble layer is being written with service fields in snake case
https://github.com/canonical/rockcraft/blob/c8d27b6e885a7ab08166920b85194eea8edd599b/rockcraft/pebble.py#L85

See an example of the error here: https://github.com/canonical/nginx-rock/pull/6#issuecomment-1513416783

## Goal

 - replace all yaml keys underscored with "-"
 - add a spread test for it

cjdcordeiro:
Support the use of a nonroot default OCI user. When provided and if it is a supported global user, the run-user will be created in the ROCK and set as the default OCI user.

Fixes https://github.com/canonical/rockcraft/issues/235
Fixes https://github.com/canonical/rockcraft/issues/238

## In this PR...
 - created `usernames` module to host the hardcoded mapping of supported global usernames, analogously to https://github.com/snapcore/snapd/blob/master/snap/system_usernames.go#L29. Only the `_daemon_` user is supported for now, with UID 584792
 - `run-user` field added to rockcraft.yaml, supporting only the values fixed in the `username` module
 - reset the ROCK's default OCI user whenever `run-user` is provided, otherwise keep it as `root`
 - update the reference docs to include the new optional field `run-user`
 - make the Rockcraft-generated content (Pebble paths and .rock metadata file) accessible to all users and not just `root`
 - add spread and unit tests

Additionally:
 - fix #235 by making sure the project attributes are converted to a dictionary while keeping the original kebab-case alias names
 - remove duplication of code by adding an alias_generator to every pydantic class

## Notes

 - we're now good to go with the `_daemon_` user.

-----


tests/spread/general/run-user/task.yaml
```diff
@@ -0,0 +1,24 @@
...
+  docker rmi --force run-user-test
+  # Install container
+  sudo /snap/rockcraft/current/bin/skopeo --insecure-policy copy oci-archive:run-user-test_latest_amd64.rock docker-daemon:run-user-test:latest
+  # Ensure container exists
+  docker images run-user-test | MATCH "run-user-test"
+  docker inspect run-user-test --format '{{.Config.User}}' | MATCH "confined_daemon"
+  
+  # ensure container doesn't exist
+  docker rm -f run-user-test-container
+  docker run --rm --entrypoint /bin/sh run-user-test -c 'whoami' | MATCH "confined_daemon"
```
tigarmo, line 19: what about `docker run --rm run-user-test exec -c 'whoami'`? To ensure that pebble itself is running as `confined_daemon` correctly

tests/unit/test_oci.py
```diff
@@ -547,6 +547,86 @@ def test_add_layer_duplicate_files(self, tmp_path, temp_tar_contents):
...
+    ):
+        mock_tag = "tag"
+        fake_tmpfs = tmp_path / "mock-tmp"
+        mock_tmpdir.return_value = fake_tmpfs
+
+        image = oci.Image("a:b", Path("/c"))
+        image.add_user(tmp_path, mock_tag, "foo", 585287)
+
+        mock_tmpdir.assert_called_once()
+        with open(fake_tmpfs / "etc/passwd") as f:
```
tigarmo, line 565: do you mind swapping these for 
```python
    check.equal(
        (fake_tmpfs / "etc/passwd").read_text(),
        "foo:x:585287:585287::/var/lib/pebble/default:/usr/bin/false\n",
    )

File: tests/unit/test_oci.py [hunk 3]
```diff
@@ -547,6 +547,86 @@ def test_add_layer_duplicate_files(self, tmp_path, temp_tar_contents):
...
+        self,
+        check,
+        mock_add_layer,
+        mock_tmpdir,
+        tmp_path,
+    ):
+        # Mock the existence of users in the base image
+        fake_etc = tmp_path / "etc"
+        fake_etc.mkdir(parents=True, exist_ok=True)
+        with open(fake_etc / "passwd", "w") as f:
```
tigarmo, line 586: same here re: `read_text()`

tigarmo (changes requested):
thanks for all this work!

rockcraft/oci.py
```diff
@@ -189,6 +190,71 @@ def add_layer(
...
+
+                # only add the shadow file is there's already one in the base image
+                with open(Path(tmpfs) / shadow_file, "a+") as shadowf:
+                    shadowf.write(
+                        (base_layer_dir / shadow_file).read_text()
+                        + f"{username}:!:{days_since_epoch}::::::\n"
+                    )
+
+            emit.progress(
+                "Adding user {username}:{uid} with group {new_group}:{new_gid}"
```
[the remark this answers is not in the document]
cjdcordeiro: good catch! Fixed

rockcraft/oci.py
```diff
@@ -312,6 +388,7 @@ def set_control_data(self, metadata: Dict[str, Any]) -> None:
         rock_metadata_file = control_data_rock_folder / "metadata.yaml"
         with rock_metadata_file.open("w", encoding="utf-8") as rock_meta:
             yaml.dump(metadata, rock_meta)
+        rock_metadata_file.chmod(0o755)
```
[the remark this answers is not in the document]
cjdcordeiro: this is too make sure any nonroot user can manipulate the file, but given that this is a text file, I guess 755 is an overkill. Changing to 644

rockcraft/project.py
```diff
@@ -191,9 +198,8 @@ class Project(pydantic.BaseModel):
     version: str
     platforms: Dict[str, Any]
     base: Literal["bare", "ubuntu:18.04", "ubuntu:20.04", "ubuntu:22.04"]
-    build_base: Optional[
-        Literal["ubuntu:18.04", "ubuntu:20.04", "ubuntu:22.04"]
-    ] = pydantic.Field(alias="build-base")
+    build_base: Optional[Literal["ubuntu:18.04", "ubuntu:20.04", "ubuntu:22.04"]]
+    run_user: Optional[Literal[tuple(SUPPORTED_GLOBAL_USERNAMES)]]  # type: ignore
```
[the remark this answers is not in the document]
cjdcordeiro: `pyright` complaing about the use of `tuple`:

```
rockcraft/project.py:202:32 - error: Type arguments for "Literal" must be None, a literal value (int, bool, str, or bytes), or an enum value
```

tests/spread/general/run-user/task.yaml
cjdcordeiro: sure

tests/unit/test_oci.py
cjdcordeiro: :+1:

rockcraft/oci.py
```diff
@@ -189,6 +190,71 @@ def add_layer(
...
+        ):
+            raise errors.RockcraftError(
+                str(
+                    f"Error while trying to create user {username}:{uid}, "
+                    f"with group {username}:{uid}...\n"
+                    " - conflict with existing user/group in the base filesystem"
+                )
+            )
+
+        existing_users += (
```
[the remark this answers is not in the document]
cjdcordeiro: right...good point. any suggestions? I can't think of a clean way right now...this might be tricky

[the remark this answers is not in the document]
cjdcordeiro: @tigarmo here's the update: https://github.com/canonical/rockcraft/pull/233/commits/bfbd7fcd294d969db3eb7e9c874e05b40169ede4

rockcraft/oci.py
```diff
@@ -190,34 +200,44 @@ def add_layer(
...
-            if (base_layer_dir / group_file).exists()
-            else ""
-        )
+        user_files = {"passwd": "", "group": "", "shadow": ""}
+
+        prime_dir_etc = prime_dir / "etc"
+        base_layer_dir_etc = base_layer_dir / "etc"
+        # Being cautious about possible changes (edits or removals) in
+        # /etc/{passwd,group,shadow} done by the user through overlay scripts.
+        # Basically:
```
tigarmo, line 223: :+1:

tigarmo: I think this merits a spread test, to ensure that we support the common case of people adding users in overlays

cjdcordeiro: sure. added in https://github.com/canonical/rockcraft/pull/233/files#diff-6adf14346ad84f5632a4ea9fbab1d93df22fbb9b16531f4261bc4a15f792cedfR23

rockcraft/oci.py
```diff
@@ -189,6 +200,83 @@ def add_layer(
...
+        for u_file in user_files:
+            if (prime_dir_etc / u_file).exists():
+                user_files[u_file] = (prime_dir_etc / u_file).read_text()
+            elif (base_layer_dir_etc / u_file).exists() and not (
+                prime_dir_etc / f".wh.{u_file}"
+            ).exists():
+                user_files[u_file] = (base_layer_dir_etc / u_file).read_text()
+
+        if (
+            f"{username}:" in user_files["passwd"]
```
lengau, line 240: Do we care about the edge case of an existing username that ends with the added username?

e.g. adding the user `daemon` but the passwd file already contains a `somedaemon` user?

(Genuine question and I presume not, but if we do we might want to ensure that the username immediately follows a newline.)

rockcraft/oci.py
```diff
@@ -189,6 +200,83 @@ def add_layer(
...
+                passwdf.write(user_files["passwd"])
+
+            with open(tmpfs_etc / "group", "a+") as groupf:
+                groupf.write(user_files["group"])
+
+            if user_files["shadow"]:
+                days_since_epoch = (datetime.utcnow() - datetime(1970, 1, 1)).days
+
+                # only add the shadow file if there's already one in the base image
+                with open(tmpfs_etc / "shadow", "a+") as shadowf:
```
lengau, line 271: Do we need specific permissions on this file?

tests/spread/general/run-user/rockcraft.yaml
```diff
@@ -0,0 +1,23 @@
...
+    on-failure: shutdown
+platforms:
+  amd64:
+
+parts:
+  my-part:
+    plugin: nil
+    overlay-script: |
+      set -x
+      useradd -R $CRAFT_OVERLAY -M -U -r test-user
```
lengau, line 23:
```suggestion
      useradd -R $CRAFT_OVERLAY -M -U -r test-user
      
```
(adds newline at the end of the file :slightly_smiling_face: )

lengau (approved):
Looks good to me! I suggested one cosmetic change and the other two comments are questions — if the answer is "no" then I'm happy with it as is!

rockcraft/oci.py
cjdcordeiro: we only append something to this file if it already exists, and afaik the existing file permissions are preserved, so we shouldn't need to touch that.

tests/spread/general/run-user/rockcraft.yaml
cjdcordeiro: ;)

rockcraft/oci.py
cjdcordeiro: nice catch. that's a good corner case to cover indeed. I've improved the conflict checking in https://github.com/canonical/rockcraft/pull/233/commits/e9db7fbfc6c6615e2961636f4c47bfe455b204ab

sed-i:
Hey @cjdcordeiro @lengau,
Does `run-user` replace the `non-root-user` part we see sprinkled here and there?

```yaml
  non-root-user:
    plugin: nil
    after: [ otelcol ]
    overlay-script: |
      # Create a user in the $CRAFT_OVERLAY chroot
      groupadd -R $CRAFT_OVERLAY -g 1001 ubuntu
      useradd -R $CRAFT_OVERLAY -M -r -u 1001 -g ubuntu ubuntu
```

PR #243, opened 2023-04-27
make pebble write service logs to stdout by default

simskij:

-----

Pebble currently does not show any logs from the service, making it impossible to know whether the service really works or not without running `pebble logs`. True to convention, we should make services log to stdout by default. 

Closes https://github.com/canonical/cos-alerter/issues/37


cjdcordeiro (approved):
lgtm. The potential conflict with someone passing an additional `-v` at container deployment doesn't seem to be a reality, since there's already a [spread test](https://github.com/canonical/rockcraft/blob/main/tests/spread/general/bare-base/task.yaml#L14) doing that and it is passing. Thanks for the contribution ;)

tigarmo (approved):
thanks!

PR #244, opened 2023-04-27
project: constrain ROCK name to pebble's rules

tigarmo:
Pebble only accepts lowercase a-z letters, numbers and hyphens. Since we use the ROCK's name on the layer file, use the same regex.

-----

### DIFF
--- a/rockcraft/pebble.py
+++ b/rockcraft/pebble.py
@@ -69,10 +69,7 @@ class Pebble:
         )
 
         new_layer_prefix = f"{(int(prefixes[-1]) + 1):03}" if prefixes else "001"
-        # TODO: we need proper name validation as Pebble is quite unforgiving
-        # about the layer's label. We also don't want to allow any character
-        # in the ROCK's name. For now, just replacing the common "_" with "-".
-        new_layer_name = f"{new_layer_prefix}-{rock_name.replace('_', '-')}.yaml"
+        new_layer_name = f"{new_layer_prefix}-{rock_name}.yaml"
 
         emit.progress(f"Preparing new Pebble layer file {new_layer_name}")
 
--- a/rockcraft/project.py
+++ b/rockcraft/project.py
@@ -18,6 +18,7 @@
 
 import operator
 import platform as host_platform
+import re
 from functools import reduce
 from typing import (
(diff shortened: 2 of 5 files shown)


rockcraft/pebble.py
```diff
@@ -69,10 +69,7 @@ def define_pebble_layer(
         )
 
         new_layer_prefix = f"{(int(prefixes[-1]) + 1):03}" if prefixes else "001"
-        # TODO: we need proper name validation as Pebble is quite unforgiving
```
tigarmo, line 72: In the end I followed pebble's direction here but we should consider the `_` case: if it's too common maybe we should still support it?

mr-cal (approved):
Looks good to me!

rockcraft/project.py
```diff
@@ -180,10 +181,33 @@ class Service(pydantic.BaseModel):
     kill_delay: Optional[str] = pydantic.Field(alias="kill-delay")
 
 
+NAME_REGEX = r"^([a-z](?:-?[a-z0-9]){2,})$"
```
cjdcordeiro, line 184: for future reference, this is the regular expression for the OCI-compliant image name and tag (the latter could be applied to rockcraft.yaml's `version` field): https://github.com/opencontainers/distribution-spec/blob/main/spec.md#pulling-manifests

rockcraft/project.py
```diff
@@ -242,7 +269,7 @@ def _validate_license(cls, rock_license: str) -> str:
     def _validate_title(cls, title: Optional[str], values: Mapping[str, Any]) -> str:
         """If title is not provided, it defaults to the provided ROCK name."""
         if not title:
-            title = values["name"]
+            title = values.get("name", "")
```
cjdcordeiro, line 272: Why does this validatory need to run `always`? Since `name` is mandatory, this `.get` looks more of a workaround to avoid a false positive exception...

cjdcordeiro (approved):
With just a couple of comments for clarification, this lgtm ;) thanks

rockcraft/project.py
sergiusens: title can be validated before name

PR #248, opened 2023-05-04
plugins: add sitecustomize.py in PythonPlugin's build

tigarmo:
The issue: running a ROCK that was built with the Python plugin by
calling "python3" or "/usr/bin/python3" (as opposed to "/bin/python3")
fails to find pip-installed packages.

The solution implemented with this commit is to add a sitecustomize
module that always adds the location of the pip-installed packages to
sys.path. This ensures that the pip-installed packages are found even
if the "global" Python interpreter is called.

CRAFT-1721

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -18,7 +18,7 @@ commonmark==0.9.1
 coverage==6.5.0
 craft-archives==0.0.3
 craft-cli==1.2.0
-craft-parts==1.19.3
+craft-parts==1.20.0
 craft-providers==1.8.0
 cryptography==38.0.4
 Deprecated==1.2.13
--- a/rockcraft/plugins/python_plugin.py
+++ b/rockcraft/plugins/python_plugin.py
@@ -17,12 +17,42 @@
 """The Rockcraft Python plugin."""
 
 import logging
-from typing import Optional
+from textwrap import dedent
+from typing import List, Optional
 
 from craft_parts.plugins import python_plugin
 from overrides import override
 
 logger = logging.getLogger(__name__)
+
+# Template for the sitecustomize module that we'll add to the payload so that
+# the pip-installed packages are found regardless of how the interpreter is
+# called.
+SITECUSTOMIZE_TEMPLATE = dedent(
+    """
+    # sitecustomize added by Rockcraft.
+    import site
+    import sys
+
+    major, minor = sys.version_info.major, sys.version_info.minor
+    site_dir = f"/lib/python{major}.{minor}/site-packages"
+    dist_dir = "/usr/lib/python3/dist-packages"
+
+    # Add the directory that contains the venv-installed packages.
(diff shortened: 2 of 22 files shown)


requirements-dev.txt
```diff
@@ -18,7 +18,7 @@ commonmark==0.9.1
 coverage==6.5.0
 craft-archives==0.0.3
 craft-cli==1.2.0
-craft-parts==1.19.3
+craft-parts==1.20.0
```
tigarmo, line 21: this bump will be removed before merging (it's being addressed in a separate PR). Used this here because I needed a Python plugin fix from that version

rockcraft/plugins/python_plugin.py
```diff
@@ -17,13 +17,25 @@
...
 
 logger = logging.getLogger(__name__)
 
+# Template for the sitecustomize module that we'll add to the payload so that
+# the pip-installed packages are found regardless of how the interpreter is
+# called.
+SITECUSTOMIZE_TEMPLATE = dedent(
+    """
+    import site
+    site.addsitedir("/lib/${py_version}/site-packages")
```
sergiusens, line 34: an alternative to the template would be to import sys here and use something like `sys.version_info.minor`
```suggestion
    import sys
    site.addsitedir(f"/lib/{sys.version_info.major}.{sys_version.minor}/site-packages")
```

tests/integration/plugins/test_python_plugin.py
```diff
@@ -33,6 +34,7 @@
 
 BARE_BASES = {"bare"}
 UBUNTU_BASES = set(ALL_BASES) - BARE_BASES
+# UBUNTU_BASES = ["ubuntu:22.04"]
```
sergiusens, line 37: what's this?

tigarmo: leftover code that I forgot to remove
fixed

tigarmo:
new sitecustomize.py template is failing in spread tests, marking this as a draft again to take a look

tigarmo:
moving this to draft _again_ to fix the order of dist-packages/site-packages

rockcraft/plugins/python_plugin.py
```diff
@@ -17,13 +17,43 @@
...
+    dist_dir = "/usr/lib/python3/dist-packages"
+
+    # Add the directory that contains the venv-installed packages.
+    site.addsitedir(site_dir)
+
+    # Make sure that this site-packages dir comes *before* the base-provided
+    # dist-packages dir in sys.path.
+    path = sys.path
+    site_index = path.index(site_dir)
+    dist_index = path.index(dist_dir)
```
sergiusens, line 48: you probably want to catch a ValueError, I don't think self compiled python has a dist_dir.

sergiusens (approved):

sergiusens (approved):

PR #252, opened 2023-05-08
Add HTG for converting a dockerfile

sed-i:
This PR adds a how-to guide for converting a multipart dockerfile to a rockcraft.
This is a summary of the work done for https://github.com/canonical/kubeflow-rocks/pull/6.

Edits welcome.


cjdcordeiro (commented):
Thanks for the kind contribution @sed-i :pray: 

IMO, this overlaps with the [existing tutorial](https://canonical-rockcraft.readthedocs-hosted.com/en/latest/tutorials/migrate-to-chiselled-rock.html) on the same topic.

Furthermore, although the mental mapping of a Dockerfile state to a Rockcraft part is usually a good starting point, I wouldn't make it a reference mechanism as more often than not, the resulting rockcraft.yaml will have a different number of `parts` than the original Dockerfile (eg. the aforementioned tutorial).

If the existing tutorial is somehow incomplete, I'd rather complement it instead of adding a new guide ;)

sed-i:
This is quite dated. Closing.

PR #255, opened 2023-05-12
docs: add chisel and rock explanations

dboddie:

-----

### DIFF
--- a/docs/_static/package-slices.svg
+++ b/docs/_static/package-slices.svg
@@ -1,0 +1,66 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+
+<svg xmlns="http://www.w3.org/2000/svg" height="360" width="500"
+     version="1.1" viewBox="0 0 500 360">
+
+<rect x="0" y="0" width="500" height="360" fill="none" />
+
+<g transform="translate(0, 0)">
+<rect x="10" y="10" width="480" height="150" stroke="black" fill="#c0c0c0" />
+
+<text x="250" y="35" font-family="Ubuntu" font-size="18"
+	  text-anchor="middle" dominant-baseline="text-after-edge"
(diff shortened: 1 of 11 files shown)


cjdcordeiro (changes requested):
thanks for this nice refactoring. 

I'm sorry for making this process a bit more tedious, but I need to make the following request, as I believe that's the best way to address such significant changes:
 - could you please split this PR in 2?
    - one PR for the refactoring, without any content changes
    - another PR, depending on the 1st one, for the content changes

The reason I'm asking for this is that it is really hard to review this PR without having a reference for comparison. I can see a few things that need changes (eg. the Chisel definition is misleading, as it's treating it as a container tool; the new svg static image is missing an arrow; old static images are left unused in the repo...), but I can only see this by having this PR and https://canonical-rockcraft.readthedocs-hosted.com/en/latest/explanation.html#what-is-chisel side by side, which is error-prone as I might miss small changes.

Sorry again

dboddie:
> thanks for this nice refactoring.
> 
> I'm sorry for making this process a bit more tedious, but I need to make the following request, as I believe that's the best way to address such significant changes:
> 
>     * could you please split this PR in 2?
>       
>       * one PR for the refactoring, without any content changes
>       * another PR, depending on the 1st one, for the content changes

No problem,  @cjdcordeiro. Thanks for the pre-review. :)

> The reason I'm asking for this is that it is really hard to review this PR without having a reference for comparison. I can see a few things that need changes (eg. the Chisel definition is misleading, as it's treating it as a container tool; the new svg static image is missing an arrow; old static images are left unused in the repo...), but I can only see this by having this PR and https://canonical-rockcraft.readthedocs-hosted.com/en/latest/explanation.html#what-is-chisel side by side, which is error-prone as I might miss small changes.

I'll close this PR and create a new one for the refactoring.

PR #262, opened 2023-05-17
tests: install docker after lxd

mr-cal:

-----

Test to install docker after LXD.

### DIFF
--- a/spread.yaml
+++ b/spread.yaml
@@ -56,12 +56,6 @@ prepare: |
   # The /snap directory does not exist in some environments
   [ ! -d /snap ] && ln -s /var/lib/snapd/snap /snap
 
-  # install docker before lxd, otherwise it will mess up with lxd networking
-  snap install docker
-
-  # make sure docker is working
-  retry -n 10 --wait 2 sh -c 'docker run --rm hello-world'
-
   # older linux releases have separate packages for lxd and lxc (lxd-client)
   if [ "$SPREAD_SYSTEM" = "ubuntu-18.04-64" ] || \
      [ "$SPREAD_SYSTEM" = "ubuntu-20.04-64" ] || \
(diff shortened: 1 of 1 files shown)


lengau (approved):
As you pointed out in Mattermost this isn't great in terms of solving the underlying problem, but it is a decent workaround.

Given that Fedora 36 is EOL as of Tuesday and Ubuntu 18.04 enters ESM this month, maybe it's okay for us to drop these? (I'd want to add at least a replacement for fedora 36 though. ) I'm giving Fedora 37, Fedora CoreOS, Debian (sid & bullseye) and Rocky Linux (8 & 9) a go  and will report back

spread.yaml
```diff
@@ -84,6 +78,11 @@ prepare: |
   lxd waitready --timeout=30
   lxd init --auto
 
+  snap install docker
+
+  # make sure docker is working
```
tigarmo, line 83: possibly add a link to the mattermost thread here? To help us in the future

lengau:
So fun fact: The following other distros fail the tests in the same way if Docker gets installed before LXD:

* Debian (bullseye and sid)
* Fedora 37
* CentOS 7

mr-cal:
> So fun fact: The following other distros fail the tests in the same way if Docker gets installed before LXD:
> 
>     * Debian (bullseye and sid)
> 
>     * Fedora 37
> 
>     * CentOS 7

I bet it is because they do not have `nftables` or `iptables-nft` installed.

mr-cal:
Latest failures: 

https://github.com/canonical/rockcraft/actions/runs/5008322142/jobs/8990735668?pr=262#step:5:126 is a known lxd bug: https://github.com/lxc/lxd/issues/11422 that occurs when the CPU is at 100% utilization

I'm not sure about the other failure (https://github.com/canonical/rockcraft/actions/runs/5008322142/jobs/8990735668?pr=262#step:5:359).

tigarmo:
That second failure is a race condition I think; the output is generated but not before "docker logs" is run.
I'm prototyping a fix here

tigarmo:
actually @mr-cal @sergiusens can either of you bypass-merge this? I'll do the fix in the flaky spread test in a separate PR

PR #269, opened 2023-05-19
docs: create separate chisel documentation

dboddie:

-----

### DIFF
--- a/docs/explanation/chisel.rst
+++ b/docs/explanation/chisel.rst
@@ -1,0 +1,79 @@
+.. _what-is-chisel:
+
+What is Chisel?
+===============
+
+As the name says, Chisel is a tool for carving and cutting. But carving and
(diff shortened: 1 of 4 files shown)


sergiusens (approved):
this doesn't deserve to be a draft anymore :-)

cjdcordeiro (approved):
thanks for this!

I'm approving, although there's a small hiccup:
 - the _static assets are being included via their absolute path in the server, which means they won't render in GitHub. Is there a strong reason not to use relative paths instead?

dboddie:
>     * the _static assets are being included via their absolute path in the server, which means they won't render in GitHub. Is there a strong reason not to use relative paths instead?

I'll change them to be relative.

cjdcordeiro (approved):
awesome thanks

PR #277, opened 2023-05-23
spread: band-aid docker installation on fedora-37

tigarmo:
Latest version of the docker snap fails to install on Fedora with the following message:

```
 error: cannot install "docker": snap "docker" assumes unsupported features:
       snapd2.59.1 (try to refresh snapd)
```

Trying to refresh snapd doesn't actually solve the issue. While we investigage, installing from the older stable channel should keep the spread tests running.

Also do some minor cleanup from the fedora 36 -> 37 update.

-----

### DIFF
--- a/spread.yaml
+++ b/spread.yaml
@@ -56,7 +56,7 @@ prepare: |
   # older linux releases have separate packages for lxd and lxc (lxd-client)
   if [ "$SPREAD_SYSTEM" = "ubuntu-18.04-64" ] || \
      [ "$SPREAD_SYSTEM" = "ubuntu-20.04-64" ] || \
-     [ "$SPREAD_SYSTEM" = "fedora-36-64" ]; then
+     [ "$SPREAD_SYSTEM" = "fedora-37-64" ]; then
     tests.pkgs remove lxd lxd-client
   else
     tests.pkgs remove lxd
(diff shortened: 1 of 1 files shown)


mr-cal:
I think this is the best workaround.  Fedora looks like they are ~3 months behind for snapd releases (https://packages.fedoraproject.org/pkgs/snapd/snapd/).

Can you add a comment to the spread test to mention that we can remove the workaround once snapd is bumped on fedora?

sergiusens:
On Wed, May 24, 2023 at 8:24 AM Tiago Nobrega ***@***.***>
wrote:

> I'm going to add a comment for sure, but I haven't yet because I'm not
> clear on the issue/fix:
>
> a) I tried "hackingly" updating snapd via snap install snapd. No idea if
> that's even possible/advisable, but even after snapd 2.59.2 was installed
> that error persisted;
>
I don't think Fedora has re-exec enabled. We could try that.

> b) The "assumes" keyword there makes me think that this is related to this
> recent Snapcraft bugfix <https://github.com/snapcore/snapcraft/pull/4152>,
> and this docker-snap merge
> <https://github.com/docker-snap/docker-snap/pull/134> is another
> indicator. @sergiusens <https://github.com/sergiusens> do you happen to
> have any idea why our bugfix would lead to this breaking?
>
I haven't check the logs, but if docker assumes a given version of snapd
and if that is not satisfied, then the installation would fail.


> This latest update to the docker snap has a bunch of fairly major changes:
> https://github.com/docker-snap/docker-snap/releases
>

I am most likely thinking is is related to the minimum snapd required

mr-cal (approved):

SamirPS (approved):

zyga:
Hey, just to mention this as I found it referenced from https://github.com/canonical/snapcraft-rocks/blob/core22-7/spread.yaml#L68C7-L68C54

Fedora is now well maintained and should get up-to-date snapd quickly. Re-execution is technically not implemented on Fedora and the remaining work is non-trivial.

PR #278, opened 2023-05-25
 parts: install package repositories in overlays

tigarmo:
This lets 'overlay-packages' include packages from the repositories
listed in 'package-repositories'.


rockcraft/parts.py
```diff
@@ -51,6 +51,8 @@ class PartsLifecycle:
     :raises PartsLifecycleError: On error initializing the parts lifecycle.
     """
 
+    _OVERLAY_CALLBACK_REGISTERED = False
```
sergiusens, line 54: can you add a code comment on what this does? Or docstring?

rockcraft/parts.py
```diff
@@ -67,6 +69,10 @@ def __init__(
 
         emit.progress("Initializing parts lifecycle")
 
+        if not PartsLifecycle._OVERLAY_CALLBACK_REGISTERED:
```
sergiusens, line 72: this seems fragile, do want to have a dedicated class to keep track of registration?

sergiusens (commented):
This makes sense, and looks good. I added some thoughts, not changes or anything like that.

Would approve once we stop using in-devel craft archives.

rockcraft/parts.py
```diff
@@ -67,6 +69,10 @@ def __init__(
...
+            callbacks.register_configure_overlay(_install_overlay_repositories)
```
cmatsuoka, line 73: In the base package exclusion I placed the callback registration in [lifecycle.run](https://github.com/canonical/rockcraft/pull/258/files#diff-2e78c02a1430e75e37cceef3157f4c33ebc785b8693642ebeda5b5d118ec5935R42), before the loop that runs lifecycle more than once. The callback function declaration could stay where it is, only the registration would be done earlier.

requirements-dev.txt
```diff
@@ -16,9 +16,11 @@ codespell==2.2.2
 colorama==0.4.6
 commonmark==0.9.1
 coverage==6.5.0
-craft-archives==1.0.0
+#craft-archives==1.0.0
```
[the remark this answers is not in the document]
lengau: It's probably worth updating `setup.cfg` with the correct minimum version dependency too

tests/spread/large/init/task.yaml
lengau, line 1: Was this deletion intended to be included in this PR?

rockcraft/parts.py
tigarmo: Updated this, please take a look

rockcraft/parts.py
tigarmo: Updated this, please take a look

rockcraft/parts.py
```diff
@@ -202,6 +214,16 @@ def launch_shell(*, cwd: Optional[pathlib.Path] = None) -> None:
         subprocess.run(["bash"], check=False, cwd=cwd)
 
 
+def _install_overlay_repositories(overlay_dir, project_info):
+    if project_info.base != "bare":
```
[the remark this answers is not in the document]
tigarmo: Oh that's a good point. I guess fundamentally the two features (overlays and `bare` base) are incompatible and even without these package repositories I would expect the build to break very soon after this when trying to update the packages in the overlay:

```
Traceback (most recent call last):
  File "/snap/rockcraft/x2/lib/python3.8/site-packages/rockcraft/parts.py", line 145, in run
    with self._lcm.action_executor() as aex:
  File "/snap/rockcraft/x2/lib/python3.8/site-packages/craft_parts/executor/executor.py", line 292, in __enter__
    self._executor.prologue()
  File "/snap/rockcraft/x2/lib/python3.8/site-packages/craft_parts/executor/executor.py", line 101, in prologue
    ctx.refresh_packages_list()
  File "/snap/rockcraft/x2/lib/python3.8/site-packages/craft_parts/overlays/overlay_manager.py", line 230, in refresh_packages_list
    self._overlay_manager.refresh_packages_list()
  File "/snap/rockcraft/x2/lib/python3.8/site-packages/craft_parts/overlays/overlay_manager.py", line 134, in refresh_packages_list
    chroot.chroot(mount_dir, packages.Repository.refresh_packages_list)
  File "/snap/rockcraft/x2/lib/python3.8/site-packages/craft_parts/overlays/chroot.py", line 61, in chroot
    raise errors.OverlayChrootExecutionError(err)
craft_parts.overlays.errors.OverlayChrootExecutionError: Overlay environment execution error: [Errno 2] No such file or directory: 'apt-get'
```

... but this problem is orthogonal to the feature (package repositories) itself, so I propose I remove this `if` here and open a bug to address the feature conflict. What do you think?

tests/spread/large/init/task.yaml
tigarmo: Yes: When I added the `large` suite I put in a "smoke" test because spread fails if there are no tests (like pytest). Now that there is a test in the suite we can remove the smoke test.

lengau: Makes sense, thanks!

tigarmo:
@sergiusens @cmatsuoka @lengau this is ready for re-review thanksss

sergiusens (approved):

lengau (approved):

rockcraft/parts.py
cmatsuoka: Yeah, that works too.

PR #279, opened 2023-05-26
ci: remove test runs on macos-latest

tigarmo:
The macos-latest workers are frequently busy and the majority of the "interesting" tests are skipped on non-linux systems anyway. We can add the OS back in the future if we need to.

-----

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -59,7 +59,7 @@ jobs:
   tests:
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


tigarmo:
I have no idea why the two macos-latest checks are there. I would expect either 0 or 3 (if the workflow update needs to hit `main` before it's actually in effect)

sergiusens (approved):

syu-w (approved):

syu-w (approved):

sergiusens:
@tigarmo fyi, spread-large was triggered :-)

PR #282, opened 2023-05-29
spread: retry grepping docker logs

tigarmo:
One possible source of flakiness in spread tests is the common idiom of starting a container and then immediately grepping the logs. Since our containers' entrypoint is Pebble, it might take a short amount of time to startup, create the services plan, and start whatever it is we want it to start.

Therefore, add a helper command to 'retry' this 'docker logs | grep' pipe.

-----


cjdcordeiro (approved):
hum, this sounds like a workaround for a wider issue: _knowing when pebble services are ready_. This will come in the near future with the concept of health checks, or, for now, we could set pebble checks on all spread tasks and rely on that to know whenever a service is ready.

in either case, this is harmless enough to get approved, and the new command sounds useful anyway

PR #290, opened 2023-06-01
ci: fix check for scheduled event in spread-large

tigarmo:
Turns out it's `event_name` and not `eventname`. Reference: https://www.eliostruyf.com/devhack-github-actions-run-step-schedule/


lengau (approved):

tigarmo:
> if you use vscode you have some good extensions to cover these syntax errors

oh man this is a great idea actually

PR #292, opened 2023-06-01
extensions: add cli commands

### LINKED ISSUE #274: Support Extensions CLI
### What needs to get done

Similar to Snapcraft, we need

- list-extensions
- expand-extensions

### Why it needs to get done

Requires #273

We need to be able to have user consumables for what extensions are available and a way to show users what the extension will do to their rockcraft.yaml when used.

tigarmo:
The commands are 'list-extensions' and 'expand-extensions' and they have the same semantics as the Snapcraft equivalents. Some minor refactoring to re-use testing data accross multiple tests.

Fixes #274

-----


rockcraft/project.py
```diff
@@ -428,6 +429,24 @@ def _validate_package_repositories(
 
         return package_repositories
 
+    def to_yaml(self) -> str:
```
tigarmo, line 432: I changed this from `yaml()` to `to_yaml()` because mypy was complaining about this being an "invalid" override, since `YamlModel.yaml()` takes many parameters and has typing issues of its own (like boolean parameters with a `None` default).

tests/unit/testing/extensions.py
```diff
@@ -0,0 +1,132 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Fake Extensions for use in tests."""
```
tigarmo, line 17: I added this module was a "compromise" between duplicating these classes in multiple tests or having a huge `conftest.py` that every unrelated test run is forced to parse. Not sure if we have a specific convention/nomenclature for this type of code internally

rockcraft/commands/extensions.py
```diff
@@ -0,0 +1,103 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Snapcraft lifecycle commands."""
```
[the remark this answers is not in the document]
sergiusens: yes :-D

rockcraft/commands/extensions.py
```diff
@@ -0,0 +1,103 @@
...
+        List available extensions and their corresponding bases.
+        """
+    )
+
+    @overrides
+    def run(self, parsed_args: argparse.Namespace):
+        """Print the list of available extensions and their bases."""
+        extension_presentation: Dict[str, ExtensionModel] = {}
+
+        # New extensions.
```
sergiusens, line 64: They are all new in Rockcraft
```suggestion
```

sergiusens (approved):

rockcraft/project.py
lengau: (Out of scope here, but a general thought:) Do we want to make this a standard thing for *craft projects? We could pop a `CraftYamlModel` into `craft-application` that does this and inherit from that

lengau (approved):

rockcraft/project.py
sergiusens: this was the original intention. But YamlModel already has a `yaml` method, might be worth looking into API compatibility a bit more

PR #300, opened 2023-06-21
docs: refactor documentation for sphinx-starter-pack

SamirPS:

-----

This commit refactor the documentation in order to be able
to merge easily the sphix-starter-pack after when they add
new functionnallity.

each file have been check with pylint,shellcheck and sphinx.

Signed_Off_By: Samir Akarioh <samir.akarioh@canonical.com>


SamirPS:
Link for the documentation: https://canonical-rockcraft--300.com.readthedocs.build/en/300/

dboddie (approved):

SamirPS:
The tests folder is because of  ``black``, I have refactored the PR to change fewer files.

.github/workflows/automatic-doc-checks.yml
```diff
@@ -0,0 +1,66 @@
+name: Automatic documentation checks
```
cjdcordeiro, line 1: Not that I'm against GH workflows, I actually prefer them, but there is already at least one Sphinx linter in the top-level Makefile, running in a different GH workflow. So I'd rather see everything somewhat unified. Example: if I do `make tests` right now, I won't get spelling nor linkcheck.

Makefile
```diff
@@ -39,17 +36,20 @@ coverage: ## Run pytest with coverage report.
 	coverage report -m
 	coverage html
 
+.PHONY: installdocs
+installdocs: ## Generate documentation.
```
cjdcordeiro, line 40: what's the diff between this and `make docs`? they both say `## Generate documentation.`

docs/.wordlist.txt
```diff
@@ -0,0 +1,76 @@
+Permalink
```
cjdcordeiro, line 1: is there a way to order this alphabetically? As it grows it gets really hard to find a specific word

docs/.wordlist.txt
```diff
@@ -0,0 +1,76 @@
...
+SSL
+Znet
+Snyk's
+Kubernetes
+interoperable
+Dockerfiles
+customized
+specialized
+behavior
+chiseled
```
cjdcordeiro, line 73: should be `chiselled`

docs/_static/favicon.png
cjdcordeiro, line 1: this is the old logo

docs/_static/js/github_issue_links.js
cjdcordeiro, line 1: Maybe I'm missing something in the starter-pack, but they say 

> You don't need to move any files...

but then this file is under `js`, which doesn't exist in the upstream repo.

cjdcordeiro (changes requested):
Looking at the current directory tree, I don't see how sync with upstream without having to check all files manually. Am I wrong? I thought the idea was to "import" the upstream project, literally, as it is, and then do the necessary file manipulation as part of the `make docs` process. This all, all the file manipulation complexity would stay hidden behind an ugle Makefile, but still easier to manage than all of these scattered files.

Makefile
SamirPS: It "s install the Python dependencies

docs/.wordlist.txt
SamirPS: done

docs/_static/favicon.png
SamirPS: Are you sure? Because the logo has been uploaded on the Sphinx pack one month ago

dboddie: It looks like the [icon from the Ubuntu assets site](https://assets.ubuntu.com/v1/49a1a858-favicon-32x32.png) was added to the starter pack: canonical/sphinx-docs-starter-pack#33
Is there another that should be used?

Makefile
```diff
@@ -39,17 +36,28 @@ coverage: ## Run pytest with coverage report.
 	coverage report -m
 	coverage html
 
+.PHONY: preparedocs
+preparedocs: ## move file from the sphinx-starter-pack to docs folder
+	git submodule update --init -- sphinx-starter-pack
+	cp sphinx-starter-pack/.sphinx/_static/* docs/_static
+	mkdir -p docs/_templates
+	cp sphinx-starter-pack/.sphinx/_templates/* docs/_templates
```
cjdcordeiro, line 44: I think you can just copy the entire folder, recursively...so no need to `mkdir`

.github/workflows/tests.yaml
```diff
@@ -55,6 +55,22 @@ jobs:
         run: |
           sudo snap install shellcheck
           make test-shellcheck
+      - name: Install doc framework
```
cjdcordeiro, line 58: You shouldn't need these steps anymore since you now have https://github.com/canonical/rockcraft/pull/300/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R141-R143

docs/explanation/chisel.rst
```diff
@@ -25,7 +25,7 @@ slices.
 The use of package slices provides Rockcraft with the ability to build minimal
 container images from the wider set of Ubuntu packages.
 
-.. figure:: /_static/package-slices.svg
+.. figure:: ../_static/package-slices.svg
```
cjdcordeiro, line 28: why the `../`? The existing form sounds safer cause it's an absolute path. If we move this chisel.rst file tomorrow, this will break

docs/explanation/rocks.rst
```diff
@@ -22,5 +22,5 @@ that container images are built. This enables the use of ROCKs as bases for
 existing build recipes, such as Dockerfiles, for further customisation and
 development.
 
-.. include:: /links.txt
+.. include:: ../links.txt
```
cjdcordeiro, line 25: same as above

cjdcordeiro (changes requested):
looking much more sustainable now! Just a few minor changes and it'll be ok afterwards

cjdcordeiro (approved):
LGTM ty

spread.yaml
```diff
@@ -14,6 +15,8 @@ include:
     - requirements-focal.txt
     - Makefile
     - rockcraft/
+    - sphinx-starter-pack/
+    - .git/
```
tigarmo, line 19: include `.git`? how come?

.gitignore
```diff
@@ -142,3 +142,11 @@ rockcraft_*.snap
...
 # rock
 *.rock
+
+# sphinx
+
+docs/warnings.txt
+docs/.sphinx/.wordlist.dic
+.DS_Store
+__pycache__
+.idea/
```
sergiusens, line 152:
```suggestion
.idea/

```

.gitmodules
```diff
@@ -2,3 +2,6 @@
 	path = tools/external
 	url = https://github.com/snapcore/snapd-testing-tools.git
 	branch = main
+[submodule "sphinx-starter-pack"]
+	path = sphinx-starter-pack
```
sergiusens, line 6: can' this be docs/sphinx-starter-pack ?

docs/.wordlist.txt
```diff
@@ -0,0 +1,76 @@
+artifacts
```
tigarmo, line 1: `artifacts` and `behavior` are interesting; shouldn't we use `artefacts` and `behaviour` instead?

docs/conf.py
```diff
@@ -44,6 +44,26 @@
...
+html_favicon = "_static/favicon.png"
+
+html_context = {
+    # Change to the discourse instance you want to be able to link to
+    # (use an empty value if you don't want to link)
+    "discourse": "https://discourse.ubuntu.com/c/rocks/117",
+    # Change to the GitHub info for your project
+    "github_url": "https://github.com/canonical/rockcraft",
+    # Change to the branch for this version of the documentation
+    "github_version": "main",
```
sergiusens, line 57: doesn' the readthedocs infra handle this for us?

docs/conf.py
```diff
@@ -44,6 +44,26 @@
...
+    # Change to the folder that contains the documentation (usually "/" or "/docs/")
+    "github_folder": "/docs/",
+    # Change to an empty value if your GitHub repo doesn't have issues enabled
+    "github_issues": "enabled",
+}
+
+
+if "discourse" in html_context:
```
sergiusens, line 65: what does this do in practice?

spread.yaml
```diff
@@ -5,6 +5,7 @@ environment:
   PROJECT_PATH: /rockcraft
   SNAPD_TESTING_TOOLS: $PROJECT_PATH/tools/external/tools
   PATH: /snap/bin:$PATH:$SNAPD_TESTING_TOOLS:$PROJECT_PATH/tools/spread
+  SPHINX_STARTER_PACK: $PROJECT_PATH/sphinx-starter-pack
```
sergiusens, line 8: this propagates to all tests, do we really need it here?

docs/Makefile
```diff
@@ -7,14 +7,43 @@ SPHINXOPTS    ?= -q
...
+PORT = 8080
 
 # Put it first so that "make" without argument is like "make help".
 help:
 	@$(SPHINXBUILD) -M help "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
 
+install:
+	@echo "... setting up virtualenv"
+	python3 -m venv env
+	. $(VENV); pip install --upgrade -r ../requirements-doc.txt -r ../requirements-focal.txt
```
tigarmo, line 20: what about the requirements in `.sphinx/pinned-requirements.txt`? It's curious that the docs can build without them

tigarmo (commented):
thanks for doing all this work @SamirPS !

docs/conf.py
SamirPS: This is needed for having the link for discourse and GitHub on the RTD

```
* ``discourse_prefix`` - the URL to the Discourse instance your project uses (needed to add links to posts using the ``:discourse:`` metadata at the top of a file)
* ``github_url`` - the link to your GitHub repository (needed to create the Edit link in the footer and the feedback button)
* ``github_version`` - the branch that contains this version of the documentation
* ``github_folder`` - the folder that contains the documentation files
```

docs/conf.py
SamirPS: Create the link at the end of each page 

```
[Ask a question on Discourse](https://discourse.ubuntu.com/c/rocks/117)
```

spread.yaml
SamirPS: no we only need it for one test

docs/Makefile
SamirPS: It's also because i didn't add all the extensions, like sphinx_tabs or didn't upgrade the version of the furo theme, for example, but I can add the extensions if you want in the requirements-doc.txt file and conf.py.

docs/.wordlist.txt
tigarmo: right, that's what I mean: if we're using en_GB (which we should), shouldn't we update the uses of `artifacts` to `artefacts`, instead of allowing the `artifacts` spelling?

docs/Makefile
tigarmo: can we use only the `.sphinx/pinned-requirements.txt` file?

spread.yaml
tigarmo: which spread test calls `make prepare-docs`?

SamirPS: the how-to/code/build-docs/task.yaml with ``make installdocs``

docs/Makefile
SamirPS: no because :

- Need to solve some version issues because some need sphinx>6 and others <4
- We will need to add our dependencies

For theses reasons, I think it's better to use requirements-doc

spread.yaml
SamirPS: Do you have an idea about it without importing the .git folder ?

docs/.wordlist.txt
SamirPS: But looking at other words, I suspect we use US English ( initializing, customized,...)

cjdcordeiro: we need to use `en_GB` so if the spelling check is not flagging those words, something's wrong

SamirPS: If I remove it from the wordlist, the spelling check flags the word.

SamirPS: i modify some RST files to use ``en_GB``

docs/Makefile
tigarmo: I see. We should probably take some time to clean-up those dependencies (iirc rockcraft doesn't need autodocs because we create docs from the code) but we can do that in the future, can you please create an issue for it?

tigarmo (approved):
nice work!

docs/Makefile
SamirPS: Here the issue: https://github.com/canonical/rockcraft/issues/306

dboddie (approved):

dboddie (approved):

Issue #284, opened 2023-05-31
Block overlay usage with `bare` bases


tigarmo:
### What needs to get done

If a project has `bare` base and either `overlay-script` or `overlay-packages`, we should fail as early as possible.

------

Ideally we should fail during project validation, but we would need a craft-parts-provided way of determining whether a project is using overlays (so that we don't keep all that logic in Rockcraft itself).

### Why it needs to get done

The two features are essentially mutually exclusive: the overlays are a way to configure the base system, and the `bare` base is about not having a base system at all.

SamirPS:
can this issue been closed?

mr-cal:
Completed in #301.

PR #301, opened 2023-06-21
project: block overlay usage with "bare" bases

tigarmo:

-----

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -18,7 +18,7 @@ commonmark==0.9.1
 coverage==6.5.0
 craft-archives==1.1.0
 craft-cli==2.0.0
-craft-parts==1.21.0
+craft-parts==1.22.0
 craft-providers==1.8.0
 cryptography==38.0.4
 Deprecated==1.2.13
--- a/requirements-doc.txt
+++ b/requirements-doc.txt
@@ -6,7 +6,7 @@ charset-normalizer==2.1.1
 colorama==0.4.6
 craft-archives==1.1.0
 craft-cli==2.0.0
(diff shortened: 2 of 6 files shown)


requirements-dev.txt
```diff
@@ -18,7 +18,8 @@ commonmark==0.9.1
 coverage==6.5.0
 craft-archives==1.1.0
 craft-cli==2.0.0
-craft-parts==1.21.0
+#craft-parts==1.21.0
+git+https://github.com/canonical/craft-parts.git@CRAFT-1770-has-overlay#egg=craft_parts
```
tigarmo, line 22: noting this to myself: update before merging

cjdcordeiro (approved):
Thanks. This LGTM, although it feels like it's patching a bigger and unrelated issue -> `overlay` has, IMO, the connotation that something is running on top of **what you have primed so far**. I.e., if we eventually introduce the concept of ordered parts and OCI layering based on parts, I'd expect the overlay script of layer 5 to run on whatever is already present in layer 4 and below. This is not possible today though, so the only reason I'm raising this point is because I'm concerned that we might be giving the wrong meaning to `overlay`

sergiusens:
Closes #284

requirements-dev.txt
cmatsuoka: Craft-parts 1.22.0 released.

cmatsuoka:
> I'd expect the overlay script of layer 5 to run on whatever is already present in layer 4 and below.

The layer stack should be assembled up to the layer you're running the overlay script on -- is something different happening?

PR #302, opened 2023-06-22
parts: add project-specific global variables

### LINKED ISSUE #299: Add project env vars like CRAFT_PROJECT_VERSION
### What needs to get done

Much like Snapcraft exposes `SNAPCRAFT_PROJECT_VERSION`, Rockcraft should add project-specific environment variables.

### Why it needs to get done

See https://chat.charmhub.io/charmhub/pl/tbap1wiiw7nebkekhjix4asu1r

tigarmo:
The implementation follows Snapcraft's: a prologue callback is used to fill the project info with the project's name and version (the other 'useful' variables are already provided by craft_parts). This lets parts use those variables (for example during their builds).

A lot of the unit test changes come from a refactoring to make adding new parameters to parts.PartsLifecycle less painful in the future.

Closes #299

-----

### DIFF
--- a/rockcraft/parts.py
+++ b/rockcraft/parts.py
@@ -58,12 +58,14 @@ class PartsLifecycle:
         self,
         all_parts: Dict[str, Any],
         *,
+        project_name: str,
         work_dir: pathlib.Path,
         part_names: Optional[List[str]],
         base_layer_dir: pathlib.Path,
         base_layer_hash: bytes,
         base: str,
         package_repositories: Optional[List[Dict[str, Any]]] = None,
+        project_vars: Optional[Dict[str, str]] = None,
     ):
         self._part_names = part_names
         self._package_repositories = package_repositories or []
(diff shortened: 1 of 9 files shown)


rockcraft/parts.py
```diff
@@ -58,12 +58,14 @@ def __init__(
         self,
         all_parts: Dict[str, Any],
         *,
+        project_name: str,
```
sergiusens, line 61: was this required to populate CRAFT_PROJECT_NAME?

sergiusens (approved):
neat work!

rockcraft/parts.py
tigarmo: yes. I could have used the `project_vars` dict for this, but `craft_parts.LifecycleManager` has a `project_name` constructor parameter which seemed like the more "correct" way to do this

cmatsuoka (approved):
Looks good!

PR #311, opened 2023-06-29
plugins: fix python shebang

tigarmo:
Use ${PARTS_PYTHON_INTERPRETER} to play nice with the upstream plugin;
If the user overrides this, there's probably a good reason (like using
Python 3.6).

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -18,7 +18,8 @@ commonmark==0.9.1
 coverage==6.5.0
 craft-archives==1.1.0
 craft-cli==2.0.0
-craft-parts==1.22.0
+#craft-parts==1.22.0
+git+https://github.com/tigarmo/craft-parts.git@fix-python-3.6#egg=craft_parts
 craft-providers==1.8.0
 cryptography==38.0.4
 Deprecated==1.2.13
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -18,8 +18,7 @@ commonmark==0.9.1
 coverage==6.5.0
 craft-archives==1.1.0
 craft-cli==2.0.0
-#craft-parts==1.22.0
-git+https://github.com/tigarmo/craft-parts.git@fix-python-3.6#egg=craft_parts
+craft-parts==1.23.0
 craft-providers==1.8.0
 cryptography==38.0.4
 Deprecated==1.2.13
--- a/rockcraft/plugins/python_plugin.py
+++ b/rockcraft/plugins/python_plugin.py
@@ -86,7 +86,7 @@ class PythonPlugin(python_plugin.PythonPlugin):
     @override
     def _get_script_interpreter(self) -> str:
         """Overridden because Python is always available in /bin/python3."""
-        return "#!/bin/python3"
+        return "#!/bin/${PARTS_PYTHON_INTERPRETER}"
 
     @override
     def get_build_commands(self) -> List[str]:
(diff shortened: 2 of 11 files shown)


requirements-dev.txt
```diff
@@ -18,7 +18,8 @@ commonmark==0.9.1
 coverage==6.5.0
 craft-archives==1.1.0
 craft-cli==2.0.0
-craft-parts==1.22.0
+#craft-parts==1.22.0
+git+https://github.com/tigarmo/craft-parts.git@fix-python-3.6#egg=craft_parts
```
tigarmo, line 22: self-reminder to update this

lengau (approved):
Looks good! I gather this is going to need a craft-parts 1.23 update?

tests/spread/general/plugin-python-3.6/rockcraft.orig.yaml
```diff
@@ -0,0 +1,23 @@
...
+    priority: always
+
+parts:
+    my-part:
+        plugin: python
+        source: .
+        stage-packages: [python3.6-venv]
+        python-packages: [black]
+        build-environment:
+            - PARTS_PYTHON_INTERPRETER: python3.6
```
cmatsuoka, line 23: What happens if the user tries to specify the full path for the interpreter? If that's not allowed, maybe we should detect and warn? Or, if that's valid, the shebang line shouldn't hardcode `/bin/`.

tigarmo: Hmm good question. This `PARTS_PYTHON_INTERPRETER` variable is used both during build and during runtime (in the generated containers) and I'm having a hard time thinking of a full path value that would work in both "stages", so I think we should block it. I'll add a check and a test

rockcraft/plugins/python_plugin.py
```diff
@@ -86,7 +86,7 @@ def _get_system_python_interpreter(self) -> Optional[str]:
     @override
     def _get_script_interpreter(self) -> str:
         """Overridden because Python is always available in /bin/python3."""
```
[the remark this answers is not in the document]
tigarmo: you're right! updated

tests/spread/general/plugin-python-3.6/rockcraft.orig.yaml
tigarmo: Added check & test

rockcraft/plugins/python_plugin.py
```diff
@@ -85,20 +85,35 @@ def _get_system_python_interpreter(self) -> Optional[str]:
...
-        original = super().get_build_commands()
+        commands = []
+
+        # Detect whether PARTS_PYTHON_INTERPRETER is a full path (not supported)
+        commands.append(
+            dedent(
+                """
+                # Detect whether PARTS_PYTHON_INTERPRETER is a full path
+                if [[ "${PARTS_PYTHON_INTERPRETER}" = /* ]]; then
+                    echo "Full paths in \"PARTS_PYTHON_INTERPRETER\" are not allowed: ${PARTS_PYTHON_INTERPRETER}"
```
cmatsuoka, line 102: Maybe absolute paths instead of full paths?

cmatsuoka (approved):
Thanks! Added a note about path naming consistency, I think it's referred to as absolute path elsewhere.

rockcraft/plugins/python_plugin.py
tigarmo: argh you are so right. I'll fix

PR #313, opened 2023-07-03
feat: re-introduce the OCI environment

cjdcordeiro:

-----

This PR brings back the `env` field, now renamed to `environment` and converted into a Dict instead of a List (to be consistent with the `services` syntax.


docs/reference/rockcraft.yaml.rst
```diff
@@ -111,6 +111,16 @@ The default OCI user. It must be a supported shared user. Currently, the only
 supported shared user is "_daemon_" (with UID/GID 584792). It defaults to
 "root" (with UID 0).
 
+``environment``
+------------
```
SamirPS, line 115: For the error on the docs workflow just add more ``-`` here

jnsgruk (commented):
Nice, thanks @cjdcordeiro :)

jnsgruk (approved):
Nice, thanks @cjdcordeiro :)

SamirPS (approved):

rockcraft/project.py
```diff
@@ -228,6 +228,7 @@ class Project(YamlModel):
     platforms: Dict[str, Any]
     base: Literal["bare", "ubuntu:18.04", "ubuntu:20.04", "ubuntu:22.04"]
     build_base: Optional[Literal["ubuntu:18.04", "ubuntu:20.04", "ubuntu:22.04"]]
+    environment: Optional[Dict[str, Optional[str]]]
```
sergiusens, line 231: why is the value `Optional[str]` and not `str`? I don't see any examples of that either

sergiusens (changes requested):
Looks good, I just have one nit that could be solved with an answer and/or documentation examples or by making the accepted values more strict

rockcraft/project.py
cjdcordeiro: basically copied from https://github.com/snapcore/snapcraft/blob/main/snapcraft/projects.py#L447. Might be redundant though.

cjdcordeiro: @sergiusens given the snapcraft reference, which solution do you prefer? (I'm ok with removing the `Optional`...in fact, that's what I had initially in https://github.com/canonical/rockcraft/commit/c45a50e58c4546e0bcdadb32b58d3d1fe1eae791#diff-0ea7d1a44d76c7792a852505490ca9e3fb738ec2a7814b051ebc828efb8ceceeR229)

tigarmo: What's the meaning of the missing string? E.g. what does

```
environment:
  X:
```

mean for the value of `X` in the final container?

tigarmo: I tried it out here and apparently the variable get the literal string `"None"` as value

sergiusens:
I believe this is optional on Snapcraft by mistake.Message ID:
***@***.***>

>

rockcraft/project.py
cjdcordeiro: alright, I've removed the `Optional`

tigarmo (approved):

sergiusens (approved):

sergiusens:
Nice quality work!

PR #317, opened 2023-07-05
extensions: new flask extension

weiiwang01:
This pull request introduces an experimental Flask extension for the Juju Platform as a Service ecosystem in rockcraft. 
The extension aims to facilitate the deployment and management of Flask-based web applications on Juju PaaS.

Here's an example project that uses the `flask-extension` https://github.com/weiiwang01/dqlite.io/blob/flask-extension/rockcraft.yaml.

- [] Have you signed the [CLA](http://www.ubuntu.com/legal/contributors/)?


rockcraft/extensions/paas_flask.py
```diff
@@ -0,0 +1,105 @@
...
+        """Return the part snippet to apply to existing parts."""
+        return {}
+
+    @override
+    def get_parts_snippet(self) -> Dict[str, Any]:
+        """Create necessary parts to facilitate the flask application.
+
+        Parts added:
+            - flask-extension/dependencies: install Python dependencies
+            - flask-extension/install-app: copy the flask project into the OCI image
```
sergiusens, line 78: these names are outdated

rockcraft/extensions/paas_flask.py
```diff
@@ -0,0 +1,105 @@
...
+from overrides import override
+
+from .extension import Extension
+
+
+class PaasFlask(Extension):
+    """An extension to enable Flask support within the Juju PaaS ecosystem.
+
+    This extension is a part of the Juju PaaS ecosystem and facilitates the deployment of
+    Flask-based applications.
```
sergiusens, line 31: This should be generic and not Juju specific

rockcraft/extensions/__init__.py
```diff
@@ -26,3 +26,7 @@
     "register",
     "unregister",
 ]
+
+from .paas_flask import PaasFlask as _PaasFlask
+
+register("paas-flask", _PaasFlask)
```
sergiusens, line 32: this should be called flask, which is what we agreed to (no paas in the name)

rockcraft/extensions/paas_flask.py
```diff
@@ -0,0 +1,105 @@
...
+                "stage-packages": ["python3-venv"],
+                "source": ".",
+                "python-packages": ["gunicorn"],
+                "python-requirements": python_requirements,
+            },
+            "paas-flask/install-app": {
+                "plugin": "dump",
+                "source": ".",
+                "organize": renaming_map,
+                "stage": list(renaming_map.values()),
```
sergiusens, line 102: how is this supposed to work? Were we not supposed to use the npm target for this?

We should add some minimal documentation on how to work with this

weiiwang01: This approach aligns with the web team's requirements, as they prefer to construct the static assets themselves before executing the `rockcraft pack` command. And let rockcraft should incorporate the built artifact into the OCI image without making any modifications. I will add some comments about this.

rockcraft/extensions/__init__.py
weiiwang01: Updated, thanks!

rockcraft/extensions/paas_flask.py
weiiwang01: Removed all reference to juju

rockcraft/extensions/paas_flask.py
jdkandersson: For flask, developers don't use npm for static assets since flask is a server side rendered framework primarily. Hence we decided not to add support for `npm` based static assets in the flask extension. For others it could make sense or those assets should be served, e.g., using a content cache instead or dedicated 12-factor extension

sergiusens:
That'a perfecly reasonable explanation. I would still like to see a spread
test for this and perhaps some documentation. Documentation can be a
separate PR, but a spread test show casing this seems like a must.

weiiwang01:
> That'a perfecly reasonable explanation. I would still like to see a spread test for this and perhaps some documentation. Documentation can be a separate PR, but a spread test show casing this seems like a must.

Spread test added, thanks!

tests/spread/general/extension-flask/rockcraft.yaml
```diff
@@ -0,0 +1,9 @@
+name: flask-extension
+summary: OCI image for a flask project.
+description: OCI image for a flask project.
+version: "0.1"
+base: bare
+license: Apache-2.0
+
+extensions:
```
sergiusens, line 8: we need this entry at least referenced now that we are referring to one

https://github.com/canonical/rockcraft/blob/main/docs/reference/rockcraft.yaml.rst

rockcraft/extensions/flask.py
```diff
@@ -0,0 +1,105 @@
...
+                "missing requirements.txt file, "
+                "flask extension requires this file with flask specified as a dependency"
+            )
+        ignores = [".git", "node_modules"]
+        source_files = [
+            f
+            for f in os.listdir(self.project_root)
+            if f not in ignores and not f.endswith(".rock")
+        ]
+        renaming_map = {f: os.path.join("srv/flask/app", f) for f in source_files}
```
sergiusens, line 87: how would one know this is the place where things need to be put?

tests/spread/general/extension-flask/rockcraft.yaml
weiiwang01: Added, thanks!

rockcraft/extensions/flask.py
weiiwang01: We need to choose a location that is agreed upon by both the charm and the rock for it to function properly. The path `/srv/flask/app` is the outcome of this agreement.

docs/how-to/use-flask-extension.rst
```diff
@@ -0,0 +1,58 @@
+The Flask Extension
```
sergiusens, line 1: this looks like a good start, it does however mixup reference, explanation, tutorial and how-to 

ref: https://diataxis.fr/

weiiwang01: Gotcha, I have splited the document into a "how-to" and a "reference".
I also added the `services` field to the extension as well following last week's discussion.

weiiwang01:
@tigarmo I have updated the pull request with some unit tests for the flask extension.

docs/how-to/use-flask-extension.rst
```diff
@@ -0,0 +1,49 @@
...
+    extensions:
+      - flask
+
+Managing project files with the flask extension
+-----------------------------------------------
+
+By default, all files within the Flask project directory are copied, excluding
+certain common files and directories, such as ``node_modules``. However,
+this behaviour can be tailored to either specifically include or exclude files
+from the Flask project directory in the rock image.
```
tigarmo, line 28:
```suggestion
from the Flask project directory in the ROCK image.
```

docs/how-to/use-flask-extension.rst
```diff
@@ -0,0 +1,49 @@
...
+
+To include only select files from the project directory in the rock image,
```
tigarmo, line 30:
```suggestion
To include only select files from the project directory in the ROCK image,
```

(and everywhere else too please)

docs/how-to/use-flask-extension.rst
```diff
@@ -0,0 +1,49 @@
...
+append the following part to ``rockcraft.yaml``:
```
tigarmo, line 31: I think the example is good but we should explain the `prime` bit just _a bit_ to give some necessary context. Something like:

> You can include and exclude files from the project directory in the ROCK image by using the standard `prime` declaration on the specially-named `flask/install-app` part. For example, to include only select files:

(and then)

> or to exclude specific files:

docs/reference/extensions.rst
```diff
@@ -0,0 +1,16 @@
+Extensions
+**********
+
+Just as the snapcraft extensions are designed to simplify snap creation,
```
tigarmo, line 4:
```suggestion
Just as the Snapcraft extensions are designed to simplify Snap creation,
```

docs/reference/extensions.rst
```diff
@@ -0,0 +1,16 @@
...
+Rockcraft extensions are crafted to expand and modify the user-provided
+rockcraft manifest file, aiming to minimise the boilerplate code when
```
tigarmo, line 6: hmm not sure we should use the word "manifest" here, I don't think we use it anywhere else in the docs. "project" is probably more appropriate.

docs/reference/extensions.rst
```diff
@@ -0,0 +1,16 @@
...
+initiating a new rock.
+
+The flask extension
```
tigarmo, line 9:
```suggestion
The ``flask`` extension
```

docs/reference/extensions.rst
```diff
@@ -0,0 +1,16 @@
+Extensions
```
tigarmo, line 1: :+1:  ty for starting this page

docs/reference/rockcraft.yaml.rst
```diff
@@ -211,6 +211,18 @@ The set of parts that compose the ROCK's contents
...
    ``services`` to define your container application.
 
+``extensions``
+--------------
+
+**Type**: list[string]
+
+**Required**: No
+
+Extensions to apply to this rock image.
```
tigarmo, line 221:
```suggestion
Extensions to enable when building the ROCK.
```

rockcraft/extensions/__init__.py
```diff
@@ -26,3 +26,7 @@
     "register",
     "unregister",
 ]
+
+from .flask import Flask as _Flask
```
tigarmo, line 30: why the rename?

rockcraft/extensions/flask.py
```diff
@@ -0,0 +1,169 @@
...
+            ),
+            install_app_part_name: self._merge_existing_part(
+                install_app_part_name, install_app_part
+            ),
+        }
+        if self.yaml_data["base"] == "bare":
+            snippet["flask/container-processing"] = {
+                "plugin": "nil",
+                "source": ".",
+                "override-prime": "craftctl default\nmkdir -p tmp\nchmod 777 tmp",
```
tigarmo, line 162: a bit surprised that this works, I guess `craftctl default` is changing the current dir to `CRAFT_PRIME`?
regardless, I think it's cleaner/more idiomatic to do something like `"override-build": "mkdir -m 777 ${CRAFT_PART_INSTALL}/tmp"`

rockcraft/extensions/flask.py
```diff
@@ -0,0 +1,169 @@
...
+            }
+        return snippet
+
+    @override
+    def get_parts_snippet(self) -> Dict[str, Any]:
+        """Create necessary parts to facilitate the flask application."""
```
tigarmo, line 168: I guess this docstring is redundant/misleading

tigarmo (changes requested):
Thanks! I haven't finished reviewing yet, will pick this up again in a little bit

docs/how-to/use-flask-extension.rst
```diff
@@ -0,0 +1,49 @@
...
+    base: bare
+    license: Apache-2.0
+
+    extensions:
+      - flask
+
+Managing project files with the flask extension
+-----------------------------------------------
+
+By default, all files within the Flask project directory are copied, excluding
```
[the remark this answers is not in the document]
weiiwang01: Yes, maybe we can confirm that in today's meeting.

docs/how-to/use-flask-extension.rst
```diff
@@ -0,0 +1,49 @@
...
+
+The Flask extension is compatible with ``bare``, ``ubuntu:20.04``, and
+``ubuntu:22.04``. To employ it, include ``extensions: [flask]`` in your
+``rockcraft.yaml`` file.
+
+Example:
+
+.. code-block:: yaml
+
+    name: example
```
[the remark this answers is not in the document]
weiiwang01: Yes, indeed, updated, thanks!

rockcraft/extensions/flask.py
```diff
@@ -0,0 +1,169 @@
...
+        services = {
+            "flask": {
+                "override": "replace",
+                "startup": "enabled",
+                "command": "/bin/python3 -m gunicorn --bind 0.0.0.0:8000 app:app",
+                "user": "_daemon_",
+                "working-dir": "/srv/flask/app",
+            }
+        }
+        existing_services = self.yaml_data.get("services", {})
```
tigarmo, line 81: I think for consistency with the part handling this should also be a `deepcopy` of the existing data

tests/spread/general/extension-flask/task.yaml
```diff
@@ -0,0 +1,31 @@
...
+  docker rmi --force flask-extension
+  # Install container
+  sudo /snap/rockcraft/current/bin/skopeo --insecure-policy copy oci-archive:flask-extension_0.1_amd64.rock docker-daemon:flask-extension:latest
+  # Ensure container exists
+  docker images flask-extension | MATCH "flask-extension"
+  
+  # ensure container doesn't exist
+  docker rm -f flask-extension-container
+  
+  # test the flask project is ready to run inside the container
```
tigarmo, line 21: Thanks for this test! I think we need two more checks here, to ensure that everything is hooked up ok:
1) Run the container without overriding the entrypoint, to make sure the pebble service works
2) Run the container with a port bind and then check that the webapp is up (for example, curl the address maybe)

tests/unit/extensions/test_extensions.py
```diff
@@ -134,3 +134,102 @@ def test_project_load_extensions(fake_extensions, tmp_path):
...
+        "working-dir": "/srv/flask/app",
+    }
+
+    parts = applied["parts"]
+    assert sorted(parts["flask/install-app"]["stage"]) == [
+        "srv/flask/app/app.py",
+        "srv/flask/app/requirements.txt",
+        "srv/flask/app/static",
+    ]
+    del parts["flask/install-app"]["stage"]
```
tigarmo, line 171: an alternative to sorting + deleting it here, which I assume you did because of the unpredictable ordering, is to have the extension always sort the list of stage files. Then it's more deterministic which is a plus

tests/unit/extensions/test_extensions.py
```diff
@@ -134,3 +134,102 @@ def test_project_load_extensions(fake_extensions, tmp_path):
...
+    (tmp_path / "requirements.txt").write_text("flask")
+    (tmp_path / "foobar").touch()
+    (tmp_path / "webapp").mkdir()
+    (tmp_path / "webapp/app.py").touch()
+    (tmp_path / "static").mkdir()
+    (tmp_path / "node_modules").mkdir()
+
+    input_yaml["extensions"] = ["flask"]
+    input_yaml["parts"] = {"flask/install-app": {"prime": ["-srv/flask/app/foobar"]}}
+    input_yaml["services"] = {
```
tigarmo, line 205: possibly add another service here, to test that this new service is preserved

rockcraft/extensions/flask.py
```diff
@@ -0,0 +1,169 @@
...
+
+    def _gen_new_parts(self) -> Dict[str, Any]:
+        """Generate new parts for the flask extension.
+
+        Parts added:
+            - flask/dependencies: install Python dependencies
+            - flask/install-app: copy the flask project into the OCI image
+        """
+        if not (self.project_root / "requirements.txt").exists():
+            raise ValueError(
```
tigarmo, line 122: I think `errors.ExtensionError` is more specific here

tests/unit/extensions/test_extensions.py
```diff
@@ -134,3 +134,102 @@ def test_project_load_extensions(fake_extensions, tmp_path):
 
     # New part
     assert parts[f"{FullExtension.NAME}/new-part"] == {"plugin": "nil", "source": None}
+
+
+def test_flask_extensions(fake_extensions, tmp_path, input_yaml, monkeypatch):
```
tigarmo, line 139: Thanks for the tests! I think a couple are missing:
- A test that checks that `apply_extensions()` fails if `requirements.txt` is missing
- A test that checks that `build_base: ubuntu:22.04` gets added if the key is missing and `base` is bare
- Since the extension has a custom merge for the parts, a test that merges existing keys. For example: a yaml that has:
```
parts:
    flask/dependencies:
        python-requirements: ["requirements-jammy.txt"]

Comment (tigarmo) [reply tests/unit/extensions/test_extensions.py:139]: In light of this, I think all of these tests (and these new ones) should be moved to a `test_flask` module in `unit/extensions`

**Review by tigarmo — CHANGES_REQUESTED** [review]
Sorry for the delay, now I'm done!

File: tests/spread/general/extension-flask/task.yaml (hunk 22 above)
Comment (tigarmo) [reply tests/spread/general/extension-flask/task.yaml:21]: example: https://github.com/canonical/rockcraft/blob/main/tests/spread/large/repo-overlay/task.yaml#L16

File: rockcraft/extensions/flask.py [hunk 27]
```diff
@@ -0,0 +1,169 @@
...
+                "missing requirements.txt file, "
+                "flask extension requires this file with flask specified as a dependency"
+            )
+        ignores = [".git", "node_modules", ".yarn"]
+        source_files = [
+            f
+            for f in os.listdir(self.project_root)
```
jdkandersson, line 129: what about using pathlib?

rockcraft/extensions/flask.py
```diff
@@ -0,0 +1,169 @@
...
+            if f not in ignores and not f.endswith(".rock")
+        ]
+        renaming_map = {f: posixpath.join("srv/flask/app", f) for f in source_files}
```
jdkandersson, line 132: I was trying to find the handling of requiring prime and checking that everything in there has the `srv/flask/app` prefix?

rockcraft/extensions/flask.py
weiiwang01: Indeed, updated, thanks!

tests/spread/general/extension-flask/task.yaml
weiiwang01: Gotcha, updated, thanks!

tests/unit/extensions/test_extensions.py
weiiwang01: Good idea, updated, thanks!

rockcraft/extensions/flask.py
weiiwang01: That has been added in the latest commit, thanks!

rockcraft/extensions/flask.py
```diff
@@ -0,0 +1,169 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""An experimental extension for the Flask framework."""
+
+import copy
+import os
+import posixpath
+from typing import Any, Dict, Optional, Tuple
```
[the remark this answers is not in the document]
weiiwang01: rockcraft is targeting `python>=3.8`

docs/how-to/use-flask-extension.rst
```diff
@@ -0,0 +1,50 @@
+Using the flask extension
+-------------------------
+
+The Flask extension is compatible with the ``bare``, ``ubuntu:20.04``, and
+``ubuntu:22.04`` bases. To employ it, include ``extensions: [flask]`` in your
+``rockcraft.yaml`` file.
+
+Example:
+
+.. code-block:: yaml
```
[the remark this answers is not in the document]
weiiwang01: Documents has been updated, thanks!

tests/unit/extensions/test_extensions.py
weiiwang01: Added, thanks!

tests/unit/extensions/test_extensions.py
weiiwang01: Gotcha, I have included more tests and move they to the `test_flask.py`, thanks!

rockcraft/extensions/flask.py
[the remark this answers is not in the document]
tigarmo: @jdkandersson is right with regards to the current best practices but the rest of rockcraft's codebase is still using the types from `typing`,  so I suggest we keep this for consistency and then in a separate issue we can convert all code

tests/spread/general/extension-flask/task.yaml
```diff
@@ -25,6 +25,11 @@ execute: |
   
   # test the part merging
   docker run --rm --entrypoint /bin/python3 flask-extension -c "import pathlib;assert not pathlib.Path('/srv/flask/app/README').exists()"
+  
+  # test the default flask service
+  docker run --rm --name flask-extension-container -d -p 8137:8000 flask-extension
```
tigarmo, line 30: we should remove this container in the `restore` step
`docker rm -f flask-extension-container`

tests/unit/extensions/test_flask.py
```diff
@@ -0,0 +1,192 @@
...
+    assert applied["parts"]["flask/container-processing"] == {
+        "plugin": "nil",
+        "source": ".",
+        "override-build": "mkdir -m 777 ${CRAFT_PART_INSTALL}/tmp",
+    }
+    assert applied["build-base"] == "ubuntu:22.04"
+
+
+@pytest.mark.usefixtures("flask_extension")
+def test_flask_extension_error(tmp_path):
```
tigarmo, line 164: thanks for testing the error cases! I think we should split this into separate tests. It's a bit of work but it makes each individual test easier to reason about, and lowers the chance that they might interfere with each other in unexpected ways

docs/how-to/use-flask-extension.rst
```diff
@@ -19,17 +19,21 @@ Example:
...
 Managing project files with the flask extension
 -----------------------------------------------
 
-By default, all files within the Flask project directory are copied, excluding
-certain common files and directories, such as ``node_modules``. However,
-this behaviour can be tailored to either specifically include or exclude files
-from the Flask project directory in the ROCK image.
+The prime declaration must be included in the specially-named
+``flask/install-app`` section to instruct the flask extension on which files
+to include or exclude from the project directory in the ROCK image.
```
tigarmo, line 34: I think we should now document the fact that every item in the `prime` list must be prefixed with `srv/flask/app`. This also gives us a chance to talk about this prefix, heh.

Here's a rough first attempt:

```
The extension places the files from the project folder in the `/srv/flask/app` directory in the final image.
Therefore, all inclusions and exclusions must be prefixed with `src/flask/app`.

**Review by tigarmo — CHANGES_REQUESTED** [review]
thanks for all this work @weiiwang01

**Comment by sergiusens** [comment]
when this is merged, a snap of rockcraft with this implementation will show up as edge/12f (the feature branch leaf name)

File: docs/how-to/use-flask-extension.rst (hunk 33 above)
Comment (weiiwang01) [reply docs/how-to/use-flask-extension.rst:34]: Yes indeed, added, thanks!

File: tests/unit/extensions/test_flask.py (hunk 32 above)
Comment (weiiwang01) [reply tests/unit/extensions/test_flask.py]: Gotcha, separate them into different test cases with appropriate names, thanks!

File: tests/spread/general/extension-flask/task.yaml (hunk 31 above)
Comment (weiiwang01) [reply tests/spread/general/extension-flask/task.yaml]: Added, thanks!

File: docs/how-to/use-flask-extension.rst [hunk 34]
```diff
@@ -32,6 +32,9 @@ Managing project files with the flask extension
 The prime declaration must be included in the specially-named
 ``flask/install-app`` section to instruct the flask extension on which files
 to include or exclude from the project directory in the ROCK image.
+And since the extension places the files from the project folder in the
```
tigarmo, line 35: Hmm a better wording might be:

```
The extension places the files from the project folder in the srv/flask/app`
directory in the final image - therefore, all inclusions and exclusions must
be prefixed with `src/flask/app`.

File: rockcraft/extensions/flask.py [hunk 35]
```diff
@@ -0,0 +1,183 @@
...
+        Parts added:
+            - flask/dependencies: install Python dependencies
+            - flask/install-app: copy the flask project into the OCI image
+        """
+        if not (self.project_root / "requirements.txt").exists():
+            raise ExtensionError(
+                "missing requirements.txt file, "
+                "flask extension requires this file with flask specified as a dependency"
+            )
+        source_files = [f.name for f in sorted(self.project_root.iterdir())]
```
sergiusens, line 127: we can go with this for now, but this extension as is, only serves 12f, as one cannot have https://github.com/flaskfactory/my-flask.git as source entry for building your application

docs/how-to/use-flask-extension.rst
weiiwang01: Updated, thank you! I have also updated the extension based on last week's meeting, changing the base directory to `/flask`, and mandating the `app:app` WSGI path.

tigarmo (approved):

tigarmo:
well done @weiiwang01 !

PR #324, opened 2023-07-13
docs: add warning for platforms field

SamirPS:
This PR add a warning note, about that the host need to be compatible with the arch written on the rockcraft.yaml file.

Signed_Off_By: Samir Akarioh <samir.akarioh@canonical.com>

-----

### DIFF
--- a/docs/reference/rockcraft.yaml.rst
+++ b/docs/reference/rockcraft.yaml.rst
@@ -146,6 +146,11 @@ The set of architecture-specific ROCKs to be built. Supported architectures are:
 Entries in the ``platforms`` dict can be free-form strings, or the name of a
 supported architecture.
 
+.. warning::
+   **All** target architectures must be compatible with the architecture of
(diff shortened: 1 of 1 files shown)


docs/reference/rockcraft.yaml.rst
```diff
@@ -146,6 +146,10 @@ The set of architecture-specific ROCKs to be built. Supported architectures are:
 Entries in the ``platforms`` dict can be free-form strings, or the name of a
 supported architecture.
 
+.. warning::
+   Currently, Rockcraft will only build for an architecture, if the
```
cjdcordeiro, line 150: I think it would read better like this:
```
**All** target architectures must be compatible with the architecture of the host where Rockcraft is being executed (i.e. emulation is not supported at the moment).
```

cjdcordeiro (changes requested):

cjdcordeiro (approved):

dboddie (approved):

lengau (approved):

PR #325, opened 2023-07-14
docs: add a documentation response protocol

cjdcordeiro:
Adds a new section to the docs, explaining how the reader can contribute to the Rockcraft docs by
either submitting an issue or becoming an external contributor (i.e. creating PRs). The response
expectations for both issues and PRs are also
described in this section.

-----


SamirPS (approved):
LGTM, but we need an issue template for documentation before merging this.

cjdcordeiro:
> LGTM, but we need an issue template for documentation before merging this.

ty. that will come in a separate PR. Not yet sure we need one cause you'll either have a documentation "bug" or "request" to make, and those templates are already there. I'll need to see if they fit, and only if they don't I'll then create a separate template

SamirPS:
Okay sound good to me

docs/how-to/contribute-docs.rst
```diff
@@ -83,3 +83,33 @@ setting up a virtual environment and building the documentation.
...
+
+Community contributions are more than welcome. To become an external contributor you
+should:
+
+1. `create a fork`_ of the Rockcraft repository,
+2. commit the changes to your fork (ideally to a new branch),
+    1. make sure to follow the project's `CONTRIBUTING`_ guidelines,
+3. create a Pull Request against the ``main`` branch.
+
+Similarly to new issues, new Pull Requests (PR) are also supervised and regularly
```
lengau, line 113: "Similarly" here is an adjective not an adverb, so either "similar" or "like" would be more technically correct.

It's not a big deal (and I had to check with a friend that I wasn't mistaken), so if you prefer the way it sounds currently that's fine, since good flow is more important than technical-correctness.

lengau (commented):
Looks good, thanks!

docs/how-to/contribute-docs.rst
cjdcordeiro, line 113:
```suggestion
Similar to new issues, new Pull Requests (PR) are also supervised and regularly
```

sergiusens (approved):

dboddie (approved):
Looks OK to me. It could be tidied up a little but that shouldn't block progress.

PR #327, opened 2023-07-17
docs: rewrite the explanation section

### LINKED ISSUE #337: doc: **bold** style not being applied
the `**` syntax doesn't seem to be having any effect. Example:
image

in this screenshot, each bullet should start with a **strong** styled sentence.

---
Document: explanation/rockcraft.rst

SamirPS:
This PR add the explanation of Pebble and
rewrite chisel,rocks and overlay-step explanation.

-----
Fixes #337 
Fixes https://github.com/canonical/rockcraft/issues/348

https://canonical-rockcraft--327.com.readthedocs.build/en/327/


docs/explanation/rocks.rst
```diff
@@ -3,22 +3,62 @@
...
-This means that ROCKs are interoperable with other tools that work with
-OCI-compliant images.
+ROCKs are container images based on Ubuntu designed to meet
+cloud-native software's security, stability, and reliability requirements.
+
+ROCKs were created to address two primary needs. Firstly, it tackles
+the current issues with container image security, as many popular
+container images often have known vulnerabilities.
+Secondly, it aims to provide a smaller container image based on
+Ubuntu to reduce the resource footprint on the host.
```
cjdcordeiro, line 13: This is controversial. 

> Firstly, it tackles the current issues with container image security, as many popular container images often have known vulnerabilities.

How?

---

> it aims to provide a smaller container image based on Ubuntu to reduce the resource footprint on the host.

This isn't true. Alpine and distroless images are still smaller.

docs/explanation/rocks.rst
```diff
@@ -3,22 +3,62 @@
...
 
-A ROCK can be stored in an existing container registry, very much like any
-Docker image, and run in the same way as any other container image.
-For example, Docker can be used to run a ROCK with the usual command line
-syntax:
+These container images comply with the `Open Container
+Initiative`_ (OCI) `image format specification <OCI_image_spec_>`_.
+They can be stored in container registries (e.g. DockerHub, ECR, ACR, Zot,..)
```
cjdcordeiro, line 17: why Zot? sounds like an odd reference out of nowhere...

docs/explanation/rocks.rst
```diff
@@ -3,22 +3,62 @@
...
+Rockcraft
+---------
+
+Rockcraft is a tool designed to build ROCKs using a declarative syntax
+(yaml). It leverages the logic of plugins, parts,and concepts that exist
+in Snapcraft and Charmcraft.
+
+Developers familiar with the creation and publication of snaps and charms
+will be able to utilise existing knowledge to create ROCKS.
+To learn why you need to use Rockcraft see: :ref:`why_use_rockcraft`
```
cjdcordeiro, line 64: sketchy section.

 1. what is the ROCKs ecosystem?
 2. this is repeating content from other pages

basically the same comments as above in the new diagram. This feels out of place and forced into this section.

cjdcordeiro (changes requested):

docs/_static/rockcraft_diagram.jpg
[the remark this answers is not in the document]
SamirPS: Yeah, this diagram simplifies many things; I see your point with Chisel and yeah, the diagram can give this impression. I ping @valentincanonical to have also his advice on it

docs/explanation/rocks.rst
SamirPS: For the first one, by building on top of Ubuntu but also thanks to Chisel.

For the second, I didn't say that ROCK image is smaller than Alpine or Distroless; what gives you this impression?

docs/explanation/rocks.rst
SamirPS: Yeah true, Zot is not a big actor on container registry also

docs/explanation/chisel.rst
```diff
@@ -12,6 +12,49 @@ needed for the ROCK to function properly.
 
 See :ref:`how_to_use_chisel` for information about using the tool.
 
+Slice definitions
```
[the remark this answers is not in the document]
cjdcordeiro: The section can stay IMO. As for Pebble, these are **concepts** that have a place in this documentation. However, we should explain more than just concepts, cause those will not change, unlike the tools' references (like the slice definitions), which might get a new field tomorrow and thus make these docs obsolete.

docs/_static/rockcraft_diagram.jpg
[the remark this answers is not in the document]
cjdcordeiro: feel free to start a shared diagram somewhere (draw.io?) and I'll help out.

You should also talk with @dboddie to understand the best way to go about this. In craft-parts, there is an [old PR](https://github.com/canonical/craft-parts/pull/428) trying to introduce a programmatic way (mermaid) to build such diagrams, so it could be a useful reference.

docs/explanation/rocks.rst
cjdcordeiro: > For the first one, by building on top of Ubuntu but also thanks to Chisel.

but Ubuntu isn't the most secure distribution image out there.

> For the second, I didn't say that ROCK image is smaller than Alpine or Distroless; what gives you this impression?

well your statement claimed that "the goal is to make smaller images", but when you think about it: an alpine-based application image will be smaller than the equivalent ROCK. And a distroless image will also be smaller than the equivalent Chiselled ROCK.

docs/explanation/rocks.rst
cjdcordeiro: maybe it's me. it just feels off. this is typically something where @dboddie should have a say.

SamirPS: I understand your comment, but yeah, let wait for David feedback on this

docs/explanation/rocks.rst
SamirPS: > but Ubuntu isn't the most secure distribution image out there.

Yeah, but it's one of the most secure. Maybe it's the choose of words. I wanted to said that the container will be more secure. 

> well your statement claimed that "the goal is to make smaller images", but when you think about it: an alpine-based application image will be smaller than the equivalent ROCK. And a distroless image will also be smaller than the equivalent Chiselled ROCK.

Yeah, you are right, I think it's a problem of words. I wanted to say that ROCK will give you a smaller image than the Ubuntu one, but not smaller than Alpine or Distroless.

Have some ideas about how I can rephrase these sentences?

docs/explanation/chisel.rst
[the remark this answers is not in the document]
SamirPS: Do you want me to add the concept to the explanation page? Or on the reference section (another PR)?

docs/explanation/rocks.rst
cjdcordeiro: I don't think these are the real values. if you check our ongoing ROCKs@Canonical whitepaper, there are plenty of values, ones stronger than others. 

IMHO, saying that "we are one of the most secure" or "amongst the smaller" is the same as saying that we are aiming to do something that others do better.

For me, the real value and innovation behind Rockcraft is:
 - the declarative format (yes there are other tools too, but unlike security and size, UX is a subjective topic so we can aim to be the best at it)
 - the Ubuntu experience - we eat our own dog food, cause we believe in ubuntu and no one really offer something that can build a truly distro-based distroless image :) this paradox is great, but we actually allow this -> chiselled ROCKs are Ubuntu distroless images :D we broke the convention
 - the consistency - while disregarded by many, one of the problems out there is there every docker image has its own ENTRYPOINT and CMD, so everytime you need to use a new image, to need to get acquainted with it. and the only way to do it is to either read the docs, or inspect the inner contents of the image. In ROCKs, once you get used to Pebble, you get used to every ROCK out there and you'll only need to know the application-specific options

(among others)

docs/explanation/chisel.rst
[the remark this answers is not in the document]
SamirPS: The pebble one?

docs/explanation/rocks.rst
SamirPS: Thanks, i have updated this section; tell me if you find this better

docs/explanation/rocks.rst
dboddie: I think maybe "ecosystem" isn't quite the word we are looking for. Perhaps "toolchain" is closer to the intended meaning.
It's good to discuss what makes a ROCK, and also which tools we use to work with ROCKs.

docs/explanation/chisel.rst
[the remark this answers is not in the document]
dboddie: I think perhaps @cjdcordeiro was referring to the chisel explanation. My only suggestion is to move the "Slice definitions" section to the end of the document.

docs/_static/rockcraft_diagram.jpg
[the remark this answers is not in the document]
SamirPS: Does the first one mainly the one from @cjdcordeiro?

docs/explanation/rocks.rst
dboddie: I think this is where the lines are blurred between ROCKs and Rockcraft. These list items are very similar to those in the "Why use Rockcraft?" document.

What we need at the start of this document (rocks.rst) is a description of what ROCKs are. Something like this:
> ROCKs are OCI-compliant container images, built on an Ubuntu base, with a consistent entry point.

There is a certain amount of unpacking that needs to happen with that sentence, which can be done in later sentences/paragraphs.
We can also say why it is important that a ROCK has those properties. Some of this is already present in the text:

> ROCKs comply with the [Open Container Initiative](https://opencontainers.org/) (OCI) [image format specification](https://github.com/opencontainers/image-spec/blob/main/spec.md). ROCKs can be stored in container registries (e.g. DockerHub, ECR, ACR,..) and used with any OCI-Compliant tools (e.g. Docker, Podman, Kubernetes,…).

We can then get into the details of what a ROCK contains and how one is built. Perhaps the details of how one is built could be moved into another document later.

docs/explanation/chisel.rst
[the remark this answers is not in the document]
cjdcordeiro: yep. I don't think this file should be changed at all. if not for the content duplication risk, then for the fact that "slice definitions" are somewhat irrelevant for a rockstar who's building rocks, and not creating slice definitions files

docs/_static/rockcraft_diagram.jpg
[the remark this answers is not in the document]
cjdcordeiro: 1. we can do. Please note though that **it will be an advanced diagram**. I.e. it will be an enhancement of the one I pasted above, alluding to OCI definitions that a typical container user wouldn't need to know. Not sure that it a good or bad thing, but from experience, many "advanced" docker users are clueless about the OCI spec, cause that's given
2. yep, definitely this one
3. could be, although I'm not sure what it would look like...I'm a bit skeptical about correlating ROCKs with Chisel, cause that's a tool. If we go down that path it actually makes much more sense to correlate it with Pebble, cause that's an integral part of all ROCKs.

docs/explanation/rocks.rst
cjdcordeiro: "toolchain" sounds better. Given that the Explanation section aggregates explanations for all of those tools, I wonder if these short explanations should be placed in `explanation/index.rst`?

docs/_static/rockcraft_diagram.jpg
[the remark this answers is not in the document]
dboddie: I see the point about chisel. Pebble should be mentioned in preference, I think.

docs/explanation/pebble.rst
```diff
@@ -0,0 +1,33 @@
+.. _pebble_explanation_page:
+
+Pebble
+======
+
+When creating container images using Dockerfile, several issues can be met.
```
dboddie, line 6: Perhaps we can start with saying what Pebble is:

> Pebble is a service manager that helps to organise a set of local service processes.

cjdcordeiro: done :)

docs/explanation/pebble.rst
```diff
@@ -3,13 +3,37 @@
...
+
+Features
+~~~~~~~~
+
 
 When creating container images, several issues can be met.
 For instance, if the entrypoint relies on the application mode
 (such as nginx and nginx-debug), creating a bash script that parses all the
-arguments provided during container deployment will be necessary.
+arguments provided during container deployment will be necessary./
```
sergiusens: what's this?

cjdcordeiro: (outdated)

docs/explanation/rocks.rst
cjdcordeiro: updated. pls re-review

docs/_static/css/custom.css
```diff
@@ -1,28 +1,3 @@
-@import url('https://fonts.googleapis.com/css2?family=Ubuntu:ital@0;1&display=swap');
```
SamirPS, line 1: To know why you remove all this CSS?

docs/explanation/chisel.rst
```diff
@@ -44,3 +44,14 @@ could see as a block of stone from which we can carve and extract only the
 small and relevant parts that we need to run our applications, thus keeping
 ROCKs small and less exposed to vulnerabilities.
 
+Slice definitions
+~~~~~~~~~~~~~~~~~
+
+A package's slices can be defined via a YAML slice definitions file. Check the
```
SamirPS, line 50: I feel this need to be done before talking about slice because this section is called ``slice definitions. ``

docs/explanation/chisel.rst
```diff
@@ -44,3 +44,14 @@ could see as a block of stone from which we can carve and extract only the
...
+`slice definitions reference
+<https://github.com/canonical/chisel/#slice-definitions>`_ for more information
+about this file's format.
+
+To find examples of existing slice definitions files, check the Chisel releases
+repository at `<https://github.com/canonical/chisel-releases>`_. Contributions
```
SamirPS, line 56: maybe put the last sentence in bold or in note format

docs/explanation/pebble.rst
```diff
@@ -0,0 +1,85 @@
...
+with PID=1).
+
+Multiple processes in a container?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Containers' best practices advocate the separation of concerns and the adoption
+of a single service per container. With the introduction of `Pebble`_ as the
+ROCKs' entrypoint, this principle is elevated to new heights:
+
+  *if multiple processes rely on shared dependencies and are tightly coupled
```
SamirPS, line 24: The `Pebble`_ not display like you want i think

docs/explanation/pebble.rst
```diff
@@ -0,0 +1,85 @@
...
+processes. This is one of the main reasons behind the gradual shift in best
+practices, as there is an increasing emphasis on adopting init processes such
+as `tini`_, `s6-overlay`_, or `Pebble`_.
+
+What to expect?
+~~~~~~~~~~~~~~~
+
+* **client-server model behind a single binary**: Pebble is injected into
+  ROCKs as a single binary which acts both as a daemon and a client to itself;
+* **declarative service definition**: the Pebble service processes (or simply
```
SamirPS, line 41: i see the point for the  ``cleaner`` but for the ``error-prone``can you detail more ?

docs/explanation/pebble.rst
```diff
@@ -0,0 +1,85 @@
...
+  *Pebble services*) are declaratively defined in YAML files called layers;.
+  When compared to nowadays use of `imperative wrapper scripts (as suggested
+  in the Docker documentation)`_, this provides a much cleaner and less
+  error-prone way for defining the processes that should run inside the
+  container;
+* **services as first-class citizens**: in contrast to wrapper scripts, Pebble
+  treats services as manageable units with a defined lifecycle and
+  service-specific definitions for health monitoring, inter-service
+  dependencies, restart policies, and much more;
+* **layering**: Pebble can stack multiple layers into a single Pebble plan
```
SamirPS, line 51: what is a pebble layer ?

docs/explanation/pebble.rst
```diff
@@ -0,0 +1,85 @@
...
+services are defined in :doc:`/reference/rockcraft.yaml` and you can also see
+:doc:`/how-to/convert-to-pebble-layer` for an example of how to convert a
+Docker entrypoint to a Pebble layer.
+
+
+.. _Pebble: https://github.com/canonical/pebble
+.. _tini: https://github.com/krallin/tini
+.. _s6-overlay: https://github.com/just-containers/s6-overlay
+.. _imperative wrapper scripts (as suggested in the Docker documentation): https://docs.docker.com/config/containers/multi-service_container/#use-a-wrapper-script
+.. _Pebble layer specification: https://github.com/canonical/pebble#layer-specification
```
SamirPS, line 85: I don't know if it's me but reading this I don't really know why should I use Pebble and not another service you listed like [tini](https://github.com/krallin/tini)

docs/explanation/rocks.rst
```diff
@@ -3,22 +3,49 @@
 ROCKs
 =====
 
-ROCKs are container images that are compliant with the `Open Container
-Initiative`_ (OCI) `image format specification <OCI_image_spec_>`_.
-This means that ROCKs are interoperable with other tools that work with
-OCI-compliant images.
+ROCKs are **Ubuntu LTS-based container images** that are designed to meet
+cloud-native software's security, stability, and reliability requirements.
```
SamirPS, line 7: Does the ubuntu image is one of the more secure and stable?  because for me, the sentence sounds like this

docs/explanation/rocks.rst
```diff
@@ -3,22 +3,49 @@
...
+* **User-centric experience**: ROCKs are described in a :ref:`declarative
+  format<rockcraft.yaml_reference>` and **built on top of familiar and reliable
+  Ubuntu LTS images**, thus offering an open and up-to-date user experience;
+* **Seamless chiselling experience**: ROCKs can be effortlessly
+  :ref:`chiselled<chisel_explanation>` using off-the-shelf primitives,
+  harnessing all the advantageous traits of "distroless" to deliver **compact
+  and secure Ubuntu-based container images**.
+
+
+Design
```
SamirPS, line 40: I think this section maybe going too fast, for a beginner who know nothing about OCI image you give too much information in 4 lines.

SamirPS: Also i think adding somewhere the build diagram you have done is a good idea

docs/explanation/chisel.rst
cjdcordeiro: most people will be users of slices and thus never actually define slices. the previous section introduces the concept, while this one drills down into how to define slices. 

"Package slices (what are these?) -> Slice definitions (learn how to create a slice)" seems like a logical flow to me and thus why the latter became a subsection of the former

docs/explanation/pebble.rst
cjdcordeiro: `cleaner` and `less error-prone` are both advantages of a declarative format. with an imperative format, it is much easier to miss a symbol or a character in your script, causing the entire entrypoint to misbehave.

docs/explanation/pebble.rst
SamirPS: i was talking about this sentence sorry:

```
if multiple processes rely on shared dependencies and are tightly coupled together (i.e. they serve a single purpose and cannot be executed independently), then the best practice entails orchestrating them within the same container, with `Pebble`_ as their manager.
```

docs/explanation/pebble.rst
cjdcordeiro: the reasons are listed in the `container-optimised init process` bullet above. I'll make the comparison more obvious

docs/explanation/rocks.rst
cjdcordeiro: I've adjusted the section. Please have another look.

As for the diagram, we can discuss that in a future PR, but I feel it belongs in the Rockcraft section.

docs/explanation/pebble.rst
cjdcordeiro: ah yes, sry. good catch. fixed

docs/explanation/chisel.rst
SamirPS: It's maybe the name of the section which confuses me. I think it's better to have @dboddie feedback on this.

docs/explanation/rocks.rst
SamirPS: i see thanks

docs/explanation/chisel.rst
```diff
@@ -44,3 +44,15 @@ could see as a block of stone from which we can carve and extract only the
...
 
+Slice definitions
+~~~~~~~~~~~~~~~~~
+
+A package's slices can be defined via a YAML slice definitions file. Check the
+`slice definitions reference
+<https://github.com/canonical/chisel/#slice-definitions>`_ for more information
+about this file's format.
+
+.. note::
```
SamirPS, line 55: i think only the ``Contributions are welcome and encouraged.`` need to be in note

cjdcordeiro: well, a "note" is for making a **remark** about something. I don't think that fits

SamirPS:
ping @pedroleaoc for review @cjdcordeiro text

Makefile
```diff
@@ -41,7 +42,8 @@ coverage: ## Run pytest with coverage report.
 .PHONY: preparedocs
 preparedocs: ## move file from the sphinx-starter-pack to docs folder
 	cp docs/sphinx-starter-pack/.sphinx/_static/* docs/_static
-	cp -R docs/sphinx-starter-pack/.sphinx/_templates docs/_templates
+	mkdir docs/_templates || true
```
tigarmo, line 45: is the `|| true` for the case where `docs/_templates` already exists? If so, you can use `mkdir -p docs/_templates` instead, which has the advantage of not failing if the dir already exists but failing for any other problem, like 'permission denied'

docs/_static/rock_diagram.png
tigarmo, line 1: this image is very nice :+1:

docs/explanation/pebble.rst
```diff
@@ -0,0 +1,88 @@
...
+
+Pebble
+======
+
+.. important::
+    **Pebble is the default entrypoint for all ROCKs!**
+
+Similar to other well-known process managers such as *supervisord*, *runit*, or
+*s6*, `Pebble`_ is a service manager that enables the seamless orchestration of
+a collection of local service processes as organised set. The main difference
```
tigarmo, line 11:
```suggestion
a collection of local service processes as an organised set. The main difference
```

docs/explanation/rocks.rst
```diff
@@ -3,22 +3,68 @@
...
-   docker run <rock> ...
+What sets ROCKs apart?
+~~~~~~~~~~~~~~~~~~~~~~
 
-Interoperability between ROCKs and other containers also extends to the way
-that container images are built. This enables the use of ROCKs as bases for
-existing build recipes, such as Dockerfiles, for further customisation and
-development.
+* **Opinionated and consistent design**: all ROCKs follow the same design,
+  aiming to minimise your full-stack disparity and adoption overhead, i.e
```
tigarmo, line 22:
```suggestion
  aiming to minimise your full-stack disparity and adoption overhead, e.g.
```

I think you meant to use pebble as an example, right?

tigarmo (commented):
thanks!

Makefile
cjdcordeiro: tbh, I don't remember why I did that xD

but your suggestion is correct. I'll change it

cjdcordeiro:
thanks for the review @tigarmo 

I've updated the PR. Please note that this PR also contains a fix for the currently broken docs header styling -> https://canonical-rockcraft.readthedocs-hosted.com/en/latest/

Makefile
```diff
@@ -119,7 +121,7 @@ test-pylint:
 
 .PHONY: test-pyright
 test-pyright:
-	pyright .
+	pyright .
```
sergiusens, line 124:
```suggestion
	pyright .
```

docs/.wordlist.txt
```diff
@@ -1,77 +1,93 @@
+ACR
```
sergiusens, line 1: are we now sorting ignoring case?

cjdcordeiro: https://github.com/canonical/sphinx-docs-starter-pack/blob/main/.wordlist.txt

sergiusens: I did a previous PR with a case sensitive sort and removed lots of duplicates, just asking if this was on purpose, and it's fine, not asking for a change or anything

cjdcordeiro: right. I see the diff. I think that re-arrangement was committed by @SamirPS , so I'm not 100% sure if there's is an underlying reason. I'm also fine either way...

SamirPS: no reason, it's was just a sort

tigarmo (approved):

lengau (approved):
This looks great! (And please keep them coming — reviewing documentation PRs is honestly one of the best ways to keep keep me up to date on the state of Rockcraft, so I'm grateful for the opportunity.)

The comments I've added below are with my mental "grammar and style knobs" tuned to maximum sensitivity. None of them are strictly necessary, but I'd rather offer 10 suggestions where you prefer the original for 9 of them than miss an improvement you'd like too. So feel free to pick and choose only the suggestions you personally prefer ­— even if that turns out to be none of them.

docs/explanation/rocks.rst
```diff
@@ -3,22 +3,68 @@
...
+  * :ref:`pebble_explanation_page` **is the official entrypoint for all
+    ROCKs**, thus providing a predictable and powerful abstraction layer
+    between the user and the container application;
+  * ROCKs extend the OCI image information by including additional **metadata**
+    inside each ROCK (e.g. at ``/.rock/metadata.yaml``), allowing container
+    applications to easily inspect the properties of the image they are running
+    on, at execution time;
+* **User-centric experience**: ROCKs are described in a :ref:`declarative
+  format<rockcraft.yaml_reference>` and **built on top of familiar and reliable
+  Ubuntu LTS images**, thus offering an open and up-to-date user experience;
```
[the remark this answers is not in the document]
cjdcordeiro: it seems that I like to use `thus` a lot :rofl:

docs/explanation/pebble.rst
```diff
@@ -0,0 +1,88 @@
...
+  together (i.e. they serve a single purpose and cannot be executed
+  independently), then the best practice entails orchestrating them within the
+  same container, with Pebble as their manager.*
+
+This new notion addressed existing pain points arising from the excessive
+separation of concerns, which results in numerous container images whose
+entrypoints lack the ability to gracefully handle the underlying child
+processes. This is one of the main reasons behind the gradual shift in the best
+practices, as there is an increasing emphasis on adopting init processes such
+as `tini`_, `s6-overlay`_, or `Pebble`_.
```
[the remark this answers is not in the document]
cjdcordeiro: fair enough. I'd rather have it here though as it flows well with the above text, IMO

cjdcordeiro (approved):

PR #329, opened 2023-07-18
feat: forbid string interpolation

cjdcordeiro:

-----

With the recent introduction of `environment`, users might expect string interpolation to work such that the following would be valid:

```yaml
environment:
  FOO: BAR
  BAZ: $FOO
```

but **it is not**. The YAML loading takes all map values as literal strings and we do not yet have any string interpolation for variable substitution.

We will though, but while that feature is being designed and implemented, it is safer to limit the user assumptions and forbid the use of bash-like variables in the `environment` key values.


lengau (approved):

tests/unit/test_project.py
```diff
@@ -197,6 +197,26 @@ def test_project_unmarshal_with_unsupported_fields(unsupported_field, yaml_loade
     )
 
 
+@pytest.mark.parametrize(
+    "variable,is_forbidden",
+    [("$BAR", True), ("BAR_$BAZ", True), ("BAR$", False)],
```
SamirPS, line 202: How ``BAR$`` is not forbidden ?

cjdcordeiro: why should it be? if it is the last char, then it is just a regular char.

SamirPS: Yeah true,  sorry

cjdcordeiro: ;)

mr-cal (approved):

sergiusens (approved):

PR #330, opened 2023-07-19
Provide a default service for the Hello world example

valentincanonical:
Provide a default service for the Hello world example. Should probably be done with other examples as well.

-----

### DIFF
--- a/docs/tutorials/code/hello-world/rockcraft.yaml
+++ b/docs/tutorials/code/hello-world/rockcraft.yaml
@@ -11,8 +11,13 @@ build-base: ubuntu:22.04
 platforms:
   amd64:  # Make sure this value matches your computer's architecture
 
-# Parts section
+services:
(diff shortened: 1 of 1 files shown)


cjdcordeiro:
this is a bit controversial. 

Pebble services do not support one-shot commands like `hello`, thus why such examples don't have any services. It is completely fine to have a ROCK without any pebble services, if the desired execution mode is leveraging one-shot instructions via `pebble exec`. having short-lived commands associated with pebble services will cause ROCKs to behave in a weird way since pebble's service restart policies and health checks will be ignored (cause the service won't live for enough time).

Useful reference:
 - https://github.com/canonical/pebble/issues/240

PS.: I do think we should support this though, but it is something that needs to be changed on the Pebble side.

sergiusens:
closing until controversy is resolved

PR #331, opened 2023-07-24
add support for health checks

cjdcordeiro:

-----

In this PR:
 - move all Pebble-related schemas into the Pebble module
 - add the schema for Pebble checks (based on https://github.com/canonical/pebble/#layer-specification)
 - change the lifecycle logic to encompass the new `checks` field alongside the `services` field, when creating the Pebble layer
 - add new field to reference documentation

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,6 +11,9 @@ line_length = 88
 # are not overwritten.
 extend-exclude = "docs/sphinx-starter-pack"
 
+[tool.pyright]
+ignore = ["docs/sphinx-starter-pack"]
+
 [tool.mypy]
 python_version = "3.8"
 exclude = [
--- a/rockcraft/lifecycle.py
+++ b/rockcraft/lifecycle.py
@@ -199,9 +199,23 @@ def _pack(
     emit.progress("Adding Pebble entrypoint")
 
     new_image.set_entrypoint()
-    if project.services:
-        new_image.set_pebble_services(
-            services=project.dict(exclude_none=True, by_alias=True)["services"],
+
+    services = (
+        project.dict(exclude_none=True, by_alias=True)["services"]
+        if project.services
+        else {}
+    )
+
+    checks = (
+        project.dict(exclude_none=True, by_alias=True)["checks"]
+        if project.checks
+        else {}
+    )
+
+    if services or checks:
+        new_image.set_pebble_layer(
+            services=services,
+            checks=checks,
             name=project.name,
             tag=project.version,
             summary=project.summary,
--- a/rockcraft/lifecycle.py
+++ b/rockcraft/lifecycle.py
@@ -200,17 +200,9 @@ def _pack(
 
     new_image.set_entrypoint()
 
-    services = (
-        project.dict(exclude_none=True, by_alias=True)["services"]
-        if project.services
-        else {}
-    )
+    services = project.dict(exclude_none=True, by_alias=True).get("services", {})
 
-    checks = (
-        project.dict(exclude_none=True, by_alias=True)["checks"]
-        if project.checks
-        else {}
-    )
+    checks = project.dict(exclude_none=True, by_alias=True).get("checks", {})
 
     if services or checks:
         new_image.set_pebble_layer(
--- a/rockcraft/oci.py
+++ b/rockcraft/oci.py
@@ -370,10 +370,10 @@ class Image:
         :param base_layer_dir: Path to the base layer's root filesystem
         """
         # pylint: disable=too-many-arguments
-        pebble_layer_content = {
+        pebble_layer_content: Dict[str, Any] = {
             "summary": summary,
             "description": description,
-        }  # type: Dict[str, Any]
+        }
 
         if services:
             emit.progress(
--- a/rockcraft/oci.py
+++ b/rockcraft/oci.py
@@ -349,18 +349,20 @@ class Image:
         _config_image(image_path, ["--clear=config.cmd"])
         emit.progress(f"Entrypoint set to {entrypoint}", permanent=True)
 
-    def set_pebble_services(
+    def set_pebble_layer(
         self,
         services: Dict[str, Any],
+        checks: Dict[str, Any],
         name: str,
         tag: str,
         summary: str,
         description: str,
         base_layer_dir: Path,
     ) -> None:
-        """Write the provided services into a Pebble layer in the filesystem.
+        """Write the provided services and checks into a Pebble layer in the filesystem.
 
-        :param services: The Pebble services to be written into the Pebble
+        :param services: The Pebble services
+        :param checks: The Pebble checks
         :param name: The name of the ROCK
         :param tag: The ROCK's image tag
         :param summary: The summary for the Pebble layer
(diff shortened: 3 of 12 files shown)


rockcraft/lifecycle.py
```diff
@@ -193,9 +193,23 @@ def _pack(
     emit.progress("Adding Pebble entrypoint")
 
     new_image.set_entrypoint()
-    if project.services:
-        new_image.set_pebble_services(
-            services=project.dict(exclude_none=True, by_alias=True)["services"],
+
+    services = (
+        project.dict(exclude_none=True, by_alias=True)["services"]
```
tigarmo, line 198: (minor) `project.dict(exclude_none=True, by_alias=True).get("services", {})` might be clearer
or we could just call `project.dict()` once and use the result for both `services` and `checks`

tests/unit/test_pebble.py
```diff
@@ -180,3 +182,236 @@ def test_define_pebble_layer(
...
+                r"value is not a valid integer[\s\S]*"
+                r"unhashable type[\s\S]*"
+                r"value is not a valid float[\s\S]*",
+            ),
+        ],
+    )
+    def test_bad_services(self, bad_service, error):
+        with pytest.raises(pydantic.ValidationError) as err:
+            _ = Service(**bad_service)
+        assert re.match(error, str(err.value)), "Unexpected Service validation error"
```
tigarmo, line 266: I think `pytest.raises(pydantic.ValidationError, match=error)` will also work
edit: ah I see now that this test was moved

docs/tutorials/code/hello-world/rockcraft.yaml
```diff
@@ -1,14 +1,20 @@
...
   amd64:  # Make sure this value matches your computer's architecture
 
+# Parts section
+
 parts:
   hello:
     plugin: nil
-    overlay-packages:
-      - hello
+    stage-packages:
```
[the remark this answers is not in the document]
cjdcordeiro: indeed, my bad. should be alright now

rockcraft/lifecycle.py
cjdcordeiro: thanks for noticing. changed it

tests/unit/test_pebble.py
cjdcordeiro: nice, didn't know that. I've applied it to all similar tests in this file.

tests/spread/general/health-checks/rockcraft.yaml
```diff
@@ -0,0 +1,61 @@
...
+base: "ubuntu:22.04"
+
+services: 
+  webserver:
+    override: replace
+    command: timeout 30 nginx -g 'daemon off;'
+    environment:
+      TZ: UTC
+    startup: enabled
+    on-failure: shutdown
```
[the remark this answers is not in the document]
cjdcordeiro: the container shuts down as well.

tigarmo (approved):

rockcraft/oci.py
```diff
@@ -349,32 +349,41 @@ def set_entrypoint(self) -> None:
...
         """
         # pylint: disable=too-many-arguments
-        service_names = list(services.keys())
-        emit.progress(f"Configuring Pebble services {', '.join(service_names)}")
         pebble_layer_content = {
             "summary": summary,
             "description": description,
-            "services": services,
-        }
+        }  # type: Dict[str, Any]
```
lengau, line 376: Linters are soon going to start complaining about type comments other than `type: ignore`, so it might be better to say `pebble_layer_content: Dict[str, Any] = {` above.

(I would have put this in as a suggestion but GitHub is being weird about suggestions here.)

rockcraft/pebble.py
```diff
@@ -18,11 +18,142 @@
...
 import yaml
 from craft_cli import emit
 
+from rockcraft.errors import ProjectValidationError
+
+
+class HttpCheck(pydantic.BaseModel):
+    """Lightweight schema validation for a Pebble HTTP check."""
+
+    url: str
```
lengau, line 33:
```suggestion
    url: pydantic.AnyHttpUrl
```

Not sure if you want to be [that specific](https://docs.pydantic.dev/1.10/usage/types/#urls) though.

rockcraft/pebble.py
```diff
@@ -18,11 +18,142 @@
...
+
+        if count > 1:
+            err = str(
+                f"Only one of {', '.join(mutually_exclusive)} "
+                "may be specified for each check."
+            )
+        elif count < 1:
+            err = str(
+                f"Must specify one of {', '.join(mutually_exclusive)} for each check."
+            )
```
lengau, line 107:
```suggestion
        mutually_exclusive = {"http", "tcp", "exec"}
        check_types = mutually_exclusive & values.keys()

        if len(check_types) > 1:
            err = str(
                f"Multiple check types specified ({', '.join(check_types)}). "
                "Each check must have exactly one type."
            )
        elif len(check_types) < 1:
            err = str(
                f"Must specify exactly one of {', '.join(mutually_exclusive)} for each check."
            )
```

This is a combination of two suggestions. I'm writing each out here so you can decide whether you want one, both, or neither:

1. I changed the error wording to specify that they must have exactly one type to hopefully prevent two iterations (no check, two checks, one check). I'm not sure whether this is a concern in practice though.
2. It looks like a user could (theoretically, at least — again, not sure how practical it is) have many checks, so telling them which check types are specified could help them find the incorrectly-specified check.

tests/spread/general/health-checks/rockcraft.yaml
```diff
@@ -0,0 +1,61 @@
...
+
+package-repositories:
+  - type: apt
+    url: https://nginx.org/packages/mainline/ubuntu
+    key-id: 573BFD6B3D8FBC641079A6ABABF5BD827BD9BF62
+    suites: 
+      - jammy
+    components:
+      - nginx
+    priority: always
```
lengau, line 46: Out of curiosity: why did you add this repository rather than using the nginx package in jammy main? 

(I'm guessing there are either packaging differences, something else you're trying to test, or that you're just more comfortable with that package. All three are totally valid, I'm just curious.)

lengau (approved):
I have some questions and some minor suggestions, but it also looks great as-is.

pyproject.toml
```diff
@@ -11,6 +11,9 @@ line_length = 88
 # are not overwritten.
 extend-exclude = "docs/sphinx-starter-pack"
 
+[tool.pyright]
+ignore = ["docs/sphinx-starter-pack"]
```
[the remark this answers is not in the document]
cjdcordeiro: y it's a good point. My rationale is that _it is better to gradually ignore rather than ignore the whole folder from the get-go_. We don't own the submodule so that's a no-brainer, but for the rest, I would leave it under scope for now and only ignore it in the future if it becomes an unnecessary nuisance.

rockcraft/oci.py
cjdcordeiro: nice one. will do.

rockcraft/pebble.py
cjdcordeiro: lgtm, thanks

tests/unit/test_pebble.py
```diff
@@ -180,3 +181,231 @@ def test_define_pebble_layer(
...
+            "on-check-failure": {"check": "restart"},
+            "backoff-delay": "10ms",
+            "backoff-factor": 1.2,
+            "backoff-limit": "1m",
+            "kill-delay": "5s",
+        }
+        _ = Service(**full_service)
+
+    def test_minimal_service(self):
+        _ = Service(override="merge", command="foo cmd")  # pyright: ignore
```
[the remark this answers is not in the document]
cjdcordeiro: why not :)

tests/spread/general/health-checks/rockcraft.yaml
cjdcordeiro: this is to mimic the upstream build recipe, from nginx -> https://github.com/nginxinc/docker-nginx/blob/1a8d87b69760693a8e33cd8a9e0c2e5f0e8b0e3c/stable/debian/Dockerfile.

iirc, the way of installing it via the standard jammy archives is slightly different, so this was just easier to inherit from our existing tutorials

Issue #333, opened 2023-07-25
global environment not being registered for CRAFT_PROJECT_VERSION


cjdcordeiro:
### Bug Description

The `CRAFT_PROJECT_VERSION` depends on the rockcraft.yaml's `version` field and thus should be registered after the YAML is loaded.

This should happen in 

https://github.com/canonical/rockcraft/blob/bdf64517487be2fac8fa28ddfe3919e08f9b3bde/rockcraft/lifecycle.py#L70

however, the callback does not seem to be working and this environment variable never gets loaded.

### To Reproduce

Run `rockcraft --verbosity debug` on https://github.com/jnsgruk/zinc-k8s-operator/blob/main/rockcraft.yaml (originally reported by @jnsgruk in https://chat.charmhub.io/charmhub/pl/h9qytisgti8s3pu6yhe1pq8hpy).

### Environment

Version: `0+git.bdf6451`

### rockcraft.yaml

```shell
name: zinc
version: 0.4.7
license: Apache-2.0

base: bare
build_base: ubuntu:22.04
platforms:
  amd64:

summary: Zinc is a lightweight search engine and alternative to Elasticsearch.
description: |
  Zinc is a search engine that does full text indexing. It is a lightweight alternative to
  elasticsearch and runs in less than 100 MB of RAM. It uses bluge as the underlying indexing
  library.

  It is very simple and easy to operate as opposed to elasticsearch which requires a couple dozen
  knobs to understand and tune.

  It is a drop-in replacement for elasticsearch if you are just ingesting data using APIs and
  searching using kibana (Kibana is not supported with zinc. Zinc provides its own UI).

run_user: _daemon_

environment:
  ZINC_DATA_PATH: /var/lib/zincsearch
  ZINC_PROMETHEUS_ENABLE: true
  ZINC_TELEMETRY: false
  ZINC_PROFILER: true

services:
  zinc:
    override: replace
    startup: enabled
    command: /bin/go-runner --log-file=/var/lib/zincsearch/zinc.log --also-stdout=true --redirect-stderr=true /bin/zincsearch

parts:
  zinc:
    plugin: go
    source: https://github.com/zincsearch/zincsearch
    source-type: git
    source-tag: v$CRAFT_PROJECT_VERSION
    build-snaps:
      - go/latest/stable
      - node/18/stable
    build-environment:
      - CGO_ENABLED: 0
      - GOOS: linux
    override-build: |
      COMMIT_HASH="$(git rev-parse HEAD)"
[...]

PR #336, opened 2023-07-26
fix global environment expansion

cjdcordeiro:

-----

Fixes https://github.com/canonical/rockcraft/issues/333

In this PR:
 - fix the global variable expansion which was not being called and thus resulting in `CRAFT_` variables not being set before running the parts' lifecycle steps
 - refactor the project loading so that it returns a YAML instead of the project itself. The latter can still be retrieved by explicitly calling the `unmarshal` function
 - adjust the unit tests accordingly
 - add `CRAFT_PROJECT_VERSION` to a spread test

**NOTE:** Although it works, I feel there might be a better way to handle this, by fiddling with the callbacks and `ProjectInfo` directly in the craft-parts module. Why can't the variable expansion be handled by the craft-parts, implicitly? I.e. the `PartsLifecycle` already takes `project_vars` as an argument, so it doesn't need access to the whole YAML


jnsgruk:
Tested this locally and confirmed that it fixes #333. Thanks @cjdcordeiro!

tests/unit/test_lifecycle.py
```diff
@@ -28,6 +28,23 @@
 pytestmark = pytest.mark.usefixtures("reset_callbacks")
 
 
+@pytest.fixture
```
jnsgruk, line 31: Some of the changes in this file seem unrelated to the environment changes?

jnsgruk (approved):

tests/unit/test_lifecycle.py
cjdcordeiro: it's a consequence. This fix implies actually having a `Project` method for getting the YAML loaded as a dict. As a result, I've changed the `return` type of our current project handling functions and thus that cascaded into a change to many unrelated tests

cjdcordeiro: for example: https://github.com/canonical/craft-parts/blob/ed49f727e60447d491f7d7b58b03711c2e6a7bfd/craft_parts/executor/environment.py#L204 expects a Dict, but we use pydantic for project validation, which mean we end up packing the ROCK with all the project info under a class object.

jnsgruk: No probs! :+1:

sergiusens (approved):
Looks good, I would like @cmatsuoka to review

cjdcordeiro:
> Looks good, I would like @cmatsuoka to review

ty @sergiusens 

what are your thoughts on my comments above about refactoring the way this is done, via craft-parts? It just feels clunky like this...

cmatsuoka (approved):
This follows the same pattern currently used in Snapcraft. Creating a `ProjectInfo` instance by hand to expand environment variables is a bit of a hack, but we can tackle this issue when rebasing on craft-application.

cjdcordeiro:
> This follows the same pattern currently used in Snapcraft. Creating a `ProjectInfo` instance by hand to expand environment variables is a bit of a hack, but we can tackle this issue when rebasing on craft-application.

that's exactly the hack I'm a bit uncomfortable with, but it appear you have a plan :) ty for the approval. re-triggering a test and then merging

PR #354, opened 2023-09-18
refactor: use craft-application for "pack"

### LINKED ISSUE #316: Rebase "rockcraft pack" onto craft-application framework
### What needs to get done

The craft-application framework reduces repeated code from various *craft applications. Rockcraft should be rebased onto craft-application before we get a stable release.

### Why it needs to get done

Reduce repeated code throughout the *craft applications.

tigarmo:
This PR makes it so invocations of `rockcraft pack` get routed to the new craft-application-based "runner". This is a lot of code because "pack" basically needs everything that we do, but a lot of the code is copy/pasted from the now-legacy runner.

It is _strongly suggested_ to review per-commit

Fixes #316


pyproject.toml
```diff
@@ -44,7 +44,7 @@ module = ["craft_archives"]
 ignore_missing_imports = true
 
 [tool.pylint.messages_control]
-disable = "too-many-ancestors,too-few-public-methods,fixme,unspecified-encoding,use-implicit-booleaness-not-comparison,unnecessary-lambda-assignment,line-too-long"
+disable = "too-many-ancestors,too-few-public-methods,fixme,unspecified-encoding,use-implicit-booleaness-not-comparison,unnecessary-lambda-assignment,line-too-long,cyclic-import"
```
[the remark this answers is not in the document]
tigarmo: yeap. No idea why, the app itself is working fine and the other linters don't complain either

tests/unit/test_project.py
```diff
@@ -592,3 +588,52 @@ def test_project_generate_metadata(yaml_loaded_data):
...
+                    build_for="amd64",
+                    base=BaseName(name="ubuntu", version="20.04"),
+                    platform_entry="amd64",
+                    build_for_variant=None,
+                ),
+            ],
+        ),
+    ],
+)
+def test_project_get_build_plan(yaml_loaded_data, platforms, expected_build_infos):
```
cmatsuoka, line 636: Maybe also test with platform names that don't match an architecture name?

rockcraft/services/image.py
```diff
@@ -0,0 +1,101 @@
...
+from pathlib import Path
+from typing import cast
+
+from craft_application import AppMetadata, BaseService, ServiceFactory
+from craft_cli import emit
+
+from rockcraft import models, oci
+
+
+@dataclass
```
cmatsuoka, line 31: Can we set this as frozen?

rockcraft/services/lifecycle.py
```diff
@@ -0,0 +1,98 @@
...
+
+class RockcraftLifecycleService(LifecycleService):
+    """Rockcraft-specific lifecycle service."""
+
+    @override
+    def setup(self) -> None:
+        """Configure extra args to the LifecycleManager."""
+        # pylint: disable=import-outside-toplevel
+        # This inner import is necessary to resolve a cyclic import
+        from rockcraft.services import RockcraftServiceFactory
```
cmatsuoka, line 43: Would it work if we import at toplevel, but protected with `TYPE_CHECKING`?

rockcraft/services/lifecycle.py
```diff
@@ -0,0 +1,98 @@
...
+            package_repositories=project.package_repositories or [],
+            project_name=project.name,
+            project_vars=project_vars,
+        )
+
+        super().setup()
+
+    @override
+    def run(self, step_name: str, part_names: list[str] | None = None) -> None:
+        """Overridden to configure package repositories."""
```
sergiusens, line 65: this feels more like a code comment, the override decorator takes care of the docstring (one of its uses)

tests/unit/test_project.py
tigarmo: oh of course! Done

rockcraft/services/image.py
tigarmo: done! I tried to set `RockcraftBuildInfo` also as frozen but then the base `BuildInfo` needs to be frozen too, which it probably right

rockcraft/services/lifecycle.py
tigarmo: > the override decorator takes care of the docstring (one of its uses)

I thought so too but pydocstyle is complaining that the method doesn't have a docstring. I ended up duplicating the docstring and adding this as a code comment

rockcraft/services/lifecycle.py
tigarmo: Interesting. If I do that, `pylint` complains that it doesn't know what `RockcraftServiceFactory` in the `cast()` call. However, if I do `cast("RockcraftServiceFactory", self._services)` it seems to work??

cmatsuoka: Maybe this solves the cyclic import warning too.

rockcraft/services/lifecycle.py
sergiusens: there's a required rule in pydocstyle for this to work, in setup.cfg style:
```
[pydocstyle]
# D107 Missing docstring in __init__ (reason: documented in class docstring)
# D203 1 blank line required before class docstring (reason: pep257 default)
# D213 Multi-line docstring summary should start at the second line (reason: pep257 default)
ignore = D107, D203, D213
ignore_decorators = overrides
```

rockcraft/services/image.py
```diff
@@ -0,0 +1,101 @@
...
+                variant=None,  # TODO
+            )
+        else:
+            base_image, source_image = oci.Image.from_docker_registry(
+                project.base,
+                image_dir=image_dir,
+                arch=self._build_for,
+                variant=None,  # TODO
+            )
+        emit.progress(f"Retrieved base {project.base} for {build_for}", permanent=True)
```
sergiusens, line 84: maybe a chance to make these not permanent

rockcraft/services/image.py
```diff
@@ -0,0 +1,101 @@
...
+
+        emit.progress(f"Extracting {base_image.image_name}")
+        rootfs = base_image.extract_to(bundle_dir)
+        emit.progress(f"Extracted {base_image.image_name}", permanent=True)
```
sergiusens, line 88: maybe a chance to make these not permanent

rockcraft/services/service_factory.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+@dataclass
+class RockcraftServiceFactory(ServiceFactory):
+    """Rockcraft-specific Service Factory."""
+
+    # pylint: disable=invalid-name
+
+    ImageClass: type[services.RockcraftImageService] = services.RockcraftImageService
+    PackageClass: type[base_services.PackageService] = services.RockcraftPackageService
```
sergiusens, line 40: maybe add a comment that this is an override?

sergiusens (approved):

sergiusens:
Not an expert on the ServiceFactory, but this looks good. Great work

rockcraft/services/lifecycle.py
cmatsuoka: Just checked, it doesn´t.

cmatsuoka (approved):
+1. Not a big fan of suppressing cyclic import warnings but this seems to be bogus.

PR #364, opened 2023-09-29
chore(deps): update dependencies

lengau:

-----


tests/unit/test_pebble.py
lengau, line 1: A pydantic update changed this output, so I updated the tests

tools/freeze-requirements.sh
lengau, line 1: Reordered this so we don't get the documentation requirements in our regular requirements.txt file.

requirements-dev.txt
```diff
@@ -1,131 +1,135 @@
...
 distro==1.8.0
-docutils==0.17.1
-exceptiongroup==1.0.4
-filelock==3.8.2
+docutils==0.18.1
+exceptiongroup==1.1.3
+filelock==3.12.4
 flake8==4.0.1
-furo==2022.12.7
-gnupg==2.3.1
```
tigarmo, line 33: how come this got dropped?

lengau: It was an indirect dependency via craft-archives, but we removed that dependency from craft-archives back in May.

tests/unit/test_pebble.py
cjdcordeiro: interesting :thinking: thanks!

lengau: It's probably not ideal that pydantic patch updates (somewhere between `1.10.0` and `1.10.13` are causing our tests to change, but that's out of scope for this PR and maybe something to ask about upstream.

tigarmo (approved):

PR #368, opened 2023-10-06
chore(snap): base migration from core20 to core22

sergiusens:

-----


snap/snapcraft.yaml
```diff
@@ -51,19 +53,20 @@ parts:
...
+        - "CFLAGS": "$(pkg-config python-3.10 yaml-0.1 --cflags)"
     after: [rockcraft-libs]
     override-build: |
       snapcraftctl build
       # python3 fixup symlink (snapcraft bug)
-      ln -sf ../usr/bin/python3.8 $SNAPCRAFT_PART_INSTALL/bin/python3
-        mkdir -p $SNAPCRAFT_PART_INSTALL/libexec/rockcraft
-        mv $SNAPCRAFT_PART_INSTALL/bin/craftctl $SNAPCRAFT_PART_INSTALL/libexec/rockcraft/
-        sed -i -e '1 s|^#!/.*|#!/snap/rockcraft/current/bin/python|' $SNAPCRAFT_PART_INSTALL/libexec/rockcraft/craftctl
+      mkdir -p $SNAPCRAFT_PART_INSTALL/libexec/rockcraft
```
tigarmo, line 67: can we use `CRAFT_PART_INSTALL` instead?? You know, in a "dogfooding" sense

sergiusens: you are correct!

tigarmo (approved):
Thanks! If the spread tests are happy then I'm happy
Also apparently the snap is smaller? ~50 MB as opposed to ~80 MB for the many rockcraft snaps that I have locally.

sergiusens:
something went wrong :sweat_smile:, will check later :-)

cmatsuoka (approved):
Looks good. Any new info about the size delta?

snap/snapcraft.yaml
```diff
@@ -9,21 +9,26 @@ confinement: classic
 grade: devel
 license: GPL-3.0
 
+# https://github.com/snapcore/snapcraft/issues/4187
+environment:
+  PATH: "$SNAP/libexec/rockcraft:/snap/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
+  LD_LIBRARY_PATH: "$SNAP/none"
```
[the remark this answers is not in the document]
syu-w: You mean the multiple `$SNAP/` PATH?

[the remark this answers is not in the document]
sergiusens: what tools require this? we should use the abspath

docs/how-to/code/install-slice/rockcraft.yaml
```diff
@@ -15,4 +15,4 @@ parts:
         source: chisel-releases/
         source-type: local
         override-build:
-            chisel cut --release ./ --root ${CRAFT_PART_INSTALL} openssl_bins
+            /snap/rockcraft/current/bin/chisel cut --release ./ --root ${CRAFT_PART_INSTALL} openssl_bins
```
sergiusens, line 18: this might mean we want chisel in libexec, @tigarmo, thoughts?

tigarmo: err probably? Is that preferable to adding `$SNAP/bin` to $PATH?

requirements-dev.txt
```diff
@@ -16,7 +16,7 @@ colorama==0.4.6
 coverage==7.3.1
 craft-archives==1.1.3
 craft-cli==2.1.0
-craft-parts==1.25.1
+craft-parts@git+https://github.com/canonical/craft-parts.git@snap-bin-path
```
tigarmo, line 19: do we still need this? I thought overlayfs would come from libexec

snap/snapcraft.yaml
```diff
@@ -105,5 +114,9 @@ parts:
     plugin: go
     source: https://github.com/canonical/chisel.git
     source-commit: bd27f8700cd7d2a6b4e0df6b10c3761c83a70485
+    build-environment:
+      - CGO: "0"
+      - CGO_ENABLED: "0"
+      - GOBIN: "${SNAPCRAFT_PART_INSTALL}/libexec/rockcraft"
```
tigarmo, line 120:
```suggestion
      - GOBIN: "${CRAFT_PART_INSTALL}/libexec/rockcraft"
```

requirements-dev.txt
syu-w: That one if from `apt`, unless we want to move it to `libexec` by hand?

sergiusens: apt for build-packages? that should come from the host

docs/how-to/code/install-slice/rockcraft.yaml
sergiusens: we talked about this, and went with libexec

requirements-dev.txt
sergiusens: from what I just saw in the PR, we are only using that code in craft-parts for the fuse overlayfs command, so this should be safe to drop now that that is in libexec.

Do NOT move apt to libexec.

syu-w: No, I mean the overlayfs is installed by apt in the parts, and that command is in `$SNAP/usr/bin`. We need to move it into libexec in some way.

sergiusens: yes, this is why I said we want to "organize" that component into libexec

tigarmo: can anyone remind me why we chose not to provision `fuse-overlayfs` in the build instance?

tigarmo: update: we talked a bit about this during standup and the reason was "destructive mode"
so the agreed-upon solution is to use `organize` to put fuse-overlayfs in libexec (as Sergio said last)

syu-w: Got it.

PR #374, opened 2023-10-10
Provide a default prime in the flask extension

weiiwang01:
Provide a default prime list in the Flask extension for users' project files unless they supply their own.

If the user provides a prime list, it will replace the default one.

### DIFF
--- a/rockcraft/extensions/flask_framework.py
+++ b/rockcraft/extensions/flask_framework.py
@@ -193,9 +193,9 @@ class FlaskFramework(Extension):
             ),
             install_app_part_name: {
                 "prime": [
-                    "flask/app/app.py",
-                    "flask/app/static",
-                    "flask/app/templates",
+                    f"flask/app/{f}"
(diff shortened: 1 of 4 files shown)


rockcraft/extensions/flask_framework.py
```diff
@@ -198,15 +191,21 @@ def _gen_new_parts(self) -> Dict[str, Any]:
             dependencies_part_name: self._merge_existing_part(
                 dependencies_part_name, dependencies_part
             ),
-            install_app_part_name: self._merge_existing_part(
-                install_app_part_name, install_app_part
-            ),
+            install_app_part_name: {
+                "prime": [
+                    "flask/app/app.py",
```
jdkandersson, line 196: The `app` directory should also be here please

weiiwang01: Added, thanks!

tigarmo (approved):
looks good! Thanks

PR #379, opened 2023-10-18
Add a profile option to the rockcraft init command

weiiwang01:
Include a `--profile` option in the `rockcraft init` command, enabling rockcraft to produce different flavors of initial project files.


rockcraft/commands/init.py
```diff
@@ -60,28 +107,34 @@ class InitCommand(BaseCommand):
...
-
-            parts:
-                my-part:
-                    plugin: nil
-            """
-    )
+    def fill_parser(self, parser):
+        """Specify command's specific parameters."""
+        parser.add_argument(
+            "--name", help="The name of the charm; defaults to the directory name"
```
tigarmo, line 113:
```suggestion
            "--name", help="The name of the ROCK; defaults to the directory name"
```

tigarmo (changes requested):
Thanks! Please update the spread test to use this new command as we discussed

rockcraft/commands/init.py
```diff
@@ -60,28 +129,34 @@ class InitCommand(BaseCommand):
...
 
     @overrides
     def run(self, parsed_args: "argparse.Namespace") -> None:
         """Run the command."""
-        init(self._INIT_TEMPLATE_YAML)
+        if not parsed_args.name:
+            parsed_args.name = pathlib.Path.cwd().name
+            emit.debug(f"Set project name to '{parsed_args.name}'")
+
+        if not re.match(r"[a-z][a-z0-9-]*[a-z0-9]$", parsed_args.name):
```
cjdcordeiro, line 151: why is this regex different from

https://github.com/canonical/rockcraft/blob/318ecdf85872e9d0769d2b62e70fea4acc77505d/rockcraft/project.py#L159

? 
In any case, I don't think it should return an error. If the parameter is optional, then the command should always work whenever that option is not provided. 

I'd instead suggest issuing a warning and populating the name with the current default `my-rock-name # the name of your ROCK`

cjdcordeiro (changes requested):

rockcraft/commands/init.py
weiiwang01: Updated, thanks!

rockcraft/commands/init.py
weiiwang01: Sorry the name regex is charmcraft's, I have updated that. I've also update the name error handling: if a user explicitly provides an invalid name, it will trigger an error. However, if the name is inferred implicitly, it will default to `my-rock-name` without causing an error.

cjdcordeiro: Thanks! You might want to add tests for those scenarios.

weiiwang01: Tests added, thanks!

cjdcordeiro (approved):
LGTM, thanks and thanks for the changes ;)

tigarmo (approved):

Issue #358, opened 2023-09-21
Move "build_for_variant" from the Project to the Image


tigarmo:
### What needs to get done

Drop the "build_for_variant" from the `Project`, and move the existing logic to the `Image` factory methods.

### Why it needs to get done

The "build_for_variant" property of platforms is "hidden" in the sense that the user can't see or set it, and is only used when fetching the base image from the registry.

PR #380, opened 2023-10-18
feat: use debian arch in platforms

### LINKED ISSUE #370: Platforms does not resolve with deb_arch
### Bug Description

The platforms resolution is a mixed bag of uts machine architectures and deb arches, craft applications are supposed to expose and use debian architectures as their UX.

This is required to solve #358

### To Reproduce

Set a platform architecture to armhf

### Environment

N/A

### rockcraft.yaml

```shell
platforms:
   armhf
```

### Relevant log output

```shell
N/A
```

tigarmo:
This commit puts Rockcraft in line with other craft tools: the
architectures listed in the platforms are in Debian format, and we
convert when necessary (such as when fetching images from a Docker
registry).

As a side-effect, we also drop the "build variant" notion from the
platforms - currently this variant is only used when fetching/creating
images, so keep the variant logic localized to rockcraft.oci.

Fixes #370, #358

### DIFF
--- a/rockcraft/oci.py
+++ b/rockcraft/oci.py
@@ -76,7 +76,7 @@ class Image:
 
         :param image_name: The image to retrieve, in ``name:tag`` format.
         :param image_dir: The directory to store local OCI images.
-        :param arch: The architecture of the Docker image to fetch, in debian format.
+        :param arch: The architecture of the Docker image to fetch, in Debian format.
         :param variant: The variant, if any, of the Docker image to fetch.
 
 
@@ -117,7 +117,7 @@ class Image:
 
         :param image_name: The image to initiate, in ``name:tag`` format.
         :param image_dir: The directory to store the local OCI image.
-        :param arch: The architecture of the OCI image to create, in debian format.
+        :param arch: The architecture of the OCI image to create, in Debian format.
         :param variant: The variant, if any, of the OCI image to create.
 
         :returns: The new image object and it's corresponding source image
@@ -129,15 +129,17 @@ class Image:
         _process_run(["umoci", "init", "--layout", image_target_no_tag])
         _process_run(["umoci", "new", "--image", str(image_target)])
 
+        # Note: umoci's docs aren't clear on this but we assume that arch-related
+        # calls must use GOARCH-format, following the OCI spec.
+        mapping = SUPPORTED_ARCHS[arch]
+
         # Unfortunately, umoci does not allow initializing an image
         # with arch and variant. We can configure the arch via
         # umoci config, but not the variant. Need to do it manually
-        _config_image(image_target, ["--architecture", arch, "--no-history"])
+        _config_image(image_target, ["--architecture", mapping.go_arch, "--no-history"])
 
-        variant = SUPPORTED_ARCHS[arch].go_variant
-
-        if variant:
-            _inject_architecture_variant(Path(image_target_no_tag), variant)
+        if mapping.go_variant:
+            _inject_architecture_variant(Path(image_target_no_tag), mapping.go_variant)
 
         # for new OCI images, the source image corresponds to the newly generated image
         return (
--- a/tests/unit/test_oci.py
+++ b/tests/unit/test_oci.py
@@ -30,6 +30,7 @@ from craft_parts.overlays import overlays
 
 import tests
 from rockcraft import errors, oci
+from rockcraft.architectures import SUPPORTED_ARCHS
 
 MOCK_NEW_USER = {
     "user": "foo",
@@ -209,10 +210,14 @@ class TestImage:
         assert arch_data.override_arch == expected_arch
         assert arch_data.override_variant == expected_variant
 
-    def test_new_oci_image(self, mock_inject_variant, mock_run):
+    @pytest.mark.parametrize("deb_arch", list(SUPPORTED_ARCHS))
+    def test_new_oci_image(self, mock_inject_variant, mock_run, deb_arch):
+        """Test that new blank images are created with the correct GOARCH values."""
+        expected = SUPPORTED_ARCHS[deb_arch]
+
         image_dir = Path("images/dir")
         image, source_image = oci.Image.new_oci_image(
-            "bare:latest", image_dir=image_dir, arch="amd64"
+            "bare:latest", image_dir=image_dir, arch=deb_arch
         )
         assert image_dir.is_dir()
         assert image.image_name == "bare:latest"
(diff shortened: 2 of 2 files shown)


rockcraft/oci.py
```diff
@@ -108,13 +121,12 @@ def new_oci_image(
...
         image_dir: Path,
         arch: str,
-        variant: Optional[str] = None,
     ) -> Tuple["Image", str]:
         """Create a new OCI image out of thin air.
 
         :param image_name: The image to initiate, in ``name:tag`` format.
         :param image_dir: The directory to store the local OCI image.
-        :param arch: The architecture of the OCI image to create.
+        :param arch: The architecture of the OCI image to create, in debian format.
```
tigarmo, line 129: I actually don't know what arch we're supposed to pass to the umoci call in `_config_image()` (line 144); the documentation doesn't specify and apparently (from my testing) it accepts anything

rockcraft/oci.py
```diff
@@ -87,9 +97,12 @@ def from_docker_registry(
 
         source_image = f"docker://{REGISTRY_URL}/{image_name}"
         copy_params = ["--retry-times", str(MAX_DOWNLOAD_RETRIES)]
+
+        docker_arch, variant = _DEB_TO_DOCKER[arch]
```
cjdcordeiro, line 101:
```suggestion
        goarch, variant = _DEB_TO_GOARCH[arch]
```

cjdcordeiro (changes requested):
Nice to see the craft_application doing some of this work :) Nice work.

Just a few notes.

sergiusens (approved):

rockcraft/models/project.py
```diff
@@ -55,64 +55,50 @@
...
     """Maps different denominations of the same architecture."""
 
     description: str
-    deb_arch: str
-    compatible_uts_machine_archs: List[str]
+    compatible_deb_archs: List[str]
     go_arch: str
 
 
+# The keys are valid debian architectures.
```
cmatsuoka, line 67: Is this mapping still necessary?

rockcraft/oci.py
```diff
@@ -49,6 +49,17 @@
 # The number of times to try downloading an image from `REGISTRY_URL`.
 MAX_DOWNLOAD_RETRIES = 5
 
+# A mapping from arch in Debian format to a pair of (arch in Docker format, arch variant).
+# The values are taken from the supported architectures in the registry that we
+# currently use (see https://gallery.ecr.aws/ubuntu/ubuntu).
+_DEB_TO_DOCKER = {
```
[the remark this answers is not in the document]
tigarmo: I didn't know that these are just Go archs, makes sense!
So I did a little refactoring and moved that _SUPPORTED_ARCHS dict to a new module, dropping this code entirely. This way both the Project and the Images can source from the same place, which makes more sense.

rockcraft/oci.py
```diff
@@ -49,6 +49,17 @@
 # The number of times to try downloading an image from `REGISTRY_URL`.
 MAX_DOWNLOAD_RETRIES = 5
 
+# A mapping from arch in Debian format to a pair of (arch in Docker format, arch variant).
```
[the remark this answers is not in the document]
tigarmo: dropped this code

rockcraft/oci.py
```diff
@@ -49,6 +49,17 @@
...
+# The values are taken from the supported architectures in the registry that we
+# currently use (see https://gallery.ecr.aws/ubuntu/ubuntu).
```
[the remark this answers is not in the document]
tigarmo: Thanks for the reference! Added to `architectures.py`

rockcraft/oci.py
```diff
@@ -87,9 +97,12 @@ def from_docker_registry(
 
         source_image = f"docker://{REGISTRY_URL}/{image_name}"
         copy_params = ["--retry-times", str(MAX_DOWNLOAD_RETRIES)]
+
+        docker_arch, variant = _DEB_TO_DOCKER[arch]
+
         platform_params = [
             "--override-arch",
-            arch,
+            docker_arch,
```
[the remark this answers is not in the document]
tigarmo: updated this code

rockcraft/oci.py
```diff
@@ -130,6 +142,9 @@ def new_oci_image(
         # with arch and variant. We can configure the arch via
         # umoci config, but not the variant. Need to do it manually
         _config_image(image_target, ["--architecture", arch, "--no-history"])
+
+        variant = {"armhf": "v7", "arm64": "v8"}.get(arch)
```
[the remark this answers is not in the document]
tigarmo: updated this code

rockcraft/oci.py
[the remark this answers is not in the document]
cjdcordeiro: alright sounds good. I'm having a look

rockcraft/architectures.py
```diff
@@ -0,0 +1,64 @@
...
+    ),
+    "i386": ArchitectureMapping(
+        description="Intel 386",
+        go_arch="386",
+    ),
+    "ppc64el": ArchitectureMapping(
+        description="PowerPC 64-bit",
+        go_arch="ppc64le",
+    ),
+    "riscv64": ArchitectureMapping(
```
cjdcordeiro, line 56: so are we keeping it in the end?

rockcraft/oci.py
```diff
@@ -68,15 +69,14 @@ def from_docker_registry(
...
-        variant: Optional[str] = None,
     ) -> Tuple["Image", str]:
         """Obtain an image from a docker registry.
 
         The image is fetched from the registry at ``REGISTRY_URL``.
 
         :param image_name: The image to retrieve, in ``name:tag`` format.
         :param image_dir: The directory to store local OCI images.
-        :param arch: The architecture of the Docker image to fetch.
+        :param arch: The architecture of the Docker image to fetch, in debian format.
```
cjdcordeiro, line 79: tiny nitpick:
```suggestion
        :param arch: The architecture of the Docker image to fetch, in Debian format.
```

rockcraft/oci.py
cjdcordeiro: did you test with an arch who's goarch is different from the debian arch? like armhf or ppc64le? if not, the easiest way to test might be to either using scalingstack or run rockcraft in `--destructive-mode` from inside a container of that architecture (using qemu)

cjdcordeiro (approved):
noice!! very clean.

I left a few non-blocking nitpicks.

Despite the approval, I'd still recommend double checking https://github.com/canonical/rockcraft/pull/380/files#r1364486923 just to be sure Umoci is working well for non-amd64 archs

rockcraft/architectures.py
tigarmo: I think we should, so we're "forwards compatible" whenever riscv64 rockcraft is available.

tigarmo:
@cjdcordeiro regarding the umoci arch: as we discussed, I updated the code to use GOARCH as that is most likely the expected format (based on umoci's code and the OCI spec) and next week I'll confirm that assumption with a manual test on a non-standard arch

PR #393, opened 2023-10-23
feat: update base notation from ":" to "@"

tigarmo:
The currently supported bases still accept ":"; future bases will only support "@".
There are 3 commits: The first is unrelated but does a necessary bump of craft-cli, the second implements the deprecation support and the third is mostly find/replace.
Edit: and then a "bonus" fourth commit where we drop support for 18.04 in ROCKs


docs/reference/rockcraft.yaml.rst
```diff
@@ -78,10 +78,14 @@ special value ``bare`` means that the ROCK will have no base system at all,
 which is typically used with static binaries or
 :ref:`Chisel slices <chisel_explanation>`.
 
+.. note::
```
tigarmo, line 81: Looks like this:
image

rockcraft/oci.py
```diff
@@ -74,14 +74,19 @@ def from_docker_registry(
 
         The image is fetched from the registry at ``REGISTRY_URL``.
 
-        :param image_name: The image to retrieve, in ``name:tag`` format.
+        :param image_name: The image to retrieve, in ``name@tag`` format.
```
tigarmo, line 77: I was a bit conflicted here one whether to accept `name:tag` or `name@tag`, but ended up with the latter for two reasons:
1) client code (the image service) would need to convert the `project.base` to `name:tag` every time, which feels like busywork that the class can do itself, and
2) it keeps things consistent because the `arch` parameter uses Debian architectures. So, the class accepts debian/craft parameters and converts them internally to docker/goarch.

lengau (approved):

cjdcordeiro (approved):
image
gif

:D thanks!

rockcraft/models/project.py
```diff
@@ -219,6 +222,33 @@ def _validate_build_base(cls, build_base: Optional[str], values: Any) -> str:
             build_base = values.get("base")
         return cast(str, build_base)
 
+    @pydantic.validator("base", pre=True)
+    @classmethod
+    def _validate_deprecated_base(cls, base_value: Optional[str]) -> Optional[str]:
+        return cls._check_deprecated_base(base_value, "base")
+
+    @pydantic.validator("build_base", pre=True)
```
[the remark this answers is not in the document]
tigarmo: I tried that but I think it needs Pydantic 2 (we're at 1.10 currently)
https://docs.pydantic.dev/1.10/usage/validators/#pre-and-per-item-validators

PR #396, opened 2023-10-26
build: set version from git

tigarmo:
This commit pulls in the "dynamic" version setting (from git) from Starbase, and updates snapcraft.yaml to use that version (via adopt-info).

A consequence of this is that, in order to be able to run "python3 setup.py --version" without the dependencies installed, the enabling of the overlay feature is moved from ``rockcraft/__init__.py`` to the lifecycle service, which makes more sense any way as it's a feature specific to that service.

-----

### DIFF
--- a/rockcraft/__init__.py
+++ b/rockcraft/__init__.py
@@ -16,9 +16,12 @@
 
 """The craft tool to create ROCKs."""
 
-from craft_parts import Features
+try:
+    from ._version import __version__
+except ImportError:  # pragma: no cover
+    from importlib.metadata import PackageNotFoundError, version
 
-__version__ = "0.0.1.dev1"
-
-
-Features(enable_overlay=True)
+    try:
+        __version__ = version("rockcraft")
+    except PackageNotFoundError:
+        __version__ = "dev"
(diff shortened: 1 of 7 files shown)


rockcraft/__init__.py
```diff
@@ -16,9 +16,12 @@
 
 """The craft tool to create ROCKs."""
 
-from craft_parts import Features
```
tigarmo, line 19: The reason that I moved this from here to the lifecycle service is because we do `python3 setup.py --version` without installed deps to get the version at snap-build-time, so this import fails.
I think it makes more sense to be in the lifecycle service anyway

snap/snapcraft.yaml
```diff
@@ -69,6 +70,9 @@ parts:
         bin/craftctl: libexec/rockcraft/craftctl
     override-build: |
       ${SNAP}/libexec/snapcraft/craftctl default
+
+      version="$(python3 setup.py --version)"
+      ${SNAP}/libexec/snapcraft/craftctl set version="$version"
```
tigarmo, line 75: I tested this by (locally) creating a signed tag, building the snap and verifying that the version is correct

cmatsuoka (approved):

sergiusens (approved):

PR #397, opened 2023-10-30
docs: add a nodejs tutorial

### LINKED ISSUE #369: Add a nodejs tutorial
### Request type

Enhancement

### What needs to get done

Add a tutorial for packing a ROCK that contains a node-based application.

### Documentation location

_No response_

### Additional information

_No response_

tigarmo:
The tutorial builds a simple "hello world" webapp and showcases using a package-repository to install newer node.

Fixes #369

-----

### DIFF
--- a/docs/tutorials/node-app.rst
+++ b/docs/tutorials/node-app.rst
@@ -1,0 +1,143 @@
+Bundle a Node.js app into a ROCK
+********************************
+
+This tutorial describes the steps needed to bundle a typical Node.js application
+into a ROCK.
+
+Prerequisites
+-------------
+- snap enabled system (https://snapcraft.io)
+- LXD installed (https://linuxcontainers.org/lxd/getting-started-cli/)
+- skopeo installed (https://github.com/containers/skopeo)
+- Docker installed (https://snapcraft.io/docker)
+- a text editor
+
+
+Install Rockcraft
+-----------------
+
+Install Rockcraft on your host:
+
+.. literalinclude:: code/node-app/task.yaml
+    :language: bash
+    :start-after: [docs:install-rockcraft]
+    :end-before: [docs:install-rockcraft-end]
+    :dedent: 2
+
+
+Project Setup
+-------------
+
+Starting in an empty folder, create a ``src/`` subdirectory. Inside it, add two
+files:
+
+The first one is the ``package.json`` listing of dependencies, with the
+following contents:
+
+.. literalinclude:: code/node-app/src/package.json
+    :language: json
+
+The second file is our sample app, a simple "hello world" server. Still inside
+``src/``, add the following contents to ``server.js``:
+
+.. literalinclude:: code/node-app/src/server.js
+    :language: javascript
+
+Next, we'll setup the Rockcraft project. In the original empty folder, create
+an empty file called ``rockcraft.yaml``. Then add the following snippets, one
+after the other:
+
+Add the metadata that describes your ROCK, such as its name and license:
+
+.. literalinclude:: code/node-app/rockcraft.yaml
+    :language: yaml
+    :start-after: [docs:rock-metadata]
+    :end-before: [docs:rock-metadata-end]
+
+Add the container entrypoint, as a `Pebble`_ service:
+
+.. literalinclude:: code/node-app/rockcraft.yaml
+    :language: yaml
+    :start-after: [docs:pebble-service]
+    :end-before: [docs:pebble-service-end]
+
+Add a part declaring the dependency on the ``nodejs`` Ubuntu package:
+
+.. literalinclude:: code/node-app/rockcraft.yaml
+    :language: yaml
+    :start-after: [docs:node-part]
+    :end-before: [docs:node-part-end]
+
+Add another part, describing how to build the app created in the ``src/``
+directory:
+
+.. literalinclude:: code/node-app/rockcraft.yaml
+    :language: yaml
+    :start-after: [docs:app-part]
+    :end-before: [docs:app-part-end]
+
+Finally, add the `NodeSource`_ Ubuntu repository so that the ROCK uses a newer
+version of Node.js than is available in the official LTS archives:
+
+.. literalinclude:: code/node-app/rockcraft.yaml
+    :language: yaml
+    :start-after: [docs:nodesource-repo]
+    :end-before: [docs:nodesource-repo-end]
+
+
+Pack the ROCK with Rockcraft
+----------------------------
+
+To build the ROCK, run:
+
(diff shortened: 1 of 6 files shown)


docs/tutorials/node-app.rst
```diff
@@ -0,0 +1,143 @@
+Bundle a Node.js app into a ROCK
```
tigarmo, line 1: Rendered version is here: https://canonical-rockcraft--397.com.readthedocs.build/en/397/tutorials/node-app/

docs/tutorials/node-app.rst
```diff
@@ -0,0 +1,143 @@
...
+********************************
+
+This tutorial describes the steps needed to bundle a typical Node.js application
+into a ROCK.
+
+Prerequisites
+-------------
+- snap enabled system (https://snapcraft.io)
+- LXD installed (https://linuxcontainers.org/lxd/getting-started-cli/)
```
lengau, line 10:
```suggestion
- LXD installed (https://documentation.ubuntu.com/lxd/en/latest/howto/initialize/)
```

IDK if that's the right link, but the docs on linuxcontainers.org are gone now

docs/tutorials/node-app.rst
```diff
@@ -0,0 +1,143 @@
...
+- skopeo installed (https://github.com/containers/skopeo)
```
lengau, line 11: Since the tutorial uses the version of skopeo included in the rockcraft snap is this necessary?

lengau (changes requested):
Ran this in a fresh Jammy VM:

```
ubuntu@racy-chinook:~$ sudo lxd init --auto
ubuntu@racy-chinook:~$ sudo snap install docker
docker 20.10.24 from Canonical✓ installed
ubuntu@racy-chinook:~$ sudo snap install rockcraft --edge --classic
rockcraft (edge) 0+git.36080ac from Sergio Schvezov ⭐ (sergiusens) installed
ubuntu@racy-chinook:~$ mkdir my_rock
ubuntu@racy-chinook:~$ cd my_rock/
ubuntu@racy-chinook:~/my_rock$ mkdir src
ubuntu@racy-chinook:~/my_rock$ nano src/package.json
ubuntu@racy-chinook:~/my_rock$ nano src/server.js
ubuntu@racy-chinook:~/my_rock$ nano rockcraft.yaml
ubuntu@racy-chinook:~/my_rock$ nano rockcraft.yaml
ubuntu@racy-chinook:~/my_rock$ rockcraft pack
```

At this point it failed because of [Docker being inconsiderate on the network](https://github.com/docker/for-linux/issues/103). 

Using the standard workaround:

```
sudo iptables -F FORWARD
sudo iptables -P FORWARD ACCEPT
```

I got further, but I hadn't prepped sufficient disk space, so I'll have to come back to it.

It's probably out of scope for this tutorial, but should we add some documentation for troubleshooting this particular issue? It seems like it'd be pretty common for Rockcraft in particular.

docs/tutorials/node-app.rst
```diff
@@ -0,0 +1,143 @@
...
+    :start-after: [docs:skopeo-copy]
+    :end-before: [docs:skopeo-copy-end]
+    :dedent: 2
+
+Since the ROCK bundles a web-app, we'll first start serving that app on local
+port 8000:
+
+..  code-block:: bash
+
+  docker run -p8000:8080 --rm -it my-node-app:latest
```
lengau, line 118:
```suggestion
  docker run -p8000:8080 --rm -it my-node-app:1.0
```

lengau:
With the suggested changes above, it works. (Including not separately installing skopeo.)

docs/tutorials/node-app.rst
tigarmo: Nope! I copied this blindly from another tutorial
I fixed this here but I'll create an issue to put this common preamble somewhere that all tutorials can reference

docs/tutorials/node-app.rst
tigarmo: Thanks a lot for catching this! When I put that "raw command" there, as opposed to using a snippet from the spread test, I was sure I was going to mess something up and well

... the reason that I did this originally is because the spread test runs the container detached, to access the served webapp, but for the tutorial it makes sense to run "interactively" so that the pebble output shows up. I thought about this for a bit and did a little cheeky hack in the task.yaml to have it both ways, please take a look

tigarmo:
@lengau while working on the review fixes I remembered that craft-parts has an npm plugin (doh), so in the interest of dogfooding it I updated the actual project to use it (it's now way simpler).

docs/tutorials/node-app.rst
```diff
@@ -0,0 +1,144 @@
...
+
+To build the ROCK, run:
+
+.. literalinclude:: code/node-app/task.yaml
+    :language: bash
+    :start-after: [docs:build-rock]
+    :end-before: [docs:build-rock-end]
+    :dedent: 2
+
+At the end of the process, a file named ``my-node-app_0.1_amd64.rock`` should be
```
cjdcordeiro, line 89: Could you actually test the OCI archive name? Since we have a history of changing the naming conventions for the output artefact, it would be nice to test this string in order to be future-proof and preemptively detect any coming changes to the output filename.

cjdcordeiro (changes requested):
Super neat! nice. thanks.

I only have one request, but it should be ok after that :)

docs/tutorials/node-app.rst
tigarmo: It's a good idea, I definitely agree with the sentiment. I updated the docs to do a `ls my-node-app_1.0_amd64.rock`, which looks slightly awkward in the tutorial but will break if we change this name.

mr-cal (approved):
Went through the tutorial, worked great for me!

cjdcordeiro (approved):
thanks for the changes

SamirPS:
hello,

i only one feeling about the tutorial, so the application is pretty simple ( good point) but i think we can explain a little more the section with add on the rockcraft.yaml.

One example is the service section, we can say " this part is defined to do this.... and if you application crash....".

WDYT?

tigarmo:
@SamirPS it's tough because a guideline of tutorials in Diátaxis is to keep explanations to a minimum, which is very hard to do sometimes. I already thought I put too much in but a tech author reviewed it and seems to be fine.

SamirPS:
I checked with other people, and they said it's good. If possible, could you please add the full rockcraft.yaml file at the end to help people to copy past the yaml file.

tigarmo:
Sure. I had to update the notation to extract the snippets so that the whole file is "clean", but I put the whole rockcraft.yaml file at the end of the "Project Setup" section

PR #398, opened 2023-10-30
snap: add skopeo as a snap application

lengau:
Allows running the rockcraft-packaged copy of skopeo as `rockcraft.skopeo`

-----


snap/snapcraft.yaml
```diff
@@ -18,6 +18,8 @@ environment:
 apps:
   rockcraft:
     command: bin/python $SNAP/bin/rockcraft
+  skopeo:
```
tigarmo, line 21: I think this is a very good idea! My only possible concern is whether we then need to maintain this as part of the "public api" forever and ever
@sergiusens what do you think?

tigarmo: ping @sergiusens

PR #402, opened 2023-11-03
fix: dynamic grade to allow to release to stable

sergiusens:

-----

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,5 +1,4 @@
 name: rockcraft
-version: git
 base: core22
 summary: A craft like experience to create ROCKS
 description: |
@@ -7,7 +6,6 @@ description: |
     to create OCI images.
 adopt-info: rockcraft
 confinement: classic
-grade: devel
 license: GPL-3.0
 
 # https://github.com/snapcore/snapcraft/issues/4187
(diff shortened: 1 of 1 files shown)


cjdcordeiro (approved):

mr-cal (approved):

snap/snapcraft.yaml
```diff
@@ -73,11 +71,12 @@ parts:
 
       version="$(python3 setup.py --version)"
       ${SNAP}/libexec/snapcraft/craftctl set version="$version"
-  
+      [ -n "$(echo $version | grep "+git")" ] && grade=devel || grade=stable
```
syu-w, line 74:
```suggestion
      [ -n "$(echo $version | grep "+git\|dev")" ] && grade=devel || grade=stable
```

tigarmo, line 74: someone needs to double-check this but I don't think the version has `+git`, but `+g` (from the setuptools versioning scheme used in starbase)

tigarmo: eg one of the dev files I have here is `rockcraft_9.9.9.post1+g5f30fc3_amd64.snap`

syu-w: The current main packed as `rockcraft_0.0.post330+g2ae4796_amd64.snap`.
But I also had some old snaps like `rockcraft_0+git.2c79cb7_amd64.snap` and `rockcraft_0+git.f51824d-dirty_amd64.snap`

sergiusens: yeah,  I changed it to "post" and tried out locally, GH was down last night, so could not push anymore :sweat_smile:

PR #408, opened 2023-11-08
chore(docs): fix version in readthedocs build

tigarmo:

-----

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -35,3 +35,5 @@ formats:
 python:
   install:
     - requirements: requirements-doc.txt
+    - method: pip
+      path: .


.readthedocs.yaml
```diff
@@ -35,3 +35,5 @@ formats:
 python:
   install:
     - requirements: requirements-doc.txt
+    - method: pip
```
cmatsuoka, line 38: What is the default method if pip is not specified?

cmatsuoka (approved):

.readthedocs.yaml
tigarmo: oh it looks like the default is `pip`, so this is superfluous

mr-cal (approved):

PR #410, opened 2023-11-17
Add entrypoint_service for pebble entrypoint services

linostar:

-----


linostar:
Tagging @deusebio for review.

rockcraft/models/project.py
```diff
@@ -326,6 +327,19 @@ def _validate_base_and_overlay(cls, item: Dict[str, Any], values) -> Dict[str, A
...
+    @classmethod
+    def _validate_entrypoint_service(
+        cls, entrypoint_service: Optional[str], values: Any
+    ) -> Optional[str]:
+        """Verify that the entrypoint_service exist in the services dict."""
+        services = values.get("services")
+        if entrypoint_service and not (
+            services and entrypoint_service in services.keys()
+        ):
+            raise ProjectValidationError("entrypoint_service used is invalid!")
```
cjdcordeiro, line 340: The user will see this message so it's better to be a bit more verbose (e.g. "The provided entrypoint-service is not a valid Pebble service"). Also, it should be `entrypoint-service` instead of `entrypoint_service` cause that's the internal variable name, and the users are not exposed to that

tests/spread/general/entrypoint-service/rockcraft.yaml
```diff
@@ -0,0 +1,22 @@
...
+  success:
+    override: replace
+    command: "/usr/bin/echo Success"
+    startup: enabled
+
+platforms:
+  amd64:
+  amd64v2:
+    build-on: ["amd64", "i386"]
+    build-for: amd64
```
cjdcordeiro, line 18:
```suggestion
```

cjdcordeiro: (no need for it in this test :) )

tests/spread/general/entrypoint-service/rockcraft.yaml
```diff
@@ -0,0 +1,22 @@
...
+base: ubuntu@22.04
+summary: A ROCK with an entrypoint-service field
+description: A ROCK with an entrypoint-service field
+license: Apache-2.0
+entrypoint-service: success
+services:
+  success:
+    override: replace
+    command: "/usr/bin/echo Success"
+    startup: enabled
```
cjdcordeiro, line 12: Just for the sake of being consistent with Pebble's MO:
 - the service name should allude to an actual service/process/daemon (e.g. "nginx", "controller", etc.). In this case, something like `test-service` fits the purpose
 - the `command` should be more configurable so that we can actually test the image in `task.yaml`. so I would suggest something like `command: echo [ foo ]`

tests/spread/general/entrypoint-service/task.yaml
```diff
@@ -0,0 +1,15 @@
+summary: container entrypoint-service test
+
+execute: |
+  run_rockcraft pack
+
+  test -f entrypoint-service-test_latest_amd64.rock
+  test -f entrypoint-service-test_latest_amd64v2.rock
```
cjdcordeiro, line 7:
```suggestion
```

tests/spread/general/entrypoint-service/task.yaml
```diff
@@ -0,0 +1,15 @@
...
+
+  # test container execution
+  sudo /snap/rockcraft/current/bin/skopeo --insecure-policy copy oci-archive:entrypoint-service-test_latest_amd64v2.rock docker-daemon:entrypoint-service-test:latest
```
cjdcordeiro, line 10:
```suggestion
  sudo /snap/rockcraft/current/bin/skopeo --insecure-policy copy oci-archive:entrypoint-service-test_latest_amd64.rock docker-daemon:entrypoint-service-test:latest
```

tests/spread/general/entrypoint-service/task.yaml
```diff
@@ -0,0 +1,15 @@
...
+  rm entrypoint-service-test_latest*.rock  
+  docker images entrypoint-service-test:latest
+  id=$(docker run --rm -d entrypoint-service-test -v)
```
cjdcordeiro, line 13: why the `-v`? 

What I think we should be doing here is:
`docker run -d entrypoint-service-test` and test that the logs say "foo"

then
`docker run -d entrypoint-service-test bar` and test that the logs say "bar"

p.s. you don't need `--rm` cause the container isn't shutting down on its own and you're deleting later on, below

tests/unit/test_project.py
```diff
@@ -251,6 +251,21 @@ def test_project_title_empty_invalid_name(yaml_loaded_data):
...
+
+
+@pytest.mark.parametrize("entrypoint_service", ["bye"])
+def test_project_entrypoint_service_invalid(yaml_loaded_data, entrypoint_service):
+    yaml_loaded_data["entrypoint_service"] = entrypoint_service
+    with pytest.raises(ProjectValidationError) as err:
+        Project.unmarshal(yaml_loaded_data)
+    assert "entrypoint_service used is invalid!" in str(err.value)
+
+
```
cjdcordeiro, line 268: could you also please test:
 - that the warning message is emitted
 - that the project loads well without this attribute. I say this because `entrypoint-service` should actually be part of ROCKCRAFT_YAML. this fixture should contain all the allowed fields, for the unmarshalling tests
 - when the value is empty

cjdcordeiro (changes requested):
Thanks for this :)

Could you please bring the PR back to WIP? we should wait for @deusebio to finish some other changes on his side before we are 100% sure we need this feature.

Left a few comments ;)

docs/reference/rockcraft.yaml.rst
```diff
@@ -159,6 +159,22 @@ when they fail. It uses Pebble's layer specification syntax, with each
...
+
+**Required**: No
+
+Makes Rockcraft extend the default OCI entrypoint
+``["/bin/pebble", "enter", "--verbose"]`` with ``["--args", "<value>"]``.
+This is used to set one of the defined Pebble services as a part of the OCI
+entrypoint.
+This option must only be used in cases where the targeted deployment
+environment has unalterable assumptions about the container image's
+entrypoint.
```
cjdcordeiro, line 175: please make it come right after `services`. It feels more fluid that way.

cjdcordeiro, line 175: this should be a Warning block. See an example in https://github.com/canonical/rockcraft/pull/410/files#diff-d88372eb092fe5a91cf204e108451fe4e9f34927f627138837af6b9083c1fb3bL175

docs/reference/rockcraft.yaml.rst
```diff
@@ -159,6 +159,22 @@ when they fail. It uses Pebble's layer specification syntax, with each
...
 
+``entrypoint-service``
+------------------------
+
+**Type**: string
+
+**Required**: No
+
+Makes Rockcraft extend the default OCI entrypoint
+``["/bin/pebble", "enter", "--verbose"]`` with ``["--args", "<value>"]``.
```
cjdcordeiro, line 170:
```suggestion
``["/bin/pebble", "enter", "--verbose"]`` with ``["--args", "<serviceName>"]``.
```

rockcraft/models/project.py
```diff
@@ -326,6 +327,24 @@ def _validate_base_and_overlay(cls, item: Dict[str, Any], values) -> Dict[str, A
             )
         return item
 
+    @pydantic.validator("entrypoint_service", always=True)
```
cjdcordeiro, line 330: you shouldn't need the `always=True`, cause you don't want to do anything unless the field is actually set, which means you can also drop the `if entrypoint_service` below, cause the validator should only run when the field is set

rockcraft/models/project.py
```diff
@@ -326,6 +327,24 @@ def _validate_base_and_overlay(cls, item: Dict[str, Any], values) -> Dict[str, A
...
+    @classmethod
+    def _validate_entrypoint_service(
+        cls, entrypoint_service: Optional[str], values: Any
+    ) -> Optional[str]:
+        """Verify that the entrypoint_service exist in the services dict."""
+        if entrypoint_service:
+            craft_cli.emit.message(
+                """Warning: defining an entrypoint-service will result in a rock with an atypical OCI Entrypoint. While that might be acceptable for testing and personal use, it shall require prior approval before submitting to a Canonical registry namespace."""
```
cjdcordeiro, line 338: how come the linters aren't complaining about the size of this line? :open_mouth:

rockcraft/models/project.py
```diff
@@ -568,6 +587,7 @@ def load_project(filename: Path) -> Dict[str, Any]:
         raise ProjectLoadError(msg) from err
 
     yaml_data = transform_yaml(filename.parent, yaml_data)
+
```
cjdcordeiro, line 590:
```suggestion
```

cjdcordeiro: added by mistake I believe...

rockcraft/services/package.py
```diff
@@ -163,7 +163,13 @@ def _pack(
...
     emit.progress("Adding Pebble entrypoint")
 
-    new_image.set_entrypoint()
+    if project.services and project.entrypoint_service in project.services:
+        new_image.set_entrypoint(
+            project.entrypoint_service,
+            project.services[project.entrypoint_service].command,
+        )
+    else:
+        new_image.set_entrypoint(None, None)
```
cjdcordeiro, line 172: I think this should be split. It's better to have pure functions that only do one job. It also makes it easier for unit testing. So, I propose:

 - 2 functions: `new_image.set_entrypoint(entrypoint_service)` and `new_image.set_cmd(entrypoint_service_command)` 
 - then this code block becomes:
```python
# you let the function handle the `None` value
new_image.set_entrypoint(project.entrypoint_service)

if project.entrypoint_service:
    # you let this new function handle the splitting of the command
    new_image.set_cmd(project.services[project.entrypoint_service].command)
```

rockcraft/oci.py
```diff
@@ -355,17 +355,28 @@ def set_default_user(self, user: str) -> None:
...
         emit.progress("Configuring entrypoint...")
+        cmd_params = ["--clear=config.cmd"]
         image_path = self.path / self.image_name
         entrypoint = [f"/{Pebble.PEBBLE_BINARY_PATH}", "enter", "--verbose"]
+        if entrypoint_service:
+            if args:
+                parts = args.split()
+            else:
+                parts = []
+            entrypoint.extend(["--args", entrypoint_service])
```
cjdcordeiro, line 371: this is the only line that matters in this function. the rest should be moved to a new function. See my explanation below in https://github.com/canonical/rockcraft/pull/410/files#diff-cf13f6f606f30afa23cf9954776527355fc805a6d81f60f29ec6899c261d83e9R172

tests/spread/general/entrypoint-service/task.yaml
```diff
@@ -0,0 +1,17 @@
...
+  # test container execution
+  sudo /snap/rockcraft/current/bin/skopeo --insecure-policy copy oci-archive:entrypoint-service-test_latest_amd64.rock docker-daemon:entrypoint-service-test:latest
+  rm entrypoint-service-test_latest*.rock  
+  docker images entrypoint-service-test:latest
+  id=$(docker run -d entrypoint-service-test)
+  grep_docker_log "$id" "foo"
+  docker rm -f "$id"
+  id=$(docker run -d entrypoint-service-test bar)
+  grep_docker_log "$id" "bar"
+  docker rm -f "$id"
```
cjdcordeiro, line 17: you're missing the `docker inspect` test we discussed offline, i.e.:

`docker inspect entrypoint-service-test:latest -f '{{json .Config.Entrypoint}}' == '["/bin/pebble","enter","--verbose", "--args", "test-service"]'`

and

`docker inspect entrypoint-service-test:latest -f '{{json .Config.Cmd}}' == '["foo"]'`

tests/unit/test_oci.py
```diff
@@ -897,7 +897,7 @@ def test_set_default_user(self, mock_run):
     def test_set_entrypoint(self, mock_run):
         image = oci.Image("a:b", Path("/c"))
 
-        image.set_entrypoint()
+        image.set_entrypoint(None, None)
```
cjdcordeiro, line 901: you also need the test for when the entrypoint-service is not None

tests/unit/test_project.py
```diff
@@ -251,6 +253,41 @@ def test_project_title_empty_invalid_name(yaml_loaded_data):
     assert "Invalid name for ROCK" in str(err.value)
 
 
+@pytest.mark.parametrize("entrypoint_service", [""])
+def test_project_entrypoint_service_empty(yaml_loaded_data, entrypoint_service):
+    yaml_loaded_data["entrypoint-service"] = entrypoint_service
+    project = Project.unmarshal(yaml_loaded_data)
+    assert project.entrypoint_service == entrypoint_service
```
cjdcordeiro, line 260: this is weird. this shouldn't pass. "" is not a valid service.

cjdcordeiro (changes requested):
I think it's almost there :)

tests/unit/test_project.py
linostar: An empty entrypoint_service will be considered the same as None.
Do you want to change that behavior?

rockcraft/models/project.py
linostar: Maybe because it's a triple-quote string? Found a bypass 😎

tests/unit/test_project.py
cjdcordeiro: I see.

Well, I guess both approaches could be right.

My gut feeling is to be conservative and not create an interface we would regret in the future. So maybe it is safer to not allow empty values for `entrypoint-service`.

docs/reference/rockcraft.yaml.rst
```diff
@@ -147,6 +147,23 @@ specification syntax exactly, with each entry defining a Pebble service. For
...
 others are optional.
 
+``entrypoint-service``
+------------------------
+
+**Type**: string
+
+**Required**: No
+
+Makes Rockcraft extend the default OCI entrypoint
```
tigarmo, line 157: I think this description is missing a link between the field and the pebble service. Something like `The optional name of the Pebble service to serve as the OCI entrypoint. If set, this makes Rockcraft extend...`

rockcraft/models/project.py
```diff
@@ -326,6 +327,26 @@ def _validate_base_and_overlay(cls, item: Dict[str, Any], values) -> Dict[str, A
             )
         return item
 
+    @pydantic.validator("entrypoint_service")
+    @classmethod
+    def _validate_entrypoint_service(
+        cls, entrypoint_service: Optional[str], values: Any
+    ) -> Optional[str]:
+        """Verify that the entrypoint_service exist in the services dict."""
```
tigarmo, line 335:
```suggestion
        """Verify that the entrypoint_service exists in the services dict."""
```

tests/unit/test_project.py
tigarmo: https://docs.pydantic.dev/latest/api/types/#pydantic.types.constr

tests/unit/test_project.py
```diff
@@ -251,6 +253,44 @@ def test_project_title_empty_invalid_name(yaml_loaded_data):
...
+        Project.unmarshal(yaml_loaded_data)
+    assert "The provided entrypoint-service used is not a valid Pebble service." in str(
+        err.value
+    )
+
+
+def test_project_entrypoint_service_absent(yaml_loaded_data):
+    yaml_loaded_data.pop("entrypoint-service")
+    project = Project.unmarshal(yaml_loaded_data)
+    assert "entrypoint-service" not in project.__dict__
```
tigarmo, line 291:
```suggestion
    assert project.entrypoint_service is None
```

tigarmo (changes requested):
Thanks! Left some comments, but I generally agree with @cjdcordeiro about not accepting empty strings in entrypoint-service and about cleaning up the set_entrypoint() method a bit

rockcraft/oci.py
```diff
@@ -355,19 +355,35 @@ def set_default_user(self, user: str) -> None:
...
+        emit.progress("Configuring CMD...")
+        image_path = self.path / self.image_name
+        cmd_params = ["--clear=config.cmd"]
+        if command:
+            parts = command.split()
+        else:
+            parts = []
+        for i in range(1, len(parts)):
+            if parts[i] != "[" and parts[i] != "]":
+                cmd_params.extend(["--config.cmd", parts[i]])
```
cjdcordeiro, line 382: only noticing this now, which means **we need another test**.

I'm 90% this won't work for most cases. basically, you're assuming that a service command will always be something like `command: app [ args ]`, but in reality, it can be `command: app -v --fixed-option --arg foo [ --some extra-arg --etc ]`.

This means:
 1. we need unit tests for assessing more variations of the service command
 2. this block of code needs to be more generic.

For 2. I suggest doing something like:

```python
# up top
import shlex
#...
def set_cmd(self, command: Optional[str]) -> None:
    """Set the OCI image CMD."""
    emit.progress("Configuring CMD...")
    image_path = self.path / self.image_name
    cmd_params = ["--clear=config.cmd"]
    command_sh_args = shlex.split(command)
    try:
        opt_args = command_sh_args[command_sh_args.index("[")+1:command_sh_args.index("]")]
    except ValueError:
        emit.debug(f"The entrypoint-service command '{command}' has no default arguments. CMD won't be set."
        return
    cmd_params.extend(opt_args)
    _config_image(image_path, cmd_params)
```
(haven't tested)

P.S.: why is `shlex` needed? Well, you must consider the following (which btw, is yet another unit test we need for this): `command: app --arg "[ foo ]" [ --optional ]`. IOW, if passed as a string, the squared brackets have no effect in pebble, and are simply treated as any other argument for the command app.

rockcraft/services/package.py
```diff
@@ -163,7 +163,11 @@ def _pack(
 
     emit.progress("Adding Pebble entrypoint")
 
-    new_image.set_entrypoint()
+    new_image.set_entrypoint(project.entrypoint_service)
+    if project.services and project.entrypoint_service in project.services:
```
cjdcordeiro, line 167:
```suggestion
    if project.services:
```

The 2nd condition is redundant because if that wasn't the case, an exception would've already been raised in the project validation at https://github.com/canonical/rockcraft/pull/410/files#diff-54f38854c7e2d7d0fb5ab8160a493b4cf8f6f7da496a149eeff2680ec7ae6dbbR343

rockcraft/services/package.py
```diff
@@ -163,7 +163,11 @@ def _pack(
...
+        new_image.set_cmd(project.services[project.entrypoint_service].command)
+    else:
+        new_image.set_cmd(None)
```
cjdcordeiro, line 170:
```suggestion
```

This is not needed. If it doesn't exist, don't set it

cjdcordeiro (changes requested):
Almost there :) just found a bug, so we'll need more tests (see my comments below).

Also, please keep a linear history for the commits, because it becomes hard to know what was already reviewed whenever there is a force-push.

rockcraft/services/package.py
linostar: Actually, I had to revert this requested change because the linters start complaining because of the following line if the change is made:
https://github.com/canonical/rockcraft/actions/runs/6983627927&source=gmail&ust=1700936191049000&usg=AOvVaw06UHcYXWgZrPqgXarPnzoQ

rockcraft/oci.py
```diff
@@ -365,24 +366,33 @@ def set_entrypoint(self, entrypoint_service: Optional[str]) -> None:
...
         _config_image(image_path, params)
         emit.progress(f"Entrypoint set to {entrypoint}")
 
     def set_cmd(self, command: Optional[str]) -> None:
         """Set the OCI image CMD."""
         emit.progress("Configuring CMD...")
         image_path = self.path / self.image_name
-        cmd_params = ["--clear=config.cmd"]
+        cmd_params = []
         if command:
```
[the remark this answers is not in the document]
linostar: `command` can be None, and splitting None in shlex is deprecated.

rockcraft/oci.py
```diff
@@ -365,24 +366,33 @@ def set_entrypoint(self, entrypoint_service: Optional[str]) -> None:
...
+        params.extend(["--clear=config.cmd"])
         _config_image(image_path, params)
         emit.progress(f"Entrypoint set to {entrypoint}")
 
     def set_cmd(self, command: Optional[str]) -> None:
         """Set the OCI image CMD."""
         emit.progress("Configuring CMD...")
         image_path = self.path / self.image_name
-        cmd_params = ["--clear=config.cmd"]
+        cmd_params = []
```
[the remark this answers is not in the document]
linostar: it was moved to set_entrypoint. otherwise if we keep it here we will have to call set_cmd in all cases. we've disccused it before iirc and you opted to have clear=config.cmd back in set_entrypoint.

rockcraft/models/project.py
```diff
@@ -342,9 +343,23 @@ def _validate_entrypoint_service(
...
+                + "a valid Pebble service."
             )
 
+        command = values.get("services")[entrypoint_service].command
+        command_sh_args = shlex.split(command)
+        # optional arg is surrounded by brackets, so check that they exist in the
+        # right order
+        try:
+            if command_sh_args.index("[") >= command_sh_args.index("]"):
+                raise ValueError
```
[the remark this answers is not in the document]
linostar: ProjectValidationError will apparently need an argument, so the error string will be duplicated here too.
Check: https://github.com/canonical/rockcraft/actions/runs/7004567791/job/19052578345

rockcraft/oci.py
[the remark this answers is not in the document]
cjdcordeiro: yes it should be in `set_entrypoint`. But what I meant is that it should also be in this fn. I get that right now, `cmd` has already been cleared when we get to this fn, and thus this might feel redundant. It's just that this is enforced by design. So I'm afraid that if tomorrow we change the ROCK's design such that one can set the CMD without setting the entrypoint, then this would be wrong.

rockcraft/models/project.py
```diff
@@ -342,9 +343,23 @@ def _validate_entrypoint_service(
...
+        except ValueError as ex:
+            raise ProjectValidationError(
+                f"The command '{entrypoint_service}' {command} does not contain optional args to "
+                + "override docker run args."
+            ) from ex
```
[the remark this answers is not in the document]
cjdcordeiro: hum interesting...ok leave it then :shrug:

rockcraft/models/project.py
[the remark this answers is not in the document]
cjdcordeiro: Noted. We need a different strategy then, cause `ValueError` is not the right exception in this case. Let's break it down:
 - if either `[` or `]` don't exit, this code won't get to `raise ValueError`, and instead raise that exception by itself;
 - if it gets to line 356 (`raise ValueError`), then it's because the indexed of either `[` or `]` are wrong
 - in both cases, we want to raise a ProjectValidationError

I'd say: either raise `IndexError` or `ProjectValidationError` with an explanatory message.

tigarmo (approved):

cjdcordeiro (approved):
noice! It LGTM :) thanks for your patience.

cjdcordeiro:
@deusebio any updates on your side?

tigarmo:
@cjdcordeiro let me know if you want me to merge this

cjdcordeiro:
will do @tigarmo . we need to wait for @deusebio

deusebio:
@cjdcordeiro product is double checking that the breaking changes we discussed in Riga are fine. Anyhow, we have a PR open [here](https://github.com/canonical/charmed-spark-rock/pull/57) that I hope (if there are no concern with the changes) will land shortly

deusebio (approved):
LGTM, the modifications are rather clear to me, although I don't know the code base enough to provide a more in-depth review. 

I just left a small point/suggestion for the repo. Totally up yo you whether it is worth pursuing.

rockcraft/oci.py
```diff
@@ -355,19 +356,41 @@ def set_default_user(self, user: str) -> None:
         _config_image(image_path, params)
         emit.progress(f"Default user set to {user}")
 
-    def set_entrypoint(self) -> None:
-        """Set the OCI image entrypoint. It is always Pebble and CMD is null."""
+    def set_entrypoint(self, entrypoint_service: Optional[str]) -> None:
```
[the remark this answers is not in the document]
linostar: I provided default to None for both set_entrypoint and set_cmd now.
As for `str | None`, switching to it would require carrying over the change in all similar places in the project, to keep the consistency. Therefore I am opting for keeping it as it is for the time being.

cjdcordeiro:
> LGTM, the modifications are rather clear to me, although I don't know the code base enough to provide a more in-depth review.
> 
> I just left a small point/suggestion for the repo. Totally up yo you whether it is worth pursuing.

thanks @deusebio 

How are things looking on your side? Do you have an updated version of your ROCK with the new Pebble services?

rockcraft/oci.py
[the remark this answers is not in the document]
linostar: @cjdcordeiro it's not the entrypoint that is None here, but the entrypoint_service. Or am I missing something?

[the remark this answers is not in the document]
deusebio: I didn't want to create a swirl here. I just meant that I kinda liked the syntax:

```
.set_entrypoint()
```

vs

```
.set_entrypoint(None)
```

by using default arguments. If this is not compliant to some internal rules you have (e.g. never provide default arguments), feel free to skip this. It was just a suggestion. 

The refactoring "Optional[str]" => " str | None", I agree it is out-of-scope as it touches a lot of places. It was also just a suggestion for your future development, provided by an outsider.

Both points do not block approval (which I have already given) nor require a re-review from my part!

deusebio:
@cjdcordeiro we have the PR I linked above. As part of the CI, we also create an artifact [here](https://github.com/canonical/charmed-spark-rock/actions/runs/7023408765?pr=57). 

I'm just waiting Rob to approve the PR once we have green light about removing jupyter.

rockcraft/oci.py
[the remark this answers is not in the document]
cjdcordeiro: :man_facepalming: you're right...read too fast. It's fine for the `entrypoint-service` arg to default to None. Nevermind me :p

cjdcordeiro:
@deusebio Cool. Approved it. Thanks

@tigarmo once https://github.com/canonical/charmed-spark-rock/pull/57 is in, this PR can also go in :)

deusebio:
@cjdcordeiro The other PR is now merged! :partying_face:

PR #414, opened 2023-11-22
[WIP] migrate to classmethod get_build_plan()

tigarmo:

-----

### DIFF
--- a/docs/how-to/code/install-slice/rockcraft.yaml
+++ b/docs/how-to/code/install-slice/rockcraft.yaml
@@ -1,6 +1,6 @@
 name: custom-openssl-rock
 base: bare
-build_base: "ubuntu@22.04"
+build-base: "ubuntu@22.04"
 version: '0.0.1'
 summary: A chiselled ROCK with a custom OpenSSL slice
 description: |
--- a/rockcraft/models/project.py
+++ b/rockcraft/models/project.py
@@ -431,12 +431,17 @@ class Project(YamlModelMixin, BaseProject):
 
         return (annotations, metadata)
 
-    def get_build_plan(self) -> List[BuildInfo]:
+    @classmethod
+    def get_build_plan(cls, data: Dict[str, Any]) -> List[BuildInfo]:
         """Obtain the list of architectures and bases from the project file."""
         build_infos: List[BuildInfo] = []
-        base = self.effective_base
 
-        for platform_entry, platform in self.platforms.items():
+        platforms = cls._validate_all_platforms(data["platforms"])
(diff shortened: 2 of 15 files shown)


rockcraft/models/project.py
```diff
@@ -431,12 +431,17 @@ def generate_metadata(
...
+    def get_build_plan(cls, data: Dict[str, Any]) -> List[BuildInfo]:
         """Obtain the list of architectures and bases from the project file."""
         build_infos: List[BuildInfo] = []
-        base = self.effective_base
 
-        for platform_entry, platform in self.platforms.items():
+        platforms = cls._validate_all_platforms(data["platforms"])
+        effective_base = data["build-base"] if "build-base" in data else data["base"]
+        name, channel = effective_base.replace(":", "@").split("@")
+        base = bases.BaseName(name, channel)
```
cmatsuoka, line 442: Maybe craft-application could provide a small model containing just platforms, base and build-base with this logic built-in?

docs/how-to/code/install-slice/rockcraft.yaml
```diff
@@ -1,6 +1,6 @@
 name: custom-openssl-rock
 base: bare
-build_base: "ubuntu@22.04"
```
tigarmo, line 3: a side-effect of this change is that I found out that we apparently accept `build_base` as well as `build-base`
weird

rockcraft/services/image.py
```diff
@@ -48,8 +48,9 @@ def __init__(
         project: models.Project,
         work_dir: Path,
         build_for: str,
+        **kwargs,
     ):
-        super().__init__(app, services, project=project)
+        super().__init__(app, services, project=project, **kwargs)
```
tigarmo, line 53: I thought it was very "bureaucratic" to have to add the new parameter to all services, even those that don't use it (like this ImageService)

tests/unit/test_project.py
```diff
@@ -625,5 +625,5 @@ def test_project_yaml(yaml_loaded_data):
 )
 def test_project_get_build_plan(yaml_loaded_data, platforms, expected_build_infos):
     yaml_loaded_data["platforms"] = platforms
-    project = Project.unmarshal(yaml_loaded_data)
-    assert project.get_build_plan() == expected_build_infos
+    build_plan = Project.get_build_plan(yaml_loaded_data)
```
tigarmo, line 628: a loss of usability here is that I can no longer call `project.get_build_plan()` without parameters even if I already have a project (with its base and platforms already validated and set)

PR #416, opened 2023-11-24
refactor: extract file processing code from oci.py

### LINKED ISSUE #296: refactor test_oci.py
### What needs to get done

pylint is now complaining about `test_oci.py` being too big. We're ignoring this for now (https://github.com/canonical/rockcraft/pull/233/files#diff-43d03eb32b3769225898764d036e6e7fad2649ce712b0d45135d084631789acfR16) but it should be fixed by splitting this test (or its source counterpart) into multiple files.

### Why it needs to get done

To fix pylint's error `too-many-lines` on `test_oci.py`

tigarmo:
This commit cleans up oci.py by removing all functions related to processing files and directories for new OCI layers and placing them in a new layers.py module. This makes sense because the code is largely OCI-independent (working off of filesystem directories) and this gives us a cleaner separation of concerns.

The corresponding unit tests are also removed from test_oci.py and placed in the new test_layers.py module. There are no functional changes, but the tests themselves are lightly updated to reflect the fact that archive_layer() is a free function.

Fixes #296

-----


rockcraft/layers.py
```diff
@@ -0,0 +1,258 @@
...
+from typing import DefaultDict, Dict, List, Optional, Tuple
+
+from craft_cli import emit
+from craft_parts.executor.collisions import paths_collide
+from craft_parts.overlays import overlays
+
+from rockcraft import errors
+
+
+def archive_layer(
```
tigarmo, line 31: the functions here are all moved from `oci.py`, with no functional changes.

tests/unit/test_layers.py
```diff
@@ -0,0 +1,327 @@
...
+    rootfs_dir = tmp_path / "rootfs"
+    rootfs_dir.mkdir()
+    # In the base layer "bin" is a symlink to "usr/bin"
+    (rootfs_dir / "usr/bin").mkdir(parents=True)
+    (rootfs_dir / "bin").symlink_to("usr/bin")
+
+    return layer_dir, rootfs_dir
+
+
+def test_archive_layer_directories(tmp_path):
```
tigarmo, line 62: the tests here are all moved over from `test_oci.py`, with no functional changes and only minor fixes: for example, this was originally named `test_add_layer_directories()` and received a fixture to inspect the tarball. Such fixture is not needed anymore because the function receives the tarpath as a parameter.

tests/unit/test_layers.py
```diff
@@ -0,0 +1,332 @@
...
+    directories end up as the same target but the directories have different
+    ownership/permissions.
+    """
+    layer_dir, rootfs_dir = duplicate_dirs_setup(tmp_path)
+
+    # Change the default permissions of the directories that will end up as
+    # "/usr/bin/dir1", to ensure that they are different.
+    if sys.platform == "win32":
+        # On Windows, chmod() can only change a directory's read-only status.
+        (layer_dir / "bin/dir1").chmod(stat.S_IWRITE)
```
tigarmo, line 267: ok I lied; when porting the tests over I also took the opportunity to ensure that these tests also work on Windows (they were skipped before because the whole test_oci.py module skips Windows). This is a change to support that.

cjdcordeiro: nice

cmatsuoka (approved):

PR #417, opened 2023-11-27
feat: prune primed files if they exist on the base

### LINKED ISSUE #363: when packing, the resulting OCI layer from the parts may have content that is already in the Ubuntu base
### Bug Description

See https://chat.charmhub.io/charmhub/pl/o5wxpb65ffbfzy7bcmi8kzftzy

### To Reproduce

See https://chat.charmhub.io/charmhub/pl/o5wxpb65ffbfzy7bcmi8kzftzy

### Environment

Nothing in special

### rockcraft.yaml

```shell
See https://chat.charmhub.io/charmhub/pl/o5wxpb65ffbfzy7bcmi8kzftzy
```

### Relevant log output

```shell
The build will be fine, but the resulting ROCK will be larger than necessary and fail the efficiency tests. Example: https://github.com/canonical/oci-factory/actions/runs/5940848690/job/16110358625
```

----

## Proposal

We should, at some point before packing the ROCK, cross check the primed contents of the parts with what's already inside the base Ubuntu image. When those contents are equal, we simply don't include them in the final squashed layer that is going to be inserted into the ROCK. 

Here are some possibilities to explore:
 - using a sort of "phantom layer" that contains the base Ubuntu image, such that the craft lifecycle considers the Ubuntu base when doing the conflict resolution it already does
 - just before packing the ROCK, add a new function to compare the primed content and the Ubuntu base's content

tigarmo:
**Note** The first commit is from PR #416 and doesn't need to be reviewed here

------------

This commit adds a lifecycle callback that "prunes" (removes) those
primed files that already exist (with same contents and permissions) in
the base layer. This is necessary because the dependency resolver
doesn't look into the base layer when deciding dependency cutoffs (and
also to support chisel slices, which only look at the target install
dir). Some notes:

1) This first implementation only looks are regular files (no symlinks
and directories), as those are the simplest to handle and provide the
biggest size "wins". If necessary, we can address symlinks and/or
directories in follow-up work.
2) This could have been implemented during the archiving of the new
layer, where we already loop on the files and look at the base layer (to
handle things like the usrmerge). We decided to implement this as part
of the lifecycle to improve the "clarity" and "debuggability" of
Rockcraft runs: after running the lifecycle, the prime directory
contains exactly what's going to be packed in the new layer.

The resulting ROCK should be "semantically" equivalent, but smaller. For
instance, the Pyfiglet ROCK in docs/tutorials/code/pyfiglet goes from
69MB to 53MB, with no functional changes.

Fixes #363

### DIFF
--- a/rockcraft/layers.py
+++ b/rockcraft/layers.py
@@ -19,11 +19,12 @@ import os
 import tarfile
 from collections import defaultdict
 from pathlib import Path
-from typing import DefaultDict, Dict, List, Optional, Tuple
+from typing import DefaultDict, Dict, List, Optional, Set, Tuple
 
 from craft_cli import emit
 from craft_parts.executor.collisions import paths_collide
 from craft_parts.overlays import overlays
+from craft_parts.permissions import Permissions
 
 from rockcraft import errors
 
@@ -35,8 +36,8 @@ def archive_layer(
 ) -> None:
     """Prepare new OCI layer by archiving its content into tar file.
 
-    :param new_layer_dir: path to the content to be archived into a layer
-    :param temp_tar_file: path to the temporary tar file holding the archived content
+    :param new_layer_dir: path to the content to be archived into a layer.
+    :param temp_tar_file: path to the temporary tar file holding the archived content.
     :param base_layer_dir: optional path to the filesystem containing the extracted
         base below this new layer. Used to preserve lower-level directory symlinks,
         like the ones from Debian/Ubuntu's usrmerge.
@@ -52,6 +53,37 @@ def archive_layer(
             filepath = layer_paths[arcname]
             emit.debug(f"Adding to layer: {filepath} as '{arcname}'")
             tar_file.add(filepath, arcname=arcname, recursive=False)
+
+
+def prune_prime_files(prime_dir: Path, files: Set[str], base_layer_dir: Path) -> None:
+    """Remove (prune) files in a prime directory if they exist in the base layer.
+
+    Given a set of filenames ``files``, this function will remove (prune) all those
+    filenames from prime dir ``prime_dir`` if the corresponding sub path exists in
+    the extracted base layer directory ``base_layer_dir`` with same contents and
+    permissions.
+
+    For example, "{prime_dir}/dir/subdir/file1" will be pruned if the matching file
+    "{base_layer_dir}/dir/subdir/file1" exists and has the same contents, owner,
+    group, and permission bits.
+
+    :param prime_dir: The directory containing the lifecycle's primed contents.
+    :param files: The set of filenames added to ``prime_dir``, as provided by
+        the corresponding post_step lifecycle callback.
+    :param base_layer_dir: The directory where the base layer was extracted.
+    """
+    emit.debug("Pruning primed files that already exist on base layer...")
+    for filename in files:
+        base_layer_file = base_layer_dir / filename
+        if base_layer_file.is_file():
+            prime_file = prime_dir / filename
+            if _all_compatible_files([base_layer_file, prime_file]):
+                emit.debug(f"Pruning: {prime_file} as it exists on the base")
+                prime_file.unlink()
+            else:
+                emit.debug(
+                    f"{prime_file} exists on the base but with different contents or permissions"
+                )
 
 
 def _gather_layer_paths(
@@ -251,8 +283,19 @@ def _all_compatible_files(paths: List[Path]) -> bool:
 
     first_file = paths[0]
 
+    permissions_first = [_get_permissions(first_file)]
+
     for other_file in paths[1:]:
-        if paths_collide(str(first_file), str(other_file)):
+        permissions_other = [_get_permissions(other_file)]
+        if paths_collide(
+            str(first_file), str(other_file), permissions_first, permissions_other
+        ):
             return False
 
     return True
+
+
+def _get_permissions(filename: Path) -> Permissions:
+    """Create a Permissions object for a given Path."""
+    stat = os.stat(filename)
+    return Permissions(owner=stat.st_uid, group=stat.st_gid, mode=oct(stat.st_mode))
--- a/rockcraft/services/lifecycle.py
+++ b/rockcraft/services/lifecycle.py
@@ -25,10 +25,11 @@ from typing import cast
 from craft_application import LifecycleService
 from craft_archives import repo
 from craft_cli import emit
-from craft_parts import Features, callbacks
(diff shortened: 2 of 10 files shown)


rockcraft/services/lifecycle.py
```diff
@@ -60,6 +61,7 @@ def setup(self) -> None:
             package_repositories=project.package_repositories or [],
             project_name=project.name,
             project_vars=project_vars,
+            rootfs_dir=image_info.base_layer_dir,
```
tigarmo, line 64: @cmatsuoka I had to add this because it looks like the `base_layer_dir` parameter doesn't end up in the ProjectInfo. Can we make that change in craft-parts?

rockcraft/layers.py
```diff
@@ -54,6 +55,32 @@ def archive_layer(
...
+    Given a set of filenames ``files``, this function will remove (prune) all those
+    filenames from prime dir ``prime_dir`` if the corresponding sub path exists in
+    the extracted base layer directory ``base_layer_dir`` with same contents and
+    permissions.
+
+    For example, "{prime_dir}/dir/subdir/file1" will be pruned if the matching file
+    "{base_layer_dir}/dir/subdir/file1" exists and has the same contents, owner,
+    group, and permission bits.
+    """
+    emit.debug(f"Pruning contents of {prime_dir} due to base layer {base_layer_dir}...")
```
cjdcordeiro, line 70: This sentence might need some love :) it's not "due to" but something like "that conflict with"

rockcraft/layers.py
```diff
@@ -54,6 +55,32 @@ def archive_layer(
...
+    """Remove (prune) files in a prime directory if they exist in the base layer.
+
+    Given a set of filenames ``files``, this function will remove (prune) all those
+    filenames from prime dir ``prime_dir`` if the corresponding sub path exists in
+    the extracted base layer directory ``base_layer_dir`` with same contents and
+    permissions.
+
+    For example, "{prime_dir}/dir/subdir/file1" will be pruned if the matching file
+    "{base_layer_dir}/dir/subdir/file1" exists and has the same contents, owner,
+    group, and permission bits.
```
cjdcordeiro, line 68: I think you should add the ":param" docstring, so that 1) it is consistent with other docstrings and 2) we can generate docs from it if need be

rockcraft/layers.py
```diff
@@ -251,8 +278,19 @@ def _all_compatible_files(paths: List[Path]) -> bool:
...
     first_file = paths[0]
 
+    permissions_first = [_get_permissions(first_file)]
+
     for other_file in paths[1:]:
-        if paths_collide(str(first_file), str(other_file)):
+        permissions_other = [_get_permissions(other_file)]
+        if paths_collide(
+            str(first_file), str(other_file), permissions_first, permissions_other
+        ):
```
cjdcordeiro, line 287: it would be useful to have a debug log to print what those files are and their permissions

tests/spread/general/prune/check_layers.py
```diff
@@ -0,0 +1,51 @@
...
+    new_layer = first
+
+intersection = new_layer.intersection(base_layer)
+
+if intersection:
+    # It's tricky to check that the intersection is completely empty: the files
+    # come from different sources that can be momentarily out of sync. Therefore,
+    # as a "best effort" here try to check that some files that are not expected
+    # to change often are *not* in both sets.
+    constant_files = {
```
cjdcordeiro, line 42: is this really needed? it feels a bit fragile.

tests/spread/general/prune/check_layers.py
cjdcordeiro, line 1: I think this script can be simplified. Here are a few tips:
 - to know which layer is the base one, you can simply inspect the image - e.g. `skopeo inspect oci-archive:foo.rock | jq .Layers` -> [0] is base
 - tools like `dive` will immediately give you duplicated contents. I guess you could achieve the same with `pkgdiff` or a simple comparison of `tar -tvf <layer>`. So why do we need those "constant_files"?

cjdcordeiro (changes requested):
Very clean :) nice!

I just left a few nitpicks and a general thought on whether we actually need the `check_layers.py` script in the spread test

tests/spread/general/prune/check_layers.py
tigarmo: > to know which layer is the base one, you can simply inspect the image - e.g. skopeo inspect oci-archive:foo.rock | jq .Layers -> [0] is base 

ah good, ty

> So why do we need those "constant_files"?

Because the list of files that have the same name but different contents will change with time, as the sources of the files (the Ubuntu archive and the base image) get "de-synced" in different ways.
I support a different way to test this would be to check for duplicate names, and then for each duplicate name make sure that the two files (from the base layer and from the "lifecycle layer") are actually different.

rockcraft/services/lifecycle.py
cmatsuoka: Yes, it's a reasonable thing to be there.

tests/spread/general/prune/check_layers.py
cjdcordeiro: Thanks for the clarification.

> I support a different way to test this would be to check for duplicate names, and then for each duplicate name make sure that the two files (from the base layer and from the "lifecycle layer") are actually different.

Y, that sounds much easier and more future-proof than inferring which files are "constant".

rockcraft/layers.py
tigarmo: In my opinion "that conflict with" is misleading: from craft-part's point-of-view, a conflict is when two files have the same name but different contents/permissions. In this case the pruning happens exactly when there is a filename match but _no_ conflict.
How about "Pruning primed files that already exist on base layer..."?

rockcraft/layers.py
tigarmo: the filenames are already printed by the calling code (in `prune_prime_files()`)
Any suggestions on how to _succinctly_ log the stats of the files?

tests/spread/general/prune/check_layers.py
tigarmo: took another pass at this, please take a look

rockcraft/layers.py
cjdcordeiro: if the files are already printed, that's good enough

tests/spread/general/prune/check_layers.py
cjdcordeiro: it got bigger xD

but I think the logic is simpler now ;)

thanks

cjdcordeiro (approved):
Noice! thanks for the changes.

PR #418, opened 2023-11-27
WIP support BuildPlanner

tigarmo:

-----

### DIFF
--- a/rockcraft/models/project.py
+++ b/rockcraft/models/project.py
@@ -36,6 +36,7 @@ import pydantic
 import spdx_lookup  # type: ignore
 import yaml
 from craft_application.models import BuildInfo
+from craft_application.models import BuildPlanner as BaseBuildPlanner
 from craft_application.models import Project as BaseProject
 from craft_archives import repo
 from craft_providers import bases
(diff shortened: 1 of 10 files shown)


rockcraft/models/project.py
```diff
@@ -102,6 +103,136 @@ def _validate_platform_set(cls, values: Mapping[str, Any]) -> Mapping[str, Any]:
...
 
+class BuildPlanner(BaseBuildPlanner):
+
+    platforms: Dict[str, Any]
+    base: Literal["bare", "ubuntu@20.04", "ubuntu@22.04"]
+    build_base: Optional[Literal["ubuntu@20.04", "ubuntu@22.04"]]
+
+    @pydantic.validator("build_base", always=True)
+    @classmethod
+    def _validate_build_base(cls, build_base: Optional[str], values: Any) -> str:
```
tigarmo, line 114: would probably need to move this validation code somewhere else so that both `Project` and `BuildPlanner` can use it

tests/unit/test_application.py
```diff
@@ -49,7 +49,7 @@ def test_application_expand_environment(default_application, new_dir):
     project_file = Path(new_dir) / "rockcraft.yaml"
     project_file.write_text(ENVIRONMENT_YAML)
 
-    project = default_application.project
+    project = default_application.get_project("amd64", "amd64", "amd64")
```
tigarmo, line 52: this part felt a bit off: it feels like if I pass the `platform` I shouldn't have to also part `build_on` and `build_for`

tigarmo (commented):
Some tests broke because the ordering of the attributes in the exported Project changed. I think this is because the base `Project` now has `platform` and `build_base`, was that an oversight?

tests/unit/test_application.py
cmatsuoka: Removed `build_on` as it's set to the host arch. Kept `build_for` to disambiguate between platform name and architecture name.

PR #422, opened 2023-11-28
Integrate with Starbase

syu-w:


pyproject.toml
```diff
@@ -30,70 +153,65 @@ use_parentheses = true
...
+exclude_also = [
+    "if (typing\\.)?TYPE_CHECKING:",
+]
 
 [tool.pyright]
-ignore = ["docs/sphinx-starter-pack"]
+strict = ["rockcraft"]
+pythonVersion = "3.10"
+pythonPlatform = "Linux"
+exclude = [
```
sergiusens, line 176: exclude the starter pack here as well

mr-cal (changes requested):

docs/_static/favicon.png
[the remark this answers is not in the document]
syu-w: docs build report this file is missing.

tox.ini
```diff
@@ -146,4 +146,4 @@ base = docs
 labels = lint
 allowlist_externals = bash, xargs
 commands_pre = bash -c '{[lint-docs]find} > {env_tmp_dir}/lint_docs_files'
-commands = xargs --no-run-if-empty --arg-file {env_tmp_dir}/lint_docs_files sphinx-lint --max-line-length 80 --enable all --ignore "docs/sphinx-starter-pack/" {posargs}
\ No newline at end of file
+commands = xargs --no-run-if-empty --arg-file {env_tmp_dir}/lint_docs_files sphinx-lint --max-line-length 80 --enable all --ignore "docs/sphinx-starter-pack/" --ignore "README.rst" {posargs}
```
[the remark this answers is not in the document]
syu-w: The long url will make the linter fail, but the badge url does not seem to be able to be wrapped. So temporarily ignore it.

tests/integration/plugins/test_python_plugin.py
```diff
@@ -111,9 +111,13 @@ def test_python_plugin_ubuntu(base, tmp_path, run_lifecycle):
     assert list(bin_dir.glob("python*")) == []
 
     # Check the shebang in the "hello" script
-    expected_shebang = "#!/bin/python3"
+    # In test env this could be replaced with sh that exec python by the venv
+    expected_shebang = ("#!/bin/python3", "#!/bin/sh")
```
cmatsuoka, line 115: I'm a bit confused here, how the starbase migration affected this test?

cmatsuoka (commented):
This is a very large changeset and difficult to examine in detail. It seems that the majority of changes are auto-generated so I just did a quick pass on those. I didn't find obvious red flags so if this is passing tests it's probably ok, and existing issues can be dealt with in follow-ups. It's a good idea however to check the comments and address them if applicable.

rockcraft/services/lifecycle.py
```diff
@@ -18,13 +18,14 @@
...
 import contextlib
 from pathlib import Path
-from typing import cast
+from typing import Any, cast
 
 from craft_application import LifecycleService
 from craft_archives import repo
 from craft_cli import emit
-from craft_parts import Features, Step, callbacks
+from craft_parts import Features, LifecycleManager, Step, callbacks
```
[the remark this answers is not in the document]
syu-w: I got:
```
rockcraft/services/lifecycle.py:35:29: TCH004 Move import `craft_parts.LifecycleManager` out of type-checking block. Import is used for more than type hinting.
```

rockcraft/cli.py
```diff
@@ -39,7 +39,7 @@ def run() -> int:
     return app.run()
 
 
-def _create_app():
+def _create_app():  # noqa: ANN202
```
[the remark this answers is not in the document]
syu-w: Fixed

snap/snapcraft.yaml
```diff
@@ -69,11 +66,22 @@ parts:
     override-build: |
       ${SNAP}/libexec/snapcraft/craftctl default
 
-      version="$(python3 setup.py --version)"
+      # craft-parts changed the python symlink in venv, and unusable at the time
+      rm -f "${CRAFT_PART_INSTALL}/bin/python3"
+      ln -s "${CRAFT_PART_INSTALL}/../../../stage/usr/bin/python3" "${CRAFT_PART_INSTALL}/bin/python3"
+
+      # stable Snapcraft not support pure pyproject.toml, install it manually
```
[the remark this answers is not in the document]
mr-cal: Then can you update the comment to reflect that?

tox.ini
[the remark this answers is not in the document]
mr-cal: I mean, either we should be fixing the README or disable linting of the README, but not both. I would have preferred it if you had done one or the other.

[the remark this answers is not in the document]
mr-cal: We can leave it as-is, now that it's already done.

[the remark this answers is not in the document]
syu-w: Fixed

snap/snapcraft.yaml
```diff
@@ -69,11 +66,22 @@ parts:
     override-build: |
       ${SNAP}/libexec/snapcraft/craftctl default
 
-      version="$(python3 setup.py --version)"
+      # craft-parts changed the python symlink in venv, and unusable at the time
```
tigarmo, line 69: It feels like this override-build script is worse than just having an "empty" `setup.py` file. If we kept the setup.py, we wouldn't have to hack around with symlinks here nor build the project manually.

pyproject.toml
```diff
@@ -1,17 +1,107 @@
...
     "Intended Audience :: Developers",
-    "License :: OSI Approved :: GNU Lesser General Public License v3 (LGPLv3)",
+    "License :: OSI Approved :: GNU General Public License v3 (GPLv3)",
     "Operating System :: MacOS :: MacOS X",
     "Operating System :: POSIX :: Linux",
     "Programming Language :: Python :: 3",
     "Programming Language :: Python :: 3.10",
 ]
 requires-python = ">=3.10"
+dependencies = [
```
tigarmo, line 20: I don't know if I agree with moving the frozen requirements here. The problem is that this ties Rockcraft to transitive dependencies forever, even if we don't need them anymore, because the application needs to freeze all of its requirements.
What was the reasoning to remove the requirements-*.txt files?

.github/workflows/tests.yaml
```diff
@@ -1,108 +1,143 @@
...
+  lint:
     runs-on: ubuntu-22.04
     steps:
-      - name: Checkout code
+      - name: Checkout
         uses: actions/checkout@v4
         with:
           fetch-depth: 0
           submodules: true
+      - name: conventional commits
```
tigarmo, line 22: should be `Conventional commits`, to be consistent with the other step names

tests/integration/plugins/test_python_plugin.py
tigarmo: This one is definitely a problem: the `#!/bin/python3` shebang should be set [by the Python plugin](https://github.com/canonical/rockcraft/blob/main/rockcraft/plugins/python_plugin.py#L87) and the purpose of the test is to verify this. `#!/bin/sh` is not an acceptable value, we need to look into this more

docs/how-to/code/build-docs/task.yaml
```diff
@@ -1,34 +0,0 @@
-###########################################
-# IMPORTANT
-# Comments matter!
-# The docs use the wrapping comments as 
-# markers for including said instructions 
-# as snippets in the docs.
-###########################################
-summary: test the "Build the docs" guide
```
tigarmo, line 8: Why did this get deleted?

rockcraft/models/project.py
```diff
@@ -337,7 +339,12 @@ def _validate_entrypoint_service(
                 "a valid Pebble service."
             )
 
-        command = values.get("services")[entrypoint_service].command
+        services = values.get("services")
```
tigarmo, line 354: Probably clearer to do `command = values["services"][entrypoint_service].command`, because the check just before this code already ensures that the dict entries exist

rockcraft/plugins/python_plugin.py
```diff
@@ -75,7 +75,7 @@ class PythonPlugin(python_plugin.PythonPlugin):
     @override
     def _should_remove_symlinks(self) -> bool:
         """Overridden because for ubuntu bases we must always remove the symlinks."""
-        return self._part_info.base != "bare"
+        return bool(self._part_info.base != "bare")
```
tigarmo, line 78: this is an odd one, what was mypy complaining about?

tigarmo (changes requested):

rockcraft/plugins/python_plugin.py
syu-w: https://github.com/python/mypy/issues/5697#issuecomment-425738430

docs/how-to/code/build-docs/task.yaml
tigarmo: But this test verifies that the [documentation on how to build the docs](https://canonical-rockcraft.readthedocs-hosted.com/en/latest/how-to/build-docs/) is working. It's used by people that want to contribute docs, etc, so it needs to be updated

syu-w: Well, then I also need to update the docs, because the command now is `tox -e build-docs`

tigarmo: yeap

tigarmo: That's the point of these tests; if you update the test, the documentation will update automatically because it consumes lines from this task.yaml.

tests/integration/plugins/test_python_plugin.py
syu-w:
```
#!/bin/sh
'''exec' /run/user/1000/tox_tmp/integration-py3.10/pytest-of-ubuntu/pytest-0/test_python_plugin_ubuntu_ubun1/parts/my-part/install/bin/python3 "$0" "$@"
' '''
# -*- coding: utf-8 -*-
import re
import sys
from hello.__main__ import main
if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw|\.exe)?$', '', sys.argv[0])
    sys.exit(main())
```

The file become like this if run under `tox`.

tigarmo: Right I understand that, but we have to figure out a way around that because as it is the test is not checking what it should be checking

syu-w: I don't see any configuration can disable that. We may have to get a separate test for it without run in tox.

syu-w: Well, I can create a spread test for that, but it is not possible to do in tox.

syu-w: Added

docs/how-to/code/build-docs/task.yaml
syu-w: Updated

medubelko:
@tigarmo Can we close this?

mr-cal:
Closing as stale.

PR #436, opened 2023-12-05
feat(docs): add definition for overlay_step in the docs

linostar:

-----

I added the missing definition for the overlay-step in explanation docs.

Note that the overlay_parameters common doc mentions `overlay-filters` field, but apparently that's a typo and it should be `overlay-files`. I fixed that in rockcraft docs, but it needs to be fixed in craft-parts docs too.

### DIFF
--- a/docs/explanation/overlay-step.rst
+++ b/docs/explanation/overlay-step.rst
@@ -8,6 +8,12 @@ The overlay step is specific to ROCKs and is configured with overlay parameters.
 To learn more about pull, build, stage and prime see
 :doc:`/reference/part_properties`
 
+The overlay step provides the means to modify the base filesystem before the
+build step is applied. If ``overlay-packages`` is used, those packages will be
+installed first. ``overlay-script`` will run the provided script in this step.
+The location of the overlay is made available in ${CRAFT_OVERLAY} environment
+variable. ``overlay-files`` can be used to specify which files will be migrated
+to the next steps, and when omitted its default value will be ``"*"``.
 
 .. Include a section about overlay parameters from the Craft Parts documentation.
 .. include:: /common/craft-parts/overlay_parameters.rst
--- a/docs/explanation/overlay-step.rst
+++ b/docs/explanation/overlay-step.rst
@@ -11,9 +11,9 @@ To learn more about pull, build, stage and prime see
 The overlay step provides the means to modify the base filesystem before the
 build step is applied. If ``overlay-packages`` is used, those packages will be
 installed first. ``overlay-script`` will run the provided script in this step.
-The location of the overlay is made available in ${CRAFT_OVERLAY} environment
-variable. ``overlay-files`` can be used to specify which files will be migrated
-to the next steps, and when omitted its default value will be ``"*"``.
+The location of the overlay is made available in the ${CRAFT_OVERLAY}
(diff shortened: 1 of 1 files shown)


cjdcordeiro (commented):
that's a nice summary.

The actual docs are in https://craft-parts.readthedocs.io/en/latest/explanation/overlays.html. Wouldn't it be better to simply inherit them? They kind of overlap already, and the ones from the craft-parts seem more complete.

linostar:
> that's a nice summary.
> 
> The actual docs are in https://craft-parts.readthedocs.io/en/latest/explanation/overlays.html. Wouldn't it be better to simply inherit them? They kind of overlap already, and the ones from the craft-parts seem more complete.

Yeah, I actually based my summary on that explanation page.
Rockcraft's overlay page only inherits the Overlay Parameters section, but we can include the rest of the sections as well if you want.

cjdcordeiro:
I'll let @tigarmo be the judge of that

docs/explanation/overlay-step.rst
```diff
@@ -8,6 +8,12 @@ The overlay step is specific to ROCKs and is configured with overlay parameters.
 To learn more about pull, build, stage and prime see
 :doc:`/reference/part_properties`
 
+The overlay step provides the means to modify the base filesystem before the
+build step is applied. If ``overlay-packages`` is used, those packages will be
+installed first. ``overlay-script`` will run the provided script in this step.
+The location of the overlay is made available in ${CRAFT_OVERLAY} environment
```
tigarmo, line 14:
```suggestion
The location of the overlay is made available in the ${CRAFT_OVERLAY} environment
```

docs/explanation/overlay-step.rst
```diff
@@ -8,6 +8,12 @@ The overlay step is specific to ROCKs and is configured with overlay parameters.
...
+variable. ``overlay-files`` can be used to specify which files will be migrated
```
tigarmo, line 15:
```suggestion
variable. ``overlay`` can be used to specify which files will be migrated
```

I believe we prefer `step` over `step-files`

tigarmo:
@linostar do you want to experiment with including craft-part's Overlay explanation? Like [this](https://github.com/canonical/rockcraft/blob/main/docs/reference/part_properties.rst?plain=1#L6)

linostar:
> @linostar do you want to experiment with including craft-part's Overlay explanation? Like [this](https://github.com/canonical/rockcraft/blob/main/docs/reference/part_properties.rst?plain=1#L6)

Sure!

linostar:
@tigarmo To be able to include overlay_step doc from craft-parts, some changes need to be done to craft-parts repo first, because the aforementioned doc isn't included in the craft_parts_docs package. We have a couple of options:
- move `overlays.rst` to `docs/base` in craft-parts so it becomes included in craft_parts_docs package
- extend the directories included in craft_parts_docs to include other dirs, or at least the overlays.rst file in its current dir

What you tell me which option would you prefer?

tigarmo:
Ah I see. We need to take a better look at how these craft-parts docs are shared, so for now I propose we just merge your paragraph as-is.

tigarmo (approved):

SamirPS (approved):

cjdcordeiro (approved):
lgtm thanks

PR #437, opened 2023-12-05
chore(linter): update to 3.10 style

syu-w:


tigarmo:
Thank you @syu-w . I have two suggestions to make 
1) Put the command that you executed to do the formatting on the commit message, so we know how we got here
2) The PR description says "mostly" automated but I don't know what's automated and what isn't. I think we should split this into two commits - the first is automated, and the second has the manual changes.

syu-w:
OK, I split it into 2 commits.

tigarmo (approved):
Thanks!

cmatsuoka (approved):

PR #440, opened 2023-12-06
refactor(project): drop ProjectValidationError

tigarmo:
Drop rockcraft's ProjectValidationError in favour of craft-application's
CraftValidationError. This lets us remove a bunch of duplicated code
related to formatting the errors and aligns us more closely with
upstream.

### DIFF
--- a/tests/unit/test_project.py
+++ b/tests/unit/test_project.py
@@ -24,13 +24,14 @@ from typing import Any
 import pydantic
 import pytest
 import yaml
+from craft_application.errors import CraftValidationError
 from craft_application.models import BuildInfo
 from craft_parts import Features
 from craft_providers.bases import BaseName
 
-from rockcraft.errors import ProjectLoadError, ProjectValidationError
-from rockcraft.models import Project, load_project
-from rockcraft.models.project import INVALID_NAME_MESSAGE, Platform
+from rockcraft.errors import ProjectLoadError
+from rockcraft.models import Project
+from rockcraft.models.project import INVALID_NAME_MESSAGE, Platform, load_project
 from rockcraft.pebble import Service
 
 _ARCH_MAPPING = {"x86": "amd64", "x64": "amd64"}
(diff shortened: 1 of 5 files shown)


tests/unit/test_project.py
```diff
@@ -426,33 +439,33 @@ def test_project_name_invalid(yaml_loaded_data, invalid_name):
...
     yaml_loaded_data["parts"]["foo"]["invalid"] = True
 
-    with pytest.raises(ProjectValidationError) as err:
-        Project.unmarshal(yaml_loaded_data)
+    with pytest.raises(CraftValidationError) as err:
+        load_project_yaml(yaml_loaded_data)
     assert str(err.value) == (
         "Bad rockcraft.yaml content:\n"
-        "- extra field 'invalid' not permitted in 'parts.foo' configuration"
+        "- extra field invalid not permitted in parts.foo configuration"
```
tigarmo, line 468: One issue that we should improve in craft-application is here: the original message (from rockcraft's version for `format_pydantic_errors()`) had single quotes around the yaml fields (`invalid` and `parts.foo`), which are now gone. I think they improve the clarity of the message

syu-w (approved):

tests/unit/test_project.py
```diff
@@ -140,7 +153,7 @@ def test_unmarshal_no_repositories(yaml_loaded_data):
 
     project = Project.unmarshal(yaml_loaded_data)
 
-    assert project.package_repositories == []
+    assert project.package_repositories is None
```
lengau, line 156: Do we care about the specific value, or would `assert not project.package_repositories` be sufficient?

tests/unit/test_project.py
lengau: Good point! https://github.com/canonical/craft-application/pull/157

lengau (approved):
This looks great! Always nice to see a PR with net removal of code

tests/unit/test_project.py
tigarmo: eh I guess I don't have a personal preference here

PR #441, opened 2023-12-06
fix: use correct error message for invalid version

### LINKED ISSUE #429: Error message for invalid versions is misleading
### Bug Description

The error message when a project has an invalid value for the `version` field is misleading. It says "Invalid name for ROCK (must contain only lowercase letters, numbers and hyphens) in field 'version'", but that restriction is for the `name` field.

### To Reproduce

Create a rockcraft.yaml with `version: invalid_version`

### Environment

rockcraft 1.0.1

### rockcraft.yaml

```shell
name: hello
title: Hello World
summary: An Hello World ROCK
description: |
  This is just an example of a Rockcraft project
  for a Hello World ROCK.
version: invalid_version
base: bare
build-base: ubuntu@22.04
license: Apache-2.0
platforms:
  amd64:

parts:
  hello:
    plugin: nil
```

### Relevant log output

```shell
Bad rockcraft.yaml content:
- Invalid name for ROCK (must contain only lowercase letters, numbers and hyphens) in field 'version'
```

tigarmo:
**Note:** The first commit bumps craft-application 1.2.0 and fixes a couple of expected messages in tests. I couldn't split this into two PRs because two of the tests (related to name validation) can only pass with the second commit.

-------------

The code refers to the 'name' message, but the bug was due to the
"error_msg_templates" in the Config: it made *every* regex validation
failure use the INVALID_NAME_MESSAGE message, even for those fields that
*aren't* the 'name'.

Since the "error_msg_templates" option is deprecated and gone in
Pydantic 2 anyway, remove it altogether and use the new
transform_pydantic_errors() hook from craft-application to preserve the
customisation of the 'name' field's message. This is necessary because
our 'name' regex is different than the 'name' regex in
craft-application's model (ours is a bit stricter), so we can just use
craft-application's message as that would be misleading.

Fixes #429

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -125,6 +125,8 @@ warn_untyped_fields = True
 # D213 Multi-line docstring summary should start at the second line (reason: pep257 default)
 # D215 Section underline is over-indented (reason: pep257 default)
 ignore = D105, D107, D203, D213, D215
+# Allow missing docstrings in methods that are overridden
+ignore_decorators = override
 
 [aliases]
 test = pytest
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -14,7 +14,7 @@ click==8.1.7
 codespell==2.2.6
 colorama==0.4.6
 coverage==7.3.2
-craft-application==1.1.0
+craft-application==1.2.0
 craft-archives==1.1.3
 craft-cli==2.5.0
 craft-parts==1.26.0
--- a/requirements-doc.txt
+++ b/requirements-doc.txt
@@ -5,7 +5,7 @@ bracex==2.4
 certifi==2023.7.22
 charset-normalizer==3.3.1
 colorama==0.4.6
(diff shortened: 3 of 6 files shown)


syu-w (approved):

setup.cfg
```diff
@@ -125,6 +125,8 @@ warn_untyped_fields = True
 # D213 Multi-line docstring summary should start at the second line (reason: pep257 default)
 # D215 Section underline is over-indented (reason: pep257 default)
 ignore = D105, D107, D203, D213, D215
+# Allow missing docstrings in methods that are overridden
+ignore_decorators = override
```
lengau, line 129: Oooh nice. I think I'll include this in the ruff config in starbase. https://docs.astral.sh/ruff/settings/#pydocstyle-ignore-decorators

lengau (approved):
Looks good!

PR #444, opened 2023-12-07
mock out call to `shutil.which` 

addyess:
rather than expecting scopeo to be locally installed, mock out the call in unit tests


addyess:
@cjdcordeiro and @tigarmo this looks like an easy fix right?  I tried to run the unit tests of this project and failed when i didn't have scopeo installed

tigarmo:
Looks good to me, probably a blind spot of ours because well, I've never _not_ had skopeo installed hehe
Thanks.

tigarmo (approved):

cjdcordeiro (approved):
good catch :)

PR #445, opened 2023-12-08
docs: explain how the prime dir is packed

### LINKED ISSUE #430: Document how primed files become an OCI layer
### Request type

Enhancement

### What needs to get done

We currently do many transformations (prime pruning, usrmerge handling, etc) that affect the contents of the payload OCI layer. We should document those transformations.

### Documentation location

_No response_

### Additional information

_No response_

tigarmo:
Fixes #430

-----


docs/explanation/lifecycle-layer.rst
```diff
@@ -0,0 +1,198 @@
+
+From lifecycle to OCI layer
```
tigarmo, line 2: rendered version: https://canonical-rockcraft--445.com.readthedocs.build/en/445/explanation/lifecycle-layer/

docs/explanation/lifecycle-layer.rst
```diff
@@ -0,0 +1,198 @@
...
+
+Pruning the ``prime`` directory
+-------------------------------
+
+Once the lifecycle is finished, Rockcraft packs the contents of the prime
+directory as a new OCI layer, directly as if the prime directory was the
+filesystem root ``/``. One consequence of the inclusion of a stage-package's
+dependencies is that the prime directory ends up having many files that the base
+Ubuntu layer already has. This can be seen, for example, by using a tool like
+``dive``:
```
cjdcordeiro, line 58: I think it's better to have a link instead: [Dive](https://github.com/wagoodman/dive)

docs/explanation/lifecycle-layer.rst
```diff
@@ -0,0 +1,198 @@
...
+
+.. figure:: /_static/dive-efficiency.png
+   :width: 75%
+   :align: center
+   :alt: Rock diagram
```
cjdcordeiro, line 63:
```suggestion
   :alt: Dive reporting an inefficient image
```

docs/explanation/lifecycle-layer.rst
```diff
@@ -0,0 +1,198 @@
...
+
+What ``dive`` tells us is that about ``60 MB`` worth of files are *duplicated*
+between the base Ubuntu Jammy layer and the "primed" layer: for example, the
+file ``/usr/lib/x86_64-linux-gnu/libcrypto.so.3`` exists both in the base layer
+(as part of the base Ubuntu system) and in the primed layer (pulled in by
+belonging to a package that is an indirect dependency of ``python3-minimal``).
+
+Since version ``1.1.0``, Rockcraft will "prune" those files in the prime
+directory that also exist, with the same contents, ownership and permissions, in
+the base layer. The end-result is semantically the same, because the layers are
```
cjdcordeiro, line 73:
```suggestion
the base layer. The end result is semantically the same because the layers are
```

docs/explanation/lifecycle-layer.rst
```diff
@@ -0,0 +1,198 @@
...
+   Pruning: /root/prime/usr/lib/x86_64-linux-gnu/perl-base/unicore/lib/PatSyn/Y.pl as it exists on the base
+   Pruning: /root/prime/usr/lib/x86_64-linux-gnu/perl-base/unicore/lib/Dt/Init.pl as it exists on the base
+   Pruning: /root/prime/usr/share/perl5/Debconf/Element/Noninteractive/Multiselect.pm as it exists on the base
+   (...)
+
+
+``usrmerge`` and the lifecycle layer
+------------------------------------
+
+After pruning, the contents of prime directory are packed as a new OCI layer.
```
cjdcordeiro, line 91:
```suggestion
After pruning, the contents of the prime directory are packed as a new OCI layer.
```

docs/explanation/lifecycle-layer.rst
```diff
@@ -0,0 +1,198 @@
...
+In concrete terms, this means that the files and directories are added to a
+`tar archive`_, which means that each file (or directory) gets added to the
+archive together with the "destination" path that it should have when the
+archive is extracted.
+
+For most cases, the file's original path (relative to the root of the archive)
```
cjdcordeiro, line 97:
```suggestion
In most cases, the file's original path (relative to the root of the archive)
```

docs/explanation/lifecycle-layer.rst
```diff
@@ -0,0 +1,198 @@
...
+   (...)
+   Pruning: /root/prime/usr/lib/x86_64-linux-gnu/perl-base/unicore/lib/Sc/Gran.pl as it exists on the base
+   Pruning: /root/prime/usr/lib/x86_64-linux-gnu/perl-base/unicore/lib/Bc/EN.pl as it exists on the base
+   Pruning: /root/prime/usr/lib/x86_64-linux-gnu/perl-base/unicore/lib/PatSyn/Y.pl as it exists on the base
+   Pruning: /root/prime/usr/lib/x86_64-linux-gnu/perl-base/unicore/lib/Dt/Init.pl as it exists on the base
+   Pruning: /root/prime/usr/share/perl5/Debconf/Element/Noninteractive/Multiselect.pm as it exists on the base
+   (...)
+
+
+``usrmerge`` and the lifecycle layer
```
cjdcordeiro, line 88: This section is really useful. I wonder though if it isn't too in-depth for a regular reader? Unlike the OCI layering (which is a well-known design feature of all container images), this `usrmerge` thing is very specific to the internal workings of the tool. I think only a developer would ever worry about this, no?

Having said that, I don't really have a better place to put it, so I guess it is ok to keep it here unless you have a better idea

docs/explanation/lifecycle-layer.rst
```diff
@@ -0,0 +1,198 @@
...
+
+   (...)
+   Creating new layer
+   (...)
+   Skipping /root/prime/bin because it exists as a symlink on the lower layer
+   (...)
+   Adding to layer: /root/prime/bin/pebble as 'usr/bin/pebble'
+   (...)
+
+One final note is that none of this applies for rocks with ``bare`` bases, as
```
cjdcordeiro, line 192: well...what if I install the pebble snap? will rockcraft raise a conflict? or will the pruning system act?

cjdcordeiro (changes requested):
good work. very useful. just a few comments

docs/explanation/lifecycle-layer.rst
```diff
@@ -0,0 +1,198 @@
+
+From lifecycle to OCI layer
+===========================
+
+Rockcraft is a tool that creates OCI images using the same concepts and
+mechanisms that create snaps and charms: the lifecycle language from
+craft-parts. There is a significant difference between the works of the Craft
```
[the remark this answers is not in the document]
sergiusens: we should try and use _Craft Parts_ everywhere as that is the title of the library https://canonical-craft-parts.readthedocs-hosted.com/en/latest/

docs/explanation/lifecycle-layer.rst
```diff
@@ -0,0 +1,198 @@
...
+
+As we can see, the prime directory has the contents of the ``python3-minimal``
+package but also many of its dependencies, direct and otherwise.
+
+Pruning the ``prime`` directory
+-------------------------------
+
+Once the lifecycle is finished, Rockcraft packs the contents of the prime
+directory as a new OCI layer, directly as if the prime directory was the
+filesystem root ``/``. One consequence of the inclusion of a stage-package's
```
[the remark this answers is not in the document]
sergiusens: this is a tricky one, if referring to the keyword, the `stage-packages` (with the mono/verbatim hint) should be used, and perhaps mention that it is a parts keyword.

We do want a nice consistent way to referring to things, similar to "_prime_ directory" below

docs/explanation/lifecycle-layer.rst
sergiusens: I would put this at the top, and specify this only applies to ubuntu bases

docs/explanation/lifecycle-layer.rst
```diff
@@ -0,0 +1,198 @@
...
+   # (...)
+   base: ubuntu@22.04
+
+   parts:
+     python-part:
+       plugin: nil
+       stage-packages:
+         - python3-minimal
+
+This rock has Ubuntu Jammy as its base, and includes ``python3-minimal``.
```
sergiusens, line 29: Always refer to Ubuntu by the release name and not code name, this is Ubuntu 22.04

docs/explanation/lifecycle-layer.rst
tigarmo: The original motivation for this page is to document how Rockcraft packs a layer, and with Ubuntu bases there's no way around talking about `usrmerge` because we've had many bug reports and user confusion about this.

This doc isn't meant to be read by everyone, but if someone is having an issue and wants to understand things in more detail then it's useful to have this page to point to.

I actually left out a bunch of extra details, this is the "simplified" version ;)

linostar (approved):

docs/explanation/lifecycle-layer.rst
```diff
@@ -0,0 +1,207 @@
+
+From lifecycle to OCI layer
+===========================
+
+Rockcraft is a tool that creates OCI images using the same concepts and
+mechanisms that create snaps and charms: the lifecycle language from
+Craft Parts. There is a significant difference between the works of the Craft
+lifecycle and the OCI specification, and one of Rockcraft's jobs is to bridge
+the gap between these two worlds. This page describes how this is accomplished.
```
lengau, line 9:
```suggestion
Craft Parts. There is a significant difference between the way the Craft lifecycle
works and the OCI specification, and one of Rockcraft's jobs is to bridge
the gap between these two worlds. This page describes how this is accomplished.
```

docs/explanation/lifecycle-layer.rst
```diff
@@ -0,0 +1,207 @@
...
+   bin  etc  lib  lib64  sbin  usr  var
+   $ ls usr/bin/
+   debconf               debconf-copydb          debconf-show  dpkg-divert              dpkg-realpath      dpkg-trigger  py3clean     python3
+   debconf-apt-progress  debconf-escape          dpkg          dpkg-maintscript-helper  dpkg-split         perl          py3compile   python3.10
+   debconf-communicate   debconf-set-selections  dpkg-deb      dpkg-query               dpkg-statoverride  perl5.34.0    py3versions  update-alternatives
+
+As we can see, the prime directory has the contents of the ``python3-minimal``
+package but also many of its dependencies, direct and otherwise. Once the
+lifecycle is finished, Rockcraft packs the contents of the prime directory as a
+new OCI layer, directly as if the prime directory was the filesystem root ``/``.
```
lengau, line 52:
```suggestion
new OCI layer, directly as if the prime directory were the filesystem root ``/``.
```

docs/explanation/lifecycle-layer.rst
```diff
@@ -0,0 +1,207 @@
+
+From lifecycle to OCI layer
```
sergiusens, line 2: We already have an Overlay Step explanation, this could just be Prime Step

docs/explanation/lifecycle-layer.rst
```diff
@@ -0,0 +1,207 @@
...
+Rockcraft is a tool that creates OCI images using the same concepts and
+mechanisms that create snaps and charms: the lifecycle language from
+Craft Parts. There is a significant difference between the works of the Craft
+lifecycle and the OCI specification, and one of Rockcraft's jobs is to bridge
+the gap between these two worlds. This page describes how this is accomplished.
+
+.. note::
+   It is not necessary to know these details to use the tool effectively, but
+   they might illuminate some concepts and help understand *why* the contents of
+   a given rock are the way they are.
```
sergiusens, line 14: this is superfluous as you shouldn't need to read any explanation documentation for correct use of the tool

sergiusens: I disagree with most of the comments in this document about it being too much, especially if I go and read about explanation on https://diataxis.fr/explanation/

That said, this clarification is essentially what explanation can cover which is why I say it is superfluous.

docs/explanation/lifecycle-layer.rst
```diff
@@ -0,0 +1,207 @@
...
+   :align: center
+   :alt: Dive reporting an inefficient image
+
+What ``dive`` tells us is that about ``60 MB`` worth of files are *duplicated*
+between the base Ubuntu 22.04 layer and the "primed" layer: for example, the
+file ``/usr/lib/x86_64-linux-gnu/libcrypto.so.3`` exists both in the base layer
+(as part of the base Ubuntu system) and in the primed layer (pulled in by
+belonging to a package that is an indirect dependency of ``python3-minimal``).
+
+Starting from version ``1.1.0``, Rockcraft will "prune" those files in the prime
```
lengau, line 79: How far out is Rockcraft 1.1? Is it worth just putting this in present tense so we don't have to update later?

tigarmo: The commit that merges this PR will likely get the `1.1.0` tag

docs/explanation/lifecycle-layer.rst
tigarmo: > this is superfluous as you shouldn't need to read any explanation documentation for correct use of the tool

I don't think this is as cut-and-dry as you suggest, but more importantly I don't think a given user *knows* that the explanations are not necessary read. I imagine that many Rockcraft users will not be familiar with Diatáxis, and in my opinion they shouldn't have to be.

docs/explanation/lifecycle-layer.rst
```diff
@@ -0,0 +1,207 @@
...
+The ``bash`` binary links to multiple dynamic libraries, but has a hardcoded
+path to the ``/lib64/ld-linux-x86-64.so.2`` dynamic loader. This loader is the
+program that does the actual finding of dynamic dependencies at runtime, and in
+an Ubuntu system its actual location is at ``/usr/lib64/ld-linux-x86-64.so.2``.
+So if the ``/lib64 -> usr/lib64`` symlink is broken because the prime directory
+contains ``lib64`` as a regular directory, then the vast majority of the
+binaries in the final rock's base system will simply fail to run because their
+loader is no longer available at ``/lib64/ld-linux-x86-64.so.2``.
+
+To fix this, Rockcraft will take the base system into account when creating the
```
lengau, line 182:
```suggestion
To fix this, Rockcraft takes the base system into account when creating the
```

docs/explanation/lifecycle-layer.rst
```diff
@@ -0,0 +1,209 @@
+
+From prime step to OCI layer
+============================
+
+Rockcraft is a tool that creates OCI images using the same concepts and
+mechanisms that create snaps and charms: the lifecycle language from
+Craft Parts. There is a significant difference between the way the Craft
```
lengau, line 7: This might be an appropriate place to cross-link to [the craft-parts documentation](https://canonical-craft-parts.readthedocs-hosted.com/en/latest/).

docs/explanation/lifecycle-layer.rst
```diff
@@ -0,0 +1,209 @@
...
+.. figure:: /_static/dive-efficiency.png
+   :width: 75%
+   :align: center
+   :alt: Dive reporting an inefficient image
+
+What ``dive`` tells us is that about ``60 MB`` worth of files are *duplicated*
+between the base Ubuntu 22.04 layer and the "primed" layer: for example, the
+file ``/usr/lib/x86_64-linux-gnu/libcrypto.so.3`` exists both in the base layer
+(as part of the base Ubuntu system) and in the primed layer (pulled in by
+belonging to a package that is an indirect dependency of ``python3-minimal``).
```
lengau, line 79: I'm a big fan of this paragraph from an accessibility standpoint, since the image isn't useful to someone using a screen reader.

lengau (approved):
This is a great explanation of what's going on and a good way for someone who's familiar with snapcraft to understand how rockcraft utilises the same core concepts.

docs/explanation/lifecycle-layer.rst
cjdcordeiro: But I think that's the problem and that's why we're discussing the `usrmerge` in the other thread below. I.e. this page addresses multiple aspects of the OCI layering process in Rockcraft. In reality, if someone reads this page, they will "Learn how Rockcraft builds an OCI layer from 'parts'".

Not trying to be a buzzkill, but I'd actually consider another approach: _we should have an explanation heading for how the rocks' layers come to be_. In that case, this page would feel natural, to me, as a subpage under the "OCI Layering" explanation heading. One more argument in favour of such a re-org is the level of knowledge that is being "explained" here. If you compare this with the other existing explanations, this is much more of a tech-dive than the others (which are more product-oriented).

docs/explanation/lifecycle-layer.rst
cjdcordeiro: See my comment above in https://github.com/canonical/rockcraft/pull/445#discussion_r1423913874.

In short, putting this side by side with the other existing explanations just feels very unnatural. 

I think this split will become more natural as we move the docs into the already agreed split "Consumer" vs "Developer". But for now, having `usrmerge` explanations at the same level as "What is a rock?" feels like a fruit salad :p

docs/explanation/lifecycle-layer.rst
cjdcordeiro: @sergiusens the comments are not about the pertinence of this explanation, but rather its place. Again referring to https://github.com/canonical/rockcraft/pull/445#discussion_r1423913874, we are missing an entry-level explanation to OCI Layering for rocks (which btw, is a common question). Then, yes, this page, would become an in-depth developer-oriented explanation of what happens when priming contents to an OCI layer.

cjdcordeiro (commented):
Left a few comments to the open discussions. But I don't want to block this from being merged so I won't request more changes in this doc. 

I do feel nonetheless that this explanation should not live at the same level, and within the same context, as the other existing ones. 

But I'm happy to address that later if you agree.

tigarmo:
For what it's worth I would be fine with moving this doc to a new "Internals" section under Explanations

cjdcordeiro (approved):
approving to allow progress. will address the re-org later

PR #446, opened 2023-12-11
chore(linter): fix some format errors

syu-w:


docs/how-to/code/create-slice/openssl.yaml
```diff
@@ -2,10 +2,10 @@ package: openssl
 slices:
     bins:
         essential:
```
tigarmo, line 4: how come it fixed the `essential` list but not the `contents` dict?

tests/spread/general/plugin-python-3.6/rockcraft.orig.yaml
```diff
@@ -9,9 +9,9 @@ platforms:
     amd64:
 
 package-repositories:
-  - type: apt
-    ppa: deadsnakes/ppa
-    priority: always
+    - type: apt
```
tigarmo, line 12: weird, why 4 spaces here but 2 everywhere else?

syu-w: It allows 2 or 4 spaces, but not allow mixed them in one file. I guess I can change it back to 2 spaces

syu-w: I also found there are other 4 spaces yaml files, do you want unify their format?

docs/how-to/code/create-slice/openssl.yaml
tigarmo: well yes, if we're going to lint this we should be consistent right? Otherwise I don't see the point

tests/spread/general/plugin-python-3.6/rockcraft.orig.yaml
tigarmo: I think if we're going to enable yamllint we should pick a single format, otherwise it's just churn for no good reason imo
I think 2-spaces is the more common standard (used in e.g. github workflows)

syu-w: Fixed to use 2 spaces only

tigarmo:
image

syu-w:
I mean this one is no-squash, and pure spaces changes... And you can always use that compare button...

lengau (approved):

PR #449, opened 2023-12-14
chore(chisel): update Chisel to v0.8.1

cjdcordeiro:

-----

In preparation for a coming backward incompatible Chisel release, this PR is updating the Chisel version to a patch release that will allow for a longer grace migration period.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -117,12 +117,6 @@ parts:
         - pkg-config
 
   chisel:
-    plugin: go
-    source: https://github.com/canonical/chisel.git
-    source-tag: v0.8.1
-    build-environment:
-      - CGO: "0"
-      - CGO_ENABLED: "0"
-      - GOBIN: "${CRAFT_PART_INSTALL}/libexec/rockcraft"
-    build-snaps:
-      - go/1.20/stable
+    plugin: nil
+    stage-snaps:
+      - chisel/latest/candidate
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -120,5 +120,7 @@ parts:
     plugin: nil
     stage-snaps:
       - chisel/latest/candidate
+    organize: 
+      bin/chisel: libexec/rockcraft/chisel
     stage:
-      - bin/chisel
+      - libexec/rockcraft/chisel
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -120,3 +120,5 @@ parts:
     plugin: nil
     stage-snaps:
       - chisel/latest/candidate
+    stage:
+      - bin/chisel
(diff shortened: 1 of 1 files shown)


cjdcordeiro:
@tigarmo have you seen this snapcraft error? I can't reproduce it locally

tigarmo:
I think there's some non-determinism here because both the `skopeo` and `chisel` parts declare different channels for the `go` build-snap (we probably want to detect that somehow @cmatsuoka)
So what's happening there is that `chisel` is being built with go 1.17

cjdcordeiro:
Alright, despite the glitch, this should work around it

snap/snapcraft.yaml
```diff
@@ -117,12 +117,10 @@ parts:
...
-    source-commit: bd27f8700cd7d2a6b4e0df6b10c3761c83a70485
-    build-environment:
-      - CGO: "0"
-      - CGO_ENABLED: "0"
-      - GOBIN: "${CRAFT_PART_INSTALL}/libexec/rockcraft"
-    build-snaps:
-      - go/1.18/stable
+    plugin: nil
+    stage-snaps:
+      - chisel/latest/candidate
```
tigarmo, line 122: isn't `latest/candidate` a moving target? Might be safer to rollback to the previous code and just use go 1.20 for the skopeo snap too

cjdcordeiro: in this case, it isn't safer. We'll need to rethink how we use Chisel.

We also do this for pebble (installing from `latest/stable`).

Bottom-line, in a couple of weeks there will be a new chisel version that will make all Rockcraft release <1.1.0 incompatible with Chisel :s

yes, I know, and there's nothing we can do.

So for now, and until we figure out a way of making Rockcraft critically independent from its dependencies, it is simply easier to always install the latest releases. We have tests for it anyway, so it's actually safer than what we do with pebble.

cjdcordeiro: Given the way Chisel works, no. That's actually what we want. I've actually been thinking about getting rid of versions for chisel, and opting instead for a `latest` release only. It's too radical, but that's the safest behavior.

Your concern is valid, but its consequences would be immediately reflected by either a build or tests failure.

tigarmo (approved):

snap/snapcraft.yaml
cmatsuoka, line 122: I'm usually not a big fan of stage snaps because in the general case they need a matching base, but go binaries are more resilient.

cmatsuoka (approved):

snap/snapcraft.yaml
cjdcordeiro: this shall be going away anyway - see https://warthogs.atlassian.net/browse/ROCKS-930

PR #455, opened 2024-01-12
add script to generate Schemastore's schema for rockcraft.yaml

linostar:

-----

Script for auto-generation of a Schemastore's schema for rockcraft.yaml


linostar:
@cjdcordeiro Ready for review.

linostar:
snap-tests are failing due to a chisel error that seems unrelated to this PR changes:

``` /rockcraft/docs/how-to/code/create-slice
+ mkdir out
+ ./chisel/chisel cut --release ./chisel-releases --root ./out openssl_bins
2024/01/13 14:35:59 Processing ./chisel-releases release...
error: chisel.yaml: cannot parse release definition: yaml: unmarshal errors:
  line 8: field v1-public-keys not found in type setup.yamlArchive
  line 10: field v1-public-keys not found in type setup.yamlRelease```

**Comment by linostar** [comment]
GH job passed: https://github.com/linostar/rockcraft/actions/runs/7530385975/job/20496635425

**Comment by tigarmo** [comment]
@linostar the spread tests are failing because the branch on your repo is too old in relation to `main`, which has the fix. Can you rebase it?

**Comment by linostar** [comment]
@tigarmo I rebased from upstream's main but doesn't seem to solve the isuue of failing spread test.

**Comment by cjdcordeiro** [comment]
Here's the fix for this problem: https://github.com/canonical/rockcraft/pull/456

File: .github/workflows/schema.yaml [hunk 1]
```diff
@@ -0,0 +1,26 @@
...
+          python-version: '3.10'
+
+      - name: Install requirements
+        run: pip install -r requirements.txt
+
+      - name: Run generator script
+        run: python .github/scripts/schema.py > generated_schema.json
+
+      - name: Check with stored schema
+        run: diff generated_schema.json schemas/rockcraft_schema.json > /dev/null
```
cjdcordeiro, line 26: Why `/dev/null`? It would be useful to know the mismatch when an error occurs...

schemas/rockcraft_schema.json
```diff
@@ -0,0 +1,647 @@
+{
+  "title": "Project",
```
cjdcordeiro, line 2: Where does this value come from? Sounds a bit generic...

.github/workflows/schema.yaml
cjdcordeiro, line 26: Just being pedantic, but this should only run when the rockcraft_schema.json file already exists, otherwise, it will throw a false positive

schemas/rockcraft_schema.json
cjdcordeiro, line 1: for consistency, you might want to follow the name naming convention as snapcraft -> https://github.com/snapcore/snapcraft/tree/main/schema

So it would be: `schema/rockcraft.json`

cjdcordeiro (changes requested):
Thanks! This is looking good, just some minor comments.

.github/scripts/schema.py
```diff
@@ -0,0 +1,60 @@
...
+
+from craft_parts import Part
+
+script_dir = os.path.dirname(os.path.realpath(__file__))
+sys.path.append(os.path.join(script_dir, "../../"))
+
+from rockcraft.models.project import Project  # noqa: E402
+
+if TYPE_CHECKING:
+    import argparse
```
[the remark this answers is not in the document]
linostar: It's not. It was a leftover.

.github/workflows/schema.yaml
tigarmo: but after this PR this file will always exist right?

.github/workflows/schema.yaml
```diff
@@ -20,7 +20,10 @@ jobs:
...
       - name: Run generator script
-        run: python .github/scripts/schema.py > generated_schema.json
+        run: python tools/schema/schema.py > generated_schema.json
 
       - name: Check with stored schema
-        run: diff generated_schema.json schemas/rockcraft_schema.json > /dev/null
+        run: |
+          if [[ -f schemas/rockcraft.json ]]; then
+            diff generated_schema.json schemas/rockcraft.json
+          fi
```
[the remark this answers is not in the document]
linostar: @cjdcordeiro  Using `find` here can be considered overkill, and would also fail its purpose if somewhere in the future we have another rockcraft.json in another dir for some reason. I think a more proper way is to fail if there is no `rockcraft.json` in the assigned folder, either by adding an `else` statement to the current command, or by simplifying the whole `if` block to something like:
```test -f schema/rockcraft.json && diff generated_schema.json schemas/rockcraft.json```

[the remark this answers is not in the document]
cjdcordeiro: y the `test` line works fine.

schema/rockcraft.json
cjdcordeiro, line 1: This is missing `extensions`

cjdcordeiro (commented):
LGTM but just noticed the same problem as on the schema shared in the rocks channel -> `extensions`

cjdcordeiro (approved):
LGTM. @tigarmo can you confirm that we shouldn't see (yet) the `extensions` field in the generated schema? I thought we would after #454 , but...

schema/rockcraft.json
cjdcordeiro: Ignoring this as discussed in a separate meeting.
 - we should care about it once https://github.com/canonical/rockcraft/tree/feature/12f is merged

linostar:
Waiting for @tigarmo approval on the other changes.

tigarmo (approved):
nice! Thanks a lot

tigarmo:
> LGTM. @tigarmo can you confirm that we shouldn't see (yet) the `extensions` field in the generated schema? I thought we would after #454 , but...

The `extensions` field isn't there because it's not part of the `Project` model (it's removed from the yaml during preprocessing)

PR #458, opened 2024-01-22
feat: support ubuntu@24.04 devel rocks

tigarmo:
This PR is a prototype/wip implementation of the same support for `devel` bases that Snapcraft has. The gist of it is: to build a noble-based rock, while noble is still in development, one can do:

```
base: ubuntu@24.04
build-base: devel
```

The PR is still in draft and has some failing integration tests but I'm putting it up for two reasons:

1) To add some spread tests to see whether we're able to build noble rocks at all (we are)
2) To have a place to discuss/argue so we can decide on the path forward and support 24.04 sooner rather than later.

So, have at it!

### DIFF
--- a/docs/reference/rockcraft.yaml.rst
+++ b/docs/reference/rockcraft.yaml.rst
@@ -70,7 +70,7 @@ The rock version, used to tag the OCI image and name the rock file.
 ``base``
 --------
 
-**Type**: One of ``ubuntu@20.04 | ubuntu@22.04 | bare``
+**Type**: One of ``ubuntu@20.04 | ubuntu@22.04 | ubuntu@24.04 | bare``
 
 **Required**: Yes
 
@@ -84,10 +84,14 @@ which is typically used with static binaries or
    The notation "ubuntu:<channel>" is also supported for some channels, but this
    format is deprecated and should be avoided.
 
+.. note::
+   Base ``ubuntu@24.04`` is still unstable and under active development. To use
+   it, ``build-base`` *must* be ``devel``.
+
 ``build-base``
 --------------
 
-**Type**: One of ``ubuntu@20.04 | ubuntu@22.04``
+**Type**: One of ``ubuntu@20.04 | ubuntu@22.04 | devel``
 
 **Required**: Yes, if ``base`` is ``bare``
 
@@ -100,6 +104,11 @@ the value of ``base``.
 .. note::
    The notation "ubuntu:<channel>" is also supported for some channels, but this
    format is deprecated and should be avoided.
+
+.. note::
+   ``devel`` is a "special" value that means "the next Ubuntu version, currently
+   in development". This means that the contents of this system changes
+   frequently and should not be relied on for production rocks.
 
 ``license``
 -----------
--- a/rockcraft/models/project.py
+++ b/rockcraft/models/project.py
@@ -117,6 +117,14 @@ INVALID_NAME_MESSAGE = (
 
 DEPRECATED_COLON_BASES = ["ubuntu:20.04", "ubuntu:22.04"]
 
+CURRENT_DEVEL_BASE = "ubuntu@24.04"
+
+DEVEL_BASE_WARNING = (
+    "The development build-base should only be used for testing purposes, "
+    "as its contents are bound to change with the opening of new Ubuntu releases, "
+    "suddenly and without warning."
+)
+
 
 class NameStr(pydantic.ConstrainedStr):
     """Constrained string type only accepting valid rock names."""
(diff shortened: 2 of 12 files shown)


tests/spread/general/base-devel/rockcraft.orig.yaml
```diff
@@ -0,0 +1,18 @@
+name: base-devel
+base: placeholder-base
```
tigarmo, line 2: this gets replaced with both `ubuntu@24.04` and `bare`

rockcraft/models/project.py
```diff
@@ -133,8 +133,8 @@ class Project(YamlModelMixin, BaseProject):
     description: str  # type: ignore[reportIncompatibleVariableOverride]
     rock_license: str = pydantic.Field(alias="license")
     platforms: dict[str, Any]
-    base: Literal["bare", "ubuntu@20.04", "ubuntu@22.04"]
-    build_base: Literal["ubuntu@20.04", "ubuntu@22.04"] | None
+    base: Literal["bare", "ubuntu@20.04", "ubuntu@22.04", "ubuntu@24.04"]
+    build_base: Literal["ubuntu@20.04", "ubuntu@22.04", "devel"] | None
```
cjdcordeiro, line 137: The docs should be updated too, including an explanation for this new cool `devel` option :)

tests/spread/general/base-devel/rockcraft.orig.yaml
cjdcordeiro, line 1: nitpick: this yaml has 4-spaces indentation, while others use 2-spaces

cjdcordeiro (changes requested):
Noice!! Just a couple of minor comments

rockcraft/models/project.py
tigarmo: oh nice
if you like the look of it then I'll work on getting this PR actually ready for merging. I'll ping you when it's done, thanks!

tigarmo:
@cjdcordeiro I think it's ready for re-review

docs/reference/rockcraft.yaml.rst
```diff
@@ -84,10 +84,14 @@ which is typically used with static binaries or
...
 
+.. note::
+   Base ``ubuntu@24.04`` is still unstable and under active development. To use
+   it, ``build-base`` *must* be ``devel``.
+
 ``build-base``
 --------------
 
-**Type**: One of ``ubuntu@20.04 | ubuntu@22.04``
+**Type**: One of ``ubuntu@20.04 | ubuntu@22.04 | devel``
```
cjdcordeiro, line 94: I think we've discussed it before, but my concern remains here: even though `devel` is quite explicit and resulting rocks should be marked as so, what will happen is that, especially for continuous pipelines, developers will need to manually change this parameter whenever 24.04 is released. 

That time window is quite narrow.

Between 24.04 release and the developer changing this recipe, rocks might be built and released, and since we don't support package pinning, this can result in a catastrophic integrity problem. Eg.: the rock `python:3.11-24.04_edge` might be shipped with `python3.12` once 24.04 is released, cause `devel` will suddenly become 24.10.

Am I looking at this wrong?

tigarmo: Ah I see what you mean. You are right: let's say that we release Rockcraft 1.2.0 with this change. In May, an existing rockcraft.yaml project with `build-base: devel` will suddenly start using 24.10 base even though the version of Rockcraft itself hasn't changed.

The solution might be to [do what Snapcraft did](https://github.com/snapcore/snapcraft/pull/4532) and specifically use the daily Noble images for `devel`, instead of the latest "rolling" stream. Then when we decide to start supporting 24.10 as "devel" we'll have to explicitly release a new Rockcraft version with the change.

tigarmo: note: The reason that I didn't adopt that Snapcraft change in this PR is that apparently there is/was a problem with the remote containing the daily images. Looks like that's been/will soon be solved, though

cjdcordeiro: Yes I think that approach would be slightly better, even though people tend to use `latest` version in their ci pipelines. Nonetheless, the use of latest is always an argument we can fight against, so ...y, you solution is an improvement

sergiusens: This should not be a problem, anyone tracking devel should be in that spirit, tracking the latest development.

Once 24.04 is release, devel indeed goes to 24.10, because it is devel, the timewindow is short now due to the nature of what we are thinking about, but if you change that to think about 26.04, the scale does make more sense.

tigarmo: Note: our agreement here during last week's sync was to follow Snapcraft's notion of `devel`, in that there is no expectation of stability because even for a specific target (like noble) things change very frequently up until stabilization. I'll add Snapcraft's warning message as agreed

cjdcordeiro: agreed. all good on my side.

rockcraft/models/project.py
```diff
@@ -119,6 +119,12 @@ def _validate_platform_set(cls, values: Mapping[str, Any]) -> Mapping[str, Any]:
 
 CURRENT_DEVEL_BASE = "ubuntu@24.04"
 
+DEVEL_BASE_WARNING = (
+    "Running rockcraft with a development base is for testing "
+    "purposes only, as the contents of the base can change suddenly "
+    "and without warning."
```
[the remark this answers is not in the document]
tigarmo: done!!

cjdcordeiro (approved):
the test is taking forever but it lgtm if it passes

sergiusens (approved):

PR #460, opened 2024-01-24
Move Permissions definition in rockcraft schema to top scope

linostar:

-----

This is to fix an issue in Rockcraft's schema.json, because `Permissions` needs to moved to the top scope `definitions` instead of being located inside `Part`'s definitions.

### DIFF
--- a/schema/rockcraft.json
+++ b/schema/rockcraft.json
@@ -617,31 +617,29 @@
           }
         }
       },
-      "additionalProperties": false,
-      "definitions": {
-        "Permissions": {
-          "title": "Permissions",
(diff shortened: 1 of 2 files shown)


linostar:
@cjdcordeiro @tigarmo This should be ready for review.

cjdcordeiro (approved):
I can confirm it works.

For future reference, and possibly automation, is there an easy way to test this?

linostar:
> I can confirm it works.
> 
> For future reference, and possibly automation, is there an easy way to test this?

We can use this github action to validate our schema:
https://github.com/OrRosenblatt/validate-json-action

cjdcordeiro:
Do you have an example of it passing & failing with this "Permissions definition" issue?

linostar:
> Do you have an example of it passing & failing with this "Permissions definition" issue?

I have an example of it failing where it should succeed :-)
https://github.com/linostar/rockcraft/actions/runs/7638164257/job/20808486483

I'm not sure but I think it doesn't support the 2020-12 schema draft. I'm experimenting with alternatives now.

tigarmo (approved):
thanks!
@cjdcordeiro @linostar do we want to do anything else on this PR before merging it?

linostar:
> thanks! @cjdcordeiro @linostar do we want to do anything else on this PR before merging it?

This should be good to merge, because it provides the needed fix. After discussing with Cris, I'll create another PR that provides automated validation of the generated rockcraft schema.

PR #471, opened 2024-01-29
Allow additionalProperties under "parts/plugin" in rockcraft schema

linostar:

-----

## Description
Currently, Rockcraft schema doesn't allow additional fields under `plugin`, a field under `parts`. There are different values that can be used for `plugin`, and for each value, there is a different set of plugin properties that can be used under that plugin. The current schema generator, using pydantic's schema function, can't crawl the subclassed PluginProperties models in order to include their respective plugin properties and whitelist them.

With the current `additionalProperties` setting of the `plugin` field set to `false`, Rockcraft schema may consider a `rockcraft.yaml` invalid when its syntax is actually correct. In order to avoid that, we are going to whitelist all additional properties under `plugin`, even if it means whitelisting incorrect properties too, letting rockcraft be the final judge of them.


linostar:
@cjdcordeiro @tigarmo Ready for review.

cjdcordeiro (approved):

tools/schema/schema.py
```diff
@@ -62,6 +62,9 @@ def generate_project_schema() -> str:
         "$ref": "#/definitions/Part"
     }
     project_schema["definitions"]["Part"] = part_schema
+    # allow "plugin" under "parts" to have extra fields such as "python-packages"
```
tigarmo, line 65: isn't this overly permissive now? It means that typos like `stage-packagse` won't be caught be the schema anymore. That sounds worse than getting validation errors for the (comparatively few) plugin-specific parameters.
There is a way, on the Rockcraft-side, to crawl the available plugins and see each of their possible properties. Is there a way to express this on the schema? Like, if `plugin` is `python` then the set of valid properties for the part is extended with `python-requirements`, etc

linostar: > isn't this overly permissive now? It means that typos like `stage-packagse` won't be caught be the schema anymore. That sounds worse than getting validation errors for the (comparatively few) plugin-specific parameters. There is a way, on the Rockcraft-side, to crawl the available plugins and see each of their possible properties. Is there a way to express this on the schema? Like, if `plugin` is `python` then the set of valid properties for the part is extended with `python-requirements`, etc

We can probably enumerate all plugin properties and make our schema whitelist them only, but there doesn't seem a way to limit each a subset of those plugin properties to one plugin value. In that case, the schema would consider `python-packages' field valid for `plugin: npm`.

linostar: Actually, it turns out we can. After diving deeper into jsonschema docs, I found that it is possible to achieve by using [if-then-else](https://json-schema.org/understanding-json-schema/reference/conditionals#ifthenelse).

cjdcordeiro: > That sounds worse than getting validation errors for the (comparatively few) plugin-specific parameters.

I'd say the opposite. There are 2 possible scenarios:
 1. `additionalProperties: false` - users get "False Positives" in their schema validation, misleading them to think their project is wrong and thus spend time trying to fix it, when in fact everything is ok.
 2. `additionalProperties: true` - users get no feedback, a false sense of security, and will eventually build a project that might contain errors. **However**, Rockcraft does project validation within, so whether it is through the JSON schema or the Rockcraft Pydantic validation, the user will always get feedback. 

Given that we are introducing the schema validation, the most logical is to be conservative (as always), and tackle it iteratively - i.e. _validate what we can_. Option 1. is dangerous IMO, as it provides a wrong message, while 2. is simply, inaccurate.

Having said this, if there's a way to add those fields to the schema, I'd obviously prefer that :)

tigarmo: I think we should iterate on adding those fields to the schema, but I'd like to chip in that in my opinion not all `part` properties are created equal: many, possibly most parts use either the `nil` or `dump` plugins which have no extra properties. So for those cases allowing additional properties has only downsides and no false positives

linostar: I added plugin property names to the scheam using schema conditionals in the latest commits.

linostar:
A couple manual tests to validate the latest changes to the schema:

Invalid plugin name:
image

nil plugin with valid and invalid plugin properties:
image

python plugin with valid and invalid plugin properties:
image

npm plugin with valid and invalid plugin properties:
image

schema/rockcraft.json
```diff
@@ -619,7 +635,3251 @@
...
+        {
+          "if": {
+            "properties": {
+              "plugin": {
+                "const": "make"
+              }
+            }
+          },
+          "then": {
+            "$comment": "common properties had to be repeated here or they would be considered invalid by the schema validator otherwise",
```
cjdcordeiro, line 2036: what's the impact of the $comment field? This is more like an internal comment

linostar: Yeah, it is an internal comment that has no impact on the schema. You asked me whether we can add comments to explain the repetitive part if anyone was to read the schema and wonder.

cjdcordeiro: Yeah I know, but I was referring to an actual `# comment`. But if you say this has no impact when using the schema, then it's alright

cjdcordeiro (approved):
Thanks for the screenshots. Seems to be working well, good job. 

As for the repetition, since it's dynamically generated, I can live with that until a better solution arises.

FYI (@tigarmo ) the failing tests are related to the installation of chisel slices in 24.04. We should probably discuss this outside this PR, because the 24.04 chisel release is something which is super prone to disruptive upstream changes (like `libc6` which has changed over the weekend). So if testing slices on an `devel` build base, we should probably be careful not to rely on too many slices.

schema/rockcraft.json
linostar: Yeah, this is the official way for commenting in json schemas:
https://json-schema.org/understanding-json-schema/reference/comments

tigarmo (approved):
Ah this looks great! Thanks a lot

tigarmo:
@linostar can you rebase/re-generate the schema?

linostar:
> @linostar can you rebase/re-generate the schema?

Done.

PR #473, opened 2024-01-31
fix(pebble): /var/lib/pebble/default has mode 0777

### LINKED ISSUE #461: `run-user` does not work properly without `services`
### Bug Description

I am building a rock with the `run-user: _daemon_` specified and without any services. The build is successful, but when I try to run the rock with docker, there is no response in the console. It's not like the terminal is freezing, there are simply no pebble logs or anything.

But if I have a service under services (it was commented out in the rockcraft.yaml), the image is working as expected.

### To Reproduce

1. Build the rock

```
rockcraft pack -v
```

2. Load it into docker-daemon.

```
skopeo copy oci-archive:hello_latest_amd64.rock docker-daemon:hello:latest
```

3. Run with docker

```
docker run hello
```

At this point, there is no response on the console. No logs.

### Environment

rockcraft version: 1.1.1
Using LXD
OS: Ubuntu 22.04 LTS

### rockcraft.yaml

```shell
name: hello
title: Hello World
summary: An Hello World rock
description: |
  This is just an example of a Rockcraft project
  for a Hello World rock.
version: latest
license: Apache-2.0

base: ubuntu@22.04

run-user: _daemon_

# services:
#   hello:
#     override: replace
#     startup: enabled
#     command: /usr/bin/hello -t

platforms:
  amd64:

parts:
  hello:
    plugin: nil
    stage-packages:
      - hello
```

### Relevant log output

```shell
https://pastebin.com/vvYdvrAA
```

rebornplusplus:
Fixes #461.

This PR ensures that the `PEBBLE` default directory `/var/lib/pebble/default` always has the mode 0777.

Previously, the mode of the `PEBBLE` dir was only set to 0777 if there were services or checks specified in the `rockcraft.yaml` file. Otherwise, it used to be 0755 with user=`root` group=`root`. This meant that if Pebble was being run as some user other than root, using the `run-user` directive or some other way, the Pebble run would fail. It would fail because Pebble could not create the necessary files (socket) inside that directory. This bug is reported in issue https://github.com/canonical/rockcraft/issues/461.

-----

### DIFF
--- a/rockcraft/pebble.py
+++ b/rockcraft/pebble.py
@@ -172,7 +172,11 @@ class Pebble:
         "stage-snaps": ["pebble/latest/stable"],
         "stage": [PEBBLE_BINARY_PATH],
         # We need this because "services" is Optional, but the directory must exist
-        "override-prime": f"craftctl default\nmkdir -p {PEBBLE_LAYERS_PATH}",
+        "override-prime": str(
+            "craftctl default\n"
+            f"mkdir -p {PEBBLE_LAYERS_PATH}\n"
+            f"chmod 777 {PEBBLE_PATH}"
+        ),
     }
 
     def define_pebble_layer(
--- a/tests/spread/general/run-user-minimal/rockcraft.yaml
+++ b/tests/spread/general/run-user-minimal/rockcraft.yaml
@@ -1,0 +1,16 @@
+name: run-user-minimal-test
+version: latest
+summary: A minimal rock with a non-root user
+description: A minimal rock that has a non-root default user
+license: GPL-3.0
+
+base: ubuntu@22.04
+
+platforms:
+    amd64:
+
+run-user: _daemon_
+
+parts:
+    my-part:
+        plugin: nil
--- a/tests/spread/general/run-user-minimal/task.yaml
+++ b/tests/spread/general/run-user-minimal/task.yaml
@@ -1,0 +1,41 @@
+summary: non-root run-user minimal test
+
+execute: |
+  run_rockcraft pack
+
+  test -f run-user-minimal-test_latest_amd64.rock
+  test ! -d work
+
+  # Ensure docker does not have this container image
+  docker rmi --force run-user-minimal-test
+  # Install container
+  sudo /snap/rockcraft/current/bin/skopeo --insecure-policy copy \
+    oci-archive:run-user-minimal-test_latest_amd64.rock \
+    docker-daemon:run-user-minimal-test:latest
+  # Ensure container exists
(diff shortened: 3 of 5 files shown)


cjdcordeiro (approved):
Nice thanks.

I just have a comment about the new spread test - couldn't we just re-use the existing `run-user` test? The task.yaml is pretty much the same, with the exception of `  timeout 15 docker run --rm --name run-user-test-container run-user-test --args test 10`, which TBH, isn't too relevant for the `run-user` test. I don't think adding more tests would slow down Spread, in theory. But I have the impression the Spread tests have been getting slower and slower, so I'm just being cautions. @tigarmo what's your perception on the scalability of the spread tests?

rebornplusplus:
> Couldn't we just re-use the existing `run-user` test? The task.yaml is pretty much the same, with the exception of ` timeout 15 docker run --rm --name run-user-test-container run-user-test --args test 10`, which TBH, isn't too relevant for the `run-user` test.

The rockcraft.yaml in that test contains `services` which changes the pebble dir permission anyway. I wanted to test without any `services` or `checks` i.e. without any additional pebble layers.

cjdcordeiro:
> > Couldn't we just re-use the existing `run-user` test? The task.yaml is pretty much the same, with the exception of ` timeout 15 docker run --rm --name run-user-test-container run-user-test --args test 10`, which TBH, isn't too relevant for the `run-user` test.
> 
> The rockcraft.yaml in that test contains `services` which changes the pebble dir permission anyway. I wanted to test without any `services` or `checks` i.e. without any additional pebble layers.

y true, and I'm wondering if the existing `run-user` test wouldn't be equally useful if we just removed `services` from it - cause after all, there's only 1 line in `task.yaml` testing that service

tigarmo:
I'm fine with either option; generally a new spread test doesn't take too much extra time:
```

cjdcordeiro:
> I'm fine with either option; generally a new spread test doesn't take too much extra time. This one took about a minute:
> 
> ```
> 2024-02-01T02:29:13.2405237Z 2024-02-01 02:29:13 Executing google:ubuntu-22.04-64:tests/spread/general/run-user-minimal (feb010222-852385) (12/115)...
> 2024-02-01T02:30:15.9623158Z 2024-02-01 02:30:15 Restoring google:ubuntu-22.04-64:tests/spread/general/run-user-minimal (feb010222-852385)...
> ``
> ```

Talked with @sergiusens a while ago and the slowness might be related to the number of tests vs the amount of available resources for the spread tests. I think we can tackle this later in a different PR, as maybe the solution is to simply vertically or horizontally scale the spread backend

tigarmo (approved):
Thanks!

PR #477, opened 2024-02-08
build(deps): update skopeo to 1.14.2

tigarmo:
Version 1.14.2 fixes the issue with Docker Daemon API mismatch when converting the rock to a Docker image. The previous version (1.9.0) is also over a year and a half old so it's a good idea to update anyway.

-----

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -89,8 +89,9 @@ parts:
         mkdir "$CRAFT_PART_INSTALL"/bin
         install -m755 umoci.static "$CRAFT_PART_INSTALL"/bin/umoci
     build-packages:
-        - golang-go
         - make
+    build-snaps:
+        - go/1.21/stable
 
   skopeo:
     plugin: nil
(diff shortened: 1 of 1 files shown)


cjdcordeiro (commented):
I can confirm it works with Go 1.21 (at least the `skopeo` part).

tigarmo:
I confirmed locally that at least one of the problems that I started having with the old skopeo is fixed. Specifically this one, when using skopeo to copy the rock to docker:
```
Using skopeo version 1.9.0
Getting image source signatures
Copying blob 6b0d87ff5f38 [--------------------------------------] 320.0KiB / 28.3MiB
Copying blob 56f4888b032b [--------------------------------------] 8.0b / 138.0MiB
Copying blob 0cf9559d33c2 [--------------------------------------] 8.0b / 853.0b
Copying blob 77f5eaaa6010 [>-------------------------------------] 8.0b / 439.0b
FATA[0000] writing blob: io: read/write on closed pipe
```

cmatsuoka (approved):
Ship it!

cjdcordeiro (approved):
:rocket:

PR #482, opened 2024-02-13
ci: always discard spread workers

lengau:

-----

A direct copy of: https://github.com/snapcore/snapcraft/blob/main/.github/workflows/spread.yml#L117

### DIFF
--- a/.github/workflows/spread-large.yaml
+++ b/.github/workflows/spread-large.yaml
@@ -45,3 +45,11 @@ jobs:
           path: tests
       - name: Run spread (large)
         run: spread tests/spread/large/
+        
+      - name: Discard spread workers
+        if: always()
+        run: |
+          shopt -s nullglob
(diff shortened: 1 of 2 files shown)


mattculler (commented):
Seems to make sense to me!  So you're telling spread to discard the reuse workers once it finishes.  Will spread delete the .spread-reuse-$PID.yaml files itself, or do you need to/should you delete the file in your "run" snippet after discarding the worker?

lengau:
@mattculler Exactly - this is ensuring that the workers always get discarded. An example use case where they previously weren't is if someone cancelled a workflow.

The `-discard` option should delete the reuse files, but the whole directory gets removed during the runner cleanup so it's not actually a huge deal if it doesn't.

mattculler (approved):
Looks good!

tigarmo (approved):

PR #492, opened 2024-02-26
Add `django-framework` extension and `django-framework` init profile

weiiwang01:
Add the `django-framework` that can package a Django project into a ROCK, similar to the `flask-framework`.
Some refactoring was performed, and some improvements were made over the `flask-framework` extension as well.

Changes include:

- Update the extension part name prefix from `flask/` to `flask-framework/`.
- Install `statsd-exporter` along with the Gunicorn server.
- Install `ca-certificates` in the ROCK image by default.
- Disable the WSGI path check when `services.flask.command` or `services.django.command` is overridden.
- Improve the quality of generated parts.

-----


rockcraft/extensions/gunicorn.py
```diff
@@ -0,0 +1,340 @@
...
+    def gen_install_app_part(self) -> Dict[str, Any]:
+        """Return the prime list for the Flask project."""
+        if "django-framework/install-app" not in self.yaml_data.get("parts", {}):
+            return {
+                "plugin": "dump",
+                "source": ".",
+                "organize": {self.name: "django/app"},
+                "stage": ["django/app"],
+                "prime": ["django/app"],
+            }
```
weiiwang01, line 315: I am not very confident about this. Is there a better way to achieve the same effect? I tried this, but it doesn't work.

```yaml
django-framework/install-app:
  plugin: dump
  source: "{self.name}"
  organize: 
    ".": django/app
```

cjdcordeiro (changes requested):

docs/how-to/use-django-extension.rst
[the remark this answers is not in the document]
weiiwang01: Updated, thanks!

docs/how-to/use-django-extension.rst
```diff
@@ -0,0 +1,69 @@
+Using the django-framework extension
+------------------------------------
+
+The Flask extension is compatible with the ``bare`` and ``ubuntu@22.04`` bases.
```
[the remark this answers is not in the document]
weiiwang01: Added, thanks!

docs/how-to/use-django-extension.rst
```diff
@@ -0,0 +1,69 @@
...
+To employ it, include ``extensions: [ django-framework ]`` in your
+``rockcraft.yaml`` file.
+
+Example:
+
+.. code-block:: yaml
```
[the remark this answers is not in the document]
weiiwang01: Gotcha, I have replaced all example blocks with real examples and added some tests for them.

docs/how-to/use-flask-extension.rst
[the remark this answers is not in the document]
weiiwang01: Yes, I have searched and replaced all instances of "ROCK"

rockcraft/extensions/gunicorn.py
```diff
@@ -0,0 +1,340 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright 2023 Canonical Ltd.
```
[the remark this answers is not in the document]
weiiwang01: Updated, thanks!

rockcraft/extensions/gunicorn.py
```diff
@@ -0,0 +1,340 @@
...
+
+
+class _GunicornBase(Extension):
+    """An extension base class for Python WSGI framework extensions."""
+
+    @staticmethod
+    @override
+    def get_supported_bases() -> Tuple[str, ...]:
+        """Return supported bases."""
+        return "bare", "ubuntu@20.04", "ubuntu@22.04", "ubuntu:20.04", "ubuntu:22.04"
```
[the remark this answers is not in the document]
weiiwang01: No, I forgot to update here, updated, thanks!

tigarmo (commented):
Should we have a spread test that actually stands up a Django app and interacts with it in some manner?

rockcraft/commands/init.py
```diff
@@ -118,6 +118,47 @@ class InitCommand(AppCommand):
                 #       - flask/app/static
                 """
         ),
+        "django-framework": textwrap.dedent(
+            """\
+                name: {name}
+                base: ubuntu@22.04 # the base environment for this Django application
+                version: '0.1' # just for humans. Semantic versioning is recommended
+                summary: A summary of your Flask application # 79 char long summary
```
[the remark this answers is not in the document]
weiiwang01: Updated, thanks!

rockcraft/extensions/gunicorn.py
weiiwang01: Does the wildcard patten here matches files start with dot `.`?

tigarmo: hmm I'm not sure. If it doesn't, then maybe this will:

```
  organize:
    "*": django/app
    ".*": django/app

**Comment by tigarmo** [comment]
> Should we have a spread test that actually stands up a Django app and interacts with it in some manner?

bump @weiiwang01

**Comment by weiiwang01** [comment]
> > Should we have a spread test that actually stands up a Django app and interacts with it in some manner?
> 
> bump @weiiwang01

Sorry, I missed this comment. I've added a new spread test for `django-extension` similar to the one for `flask-extension`.

File: rockcraft/extensions/gunicorn.py (hunk 1 above)
Comment (weiiwang01) [reply rockcraft/extensions/gunicorn.py]: I did some experiment and found that `organize` indeed uses `glob` rules (which don't match files starting with `.`) instead of `fnmatch` rules. Here is the final working version (the target directory needs to end with `/`):

```yaml
  organize:
    "*": django/app/
    ".*": django/app/
```

Thank you for your suggestion! updated.

tigarmo:
@weiiwang01 as we discussed I moved the `feature/12f` branch to the current `main`. Feel free to close this PR and open a new one, or update this one, it's up to you. Thanks

weiiwang01:
> @weiiwang01 as we discussed I moved the `feature/12f` branch to the current `main`. Feel free to close this PR and open a new one, or update this one, it's up to you. Thanks

Thanks! since there isn't too much conflict, I've updated this PR and cleaned up the commits.

tests/spread/__init__.py
tigarmo, line 1: why are these `__init__.py` files here? These aren't Python packages

tests/spread/general/extension-django/task.yaml
```diff
@@ -0,0 +1,34 @@
+summary: django extension test
```
tigarmo, line 1: this is very nice 👍🏾

tests/spread/__init__.py
weiiwang01: It must be automatically created by the IDE, I have removed those files, thanks!

tigarmo (approved):

docs/how-to/code/use-django-extension/example/example_django/example_django/asgi.py
```diff
@@ -0,0 +1,16 @@
...
+
+For more information on this file, see
+https://docs.djangoproject.com/en/5.0/howto/deployment/asgi/
+"""
+
+import os
+
+from django.core.asgi import get_asgi_application
+
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", "example_django.settings")
```
lengau, line 14: Is this the preferred way to do this? It seems to me like putting this inside the rock's environment would be cleaner (though IDK what downsides it would have).

docs/how-to/code/use-django-extension/task.yaml
```diff
@@ -0,0 +1,21 @@
...
+
+execute: |
+  cd example
+  rockcraft pack --verbosity debug
+  cd ../override_example
+  rockcraft pack --verbosity debug
+  cd ..
+
+restore: |
+  rm -rf */*.rock
```
lengau, line 21:
```suggestion
environment:
  ROCKCRAFT_ENABLE_EXPERIMENTAL_EXTENSIONS: "1"
  EXAMPLE_DIR/example: example
  EXAMPLE_DIR/override_example: override_example

execute: |
  cd $EXAMPLE_DIR
  rockcraft pack --verbosity debug

restore: |
  cd $EXAMPLE_DIR
  rm *.rock
  rockcraft clean
```

This uses [variants](https://github.com/snapcore/spread?tab=readme-ov-file#variants) to break the test up. That way if `example` starts failing, it'll independently check whether `override_example` still succeeds.

tests/spread/general/extension-django/task.yaml
```diff
@@ -0,0 +1,34 @@
...
+
+execute: |
+  export ROCKCRAFT_ENABLE_EXPERIMENTAL_EXTENSIONS=true
```
lengau, line 4:
```suggestion

environment:
  ROCKCRAFT_ENABLE_EXPERIMENTAL_EXTENSIONS: "true"

execute: |
```

lengau (approved):
Looks good overall. I made some minor suggestions, but they're definitely not blockers for this PR.

docs/how-to/code/use-django-extension/example/example_django/example_django/asgi.py
weiiwang01: Yes, all Django projects in the tests are generated using the Django admin tool, so I think it's okay.

docs/how-to/code/use-django-extension/task.yaml
weiiwang01: Learned, thanks!

tigarmo:
Spread failures are unrelated (I'm working on them)

PR #519, opened 2024-04-01
spread: reorganize spread tests

tigarmo:
Reorganize the spread tests to make better use of CI resources (Google VMs). The majority of the tests work the same regardless of host system, so it's a waste to run them on jammy _and_ focal _and_ fedora. Instead, bump the number of jammy machines and have them run the majority of the tests (renamed from `general` to `rockcraft`), and add a new suite `foreign` where we can add tests that are explicitly designed to test cross-system support.

-----


mattculler (approved):

spread.yaml
```diff
@@ -122,11 +124,19 @@ suites:
...
+    - ubuntu-22.04-64
+
+  tests/spread/rockcraft/:
+    summary: tests for rockcraft core functionality, independent of host system
+    systems:
+    - ubuntu-22.04-64
 
-  tests/spread/general/:
-    summary: tests for rockcraft core functionality
+  tests/spread/all-systems/:
```
sergiusens, line 135: all systems being the suite and multiple being in the summary seems contradictory, let's rename this to the agreed "rockcraft" or "rockcraft-itself" (this is what I was referring to fwiw, the vanity name for the launchpad project on launchpad https://bugs.launchpad.net/launchpad)

tigarmo: Can you elaborate on why you think `itself` is appropriate? From what I understand, it makes sense in launchpad's case because it is literally using launchpad to host launchpad. In rockcraft's case there is no "inception" going on - there are tests for rockcraft functionality, which should be independent of system, and tests that we explicitly want to run on multiple hosts, largely to run the generated rock.
I renamed the existing "general" suite to "rockcraft" for the tests that are host-independent. Would "multiple-systems" be better than "all-systems"?

sergiusens: maybe call it alternative-systems or foreign-systems, or just foreign

sergiusens: the rockcraft itself is mostly perspective, does rockcraf work on fedora is a quesion about rockcraft itself working on fedora and not so much about testing the features of rockcraft

tigarmo: I'm fine with `foreign` (updated)

sergiusens (approved):
:partying_face:

tigarmo:
@linostar fyi

tigarmo:
@sergiusens can you please squash-merge this? The spread failures are due to the 24.04 archives

PR #520, opened 2024-04-02
fix(spread): more robust check of non-pruned files

tigarmo:
The spread test for the "layer pruning" feature worked by comparing the stats of two files if they exist in both layers. These stats - filesize, ownership and permission bits - were enough up until now because the non-pruned files differed on at least the filesize.

The recent release of libuuid1 broke this expectation: this new package has a ``libuuid.so.1.3.0`` with identical file size as the previous version, but different contents. So the pruning correctly kept the file, but the test failed.

Update the test to _also_ check the files' contents if necessary.

-----

### DIFF
--- a/tests/spread/general/prune/check_layers.py
+++ b/tests/spread/general/prune/check_layers.py
@@ -1,17 +1,24 @@
+import filecmp
 import json
+import shutil
 import subprocess
 import sys
 import tarfile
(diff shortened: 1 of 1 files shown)


mr-cal (approved):

cmatsuoka (approved):
LGTM!

PR #521, opened 2024-04-02
refactor: port code to latest craft-app update

tigarmo:
The brunt of the update is updating usages of platform/build-for to the build plan. This largely affects the image and package services, but the semantics should be unchanged because it's the same data just delivered differently.

A bit more context here is that the `feature/craft-application` branch already had a bunch of migration work done, so I just rebased that branch and created this one off it. That branch bumps craft-application to 2.3.0.

-----

### DIFF
--- a/docs/how-to/code/use-flask-extension/example/rockcraft.yaml
+++ b/docs/how-to/code/use-flask-extension/example/rockcraft.yaml
@@ -3,7 +3,11 @@ summary: A Flask application
 description: A rock packing a Flask application via the flask extension
 version: "0.1"
 base: bare
+build-base: ubuntu@22.04
 license: Apache-2.0
 
 extensions:
   - flask-framework
+
+platforms:
+  amd64:
--- a/rockcraft/services/lifecycle.py
+++ b/rockcraft/services/lifecycle.py
@@ -27,6 +27,7 @@ from rockcraft import layers
 from rockcraft.models.project import Project
 
 # Enable the craft-parts features that we use
+Features.reset()
 Features(enable_overlay=True)
 
 
--- a/rockcraft/services/package.py
+++ b/rockcraft/services/package.py
@@ -21,11 +21,12 @@ import pathlib
 import typing
 from typing import cast
 
-from craft_application import AppMetadata, PackageService, models, util
+from craft_application import AppMetadata, PackageService, errors, models
+from craft_application.models import BuildInfo
 from craft_cli import emit
 from overrides import override  # type: ignore[reportUnknownVariableType]
 
-from rockcraft import errors, oci
+from rockcraft import oci
 from rockcraft.models import Project
 from rockcraft.usernames import SUPPORTED_GLOBAL_USERNAMES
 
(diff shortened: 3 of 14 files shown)


docs/how-to/code/use-flask-extension/example/rockcraft.yaml
```diff
@@ -3,7 +3,11 @@ summary: A Flask application
 description: A rock packing a Flask application via the flask extension
 version: "0.1"
 base: bare
+build-base: ubuntu@22.04
```
tigarmo, line 6: @weiiwang01 @gregory-schiano heads-up that we have a problem here: the context is in the [commit message](https://github.com/canonical/rockcraft/pull/521/commits/534bc41f4550ad628600b0bc4d51034b498dcdad). We should talk soon but after this is merged whatever projects are using the flask extension without a build-base (and a bare base) and/or no explicit platforms are going to break.

rockcraft/services/lifecycle.py
```diff
@@ -27,6 +27,7 @@
 from rockcraft.models.project import Project
 
 # Enable the craft-parts features that we use
+Features.reset()
```
lengau, line 30: I'm curious what drove this change. Nothing wrong with it, but I'm wondering if you uncovered a bug in craft-application?

rockcraft/services/package.py
```diff
@@ -64,35 +63,22 @@ def pack(self, prime_dir: pathlib.Path, dest: pathlib.Path) -> list[pathlib.Path
         image_service = services.image
         image_info = image_service.obtain_image()
 
-        platform = self._platform
+        if not self._build_plan:
```
lengau, line 66: Several apps are doing this now. Do you think the upstream PackageService should take a build plan and provide a `_build_info` property that does these checks?

lengau (approved):
LGTM!

rockcraft/services/package.py
tigarmo: I definitely agree that we should upstream this somehow, but in my case the ImageService _also_ needs these checks (and it happens even earlier than the LifecycleService so it's arguable the most important)
Maybe we should expose a "validate_single_build_plan()" function that does these checks and returns a single BuildInfo

docs/how-to/code/use-flask-extension/example/rockcraft.yaml
tigarmo: cc @jdkandersson

rockcraft/services/package.py
lengau: Feature request made: https://github.com/canonical/craft-application/issues/297

We can figure out details over there; no need to bog down this PR for that.

rockcraft/services/lifecycle.py
tigarmo: There is a bug but I think only happens in tests. The issue is that a fixture is enabling the overlay feature, and then _later_ this module gets imported and tries to re-enable it. The singleton then complains that the `Features` class has already been instantiated. We can fix that in a number of ways but I think it's low-ish priority now

docs/how-to/code/use-flask-extension/example/rockcraft.yaml
jdkandersson: @tigarmo thanks for the note, I think it is fine. I propose we just get the extension to raise an error if the build-base/ platforms is missing. Any thoughts @gregory-schiano?

tigarmo: @jdkandersson I see that you've already put up a PR for this, which is fine, but it's possibly redundant because a project that lacks build-base or platforms is going to fail validation before getting to the extensions now (this was the motivation for this original change that I had to do here). Still, no harm in doing two checks I suppose

PR #523, opened 2024-04-04
Add jammy and x64 labels for self hosted runners

codethulu:

-----


lengau:
Thanks for this @codethulu! Our needs are slightly different here, so I've created #524 as a superseding PR to get our stuff properly tagged.

PR #524, opened 2024-04-04
ci: adjust tags for spread runs

lengau:
This ensures we can use any machine with the spread-installed tag. Necessary for upcoming changes to Canonical's self-hosted runners.

Spread will fail until a repo admin adjusts the runner tags.

-----


tigarmo (approved):
thanksss

SamirPS (approved):

PR #527, opened 2024-04-05
Merge craft-application update into main

tigarmo:
The update to the latest craft-application update was done over time in this `feature/craft-application` branch, which is now ready to be integrated back into main. Only the latest commit needs to be reviewed because we might as well update to the very new `2.4.0` - all other commits have already been reviewed in various past PRs.

### DIFF
--- a/docs/how-to/code/use-flask-extension/example/rockcraft.yaml
+++ b/docs/how-to/code/use-flask-extension/example/rockcraft.yaml
@@ -3,7 +3,11 @@ summary: A Flask application
 description: A rock packing a Flask application via the flask extension
 version: "0.1"
 base: bare
+build-base: ubuntu@22.04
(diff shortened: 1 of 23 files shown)


cmatsuoka:
> Only the latest commit needs to be reviewed

Is this the one that updates craft-application to 2.4.0?

tigarmo:
> > Only the latest commit needs to be reviewed
> 
> Is this the one that updates craft-application to 2.4.0?

yes

cmatsuoka (approved):
LGTM. Also moving build-base settings from the extension to user if the extension is set and base is bare sounds like a sensible decision in the long term.

lengau (approved):

lengau:
Rebased on main, should be ready to rebase-merge when CI completes

PR #528, opened 2024-04-08
fix(pebble): place pebble in a separate location

tigarmo:
The context here is that starting from Ubuntu 24.04, the 'base-files' package (and related chisel slices) provides "bin" as a symlink to "usr/bin". This breaks the previous "phantom" pebble part because it created "bin" as a regular directory which then conflicts with the symlink.

Moving forward, the pebble binary is now placed in ".rock/bin/". This way we won't get further collisions and this reflects the fact that the location of the binary is, and should be seen, as an implementation detail.

-----


tigarmo:
Spread failures are unrelated (the new test is passing)

tigarmo:
@linostar please take a look too (can't add you as a reviewer apparently)

rockcraft/pebble.py
```diff
@@ -166,18 +166,28 @@ class Pebble:
 
     PEBBLE_PATH = "var/lib/pebble/default"
     PEBBLE_LAYERS_PATH = f"{PEBBLE_PATH}/layers"
-    PEBBLE_BINARY_PATH = "bin/pebble"
-    PEBBLE_PART_SPEC = {
+    PEBBLE_BINARY_DIR = ".rock/bin"
```
cjdcordeiro, line 169: isn't `.rock` globally defined? it should be cause we're putting the control data in there. Hardcoding it here will make it harder in the future if we decide to rename it

rockcraft/pebble.py
```diff
@@ -229,3 +239,34 @@ def define_pebble_layer(
...
+        entrypoint = [f"/{pebble_path}", "enter"]
+
+        if is_legacy:
+            entrypoint += ["--verbose"]
+
+        return entrypoint
+
+    @staticmethod
+    def _is_focal_or_jammy(build_base: str) -> bool:
+        return build_base in ("ubuntu@20.04", "ubuntu@22.04")
```
cjdcordeiro, line 272: is the build_base already transformed to use `@` during the pydantic validation? Cause if not, and someone still uses the `:` notation, this condition won't work

tests/spread/rockcraft/chisel-dotnet/rockcraft.yaml
cjdcordeiro, line 1: why are you adding this spread test in this PR? I.e. it is useful, but the PR's change is mainly targeting the Pebble location for 24.04+

rockcraft/pebble.py
```diff
@@ -229,3 +239,34 @@ def define_pebble_layer(
             )
 
         tmp_new_layer.chmod(0o777)
+
+    @staticmethod
+    def get_part_spec(build_base: str) -> dict[str, Any]:
+        """Get the partpart that providers the pebble binary for a given build base."""
```
cmatsuoka, line 245:
```suggestion
        """Get the part providing the pebble binary for a given build base."""
```

tests/spread/rockcraft/chisel-dotnet/rockcraft.yaml
cjdcordeiro: is it just cause you've made the previous .net spread test into a generic one? If so...I don't think we need a .net-specific test, unless there's something special about it

doesn't harm though. so it's up to you...just checking

cjdcordeiro (changes requested):
very clean! well done.

just a couple of nits

cmatsuoka (approved):
Looks great!

rockcraft/pebble.py
tigarmo: yes, we [replace ":" with "@"](https://github.com/canonical/rockcraft/blob/main/rockcraft/models/project.py#L192) during validation

tests/spread/rockcraft/chisel-dotnet/rockcraft.yaml
tigarmo: yeap I moved/renamed it because I wanted to add a new one, but I agree that we can remove it. I'll do it.

rockcraft/models/project.py
```diff
@@ -551,4 +551,7 @@ def _add_pebble_data(yaml_data: dict[str, Any]) -> None:
         # Project already has a pebble part: this is not supported.
         raise CraftValidationError('Cannot override the default "pebble" part')
 
-    parts["pebble"] = Pebble.PEBBLE_PART_SPEC
+    model = BuildPlanner.unmarshal(yaml_data)
+    build_base = model.build_base if model.build_base else model.base
```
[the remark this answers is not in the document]
tigarmo: As an aside outside of this PR, I often think that `build_base` should just be a `str` and not `str | None`

tests/spread/rockcraft/chisel-dotnet/rockcraft.yaml
tigarmo: done

rockcraft/pebble.py
tigarmo: That's a good idea - added a new `constants` module to hold this and future constants

tigarmo:
Spread failures are expected

cjdcordeiro (approved):
nice! lgtm thanks

linostar (commented):
LGTM. I have but a small suggestion, which is to add the new path `.rocks/bin` to $PATH, so that if someone is running `docker ... exec pebble <command>` they wouldn't have to guess and supply the path to pebble there.

tigarmo:
> LGTM. I have but a small suggestion, which is to add the new path `.rocks/bin` to $PATH, so that if someone is running `docker ... exec pebble <command>` they wouldn't have to guess and supply the path to pebble there.

@cjdcordeiro do we want to support this case?

cjdcordeiro:
> > LGTM. I have but a small suggestion, which is to add the new path `.rocks/bin` to $PATH, so that if someone is running `docker ... exec pebble <command>` they wouldn't have to guess and supply the path to pebble there.
> 
> @cjdcordeiro do we want to support this case?

I think we need that. @linostar can you confirm that with this PR, `docker exec <rock> pebble services` doesn't work? If so, then we urgently need the binary in $PATH, and we also need a spread test to cover these cases.

(FYI @sergiusens )

linostar:
> > > LGTM. I have but a small suggestion, which is to add the new path `.rocks/bin` to $PATH, so that if someone is running `docker ... exec pebble <command>` they wouldn't have to guess and supply the path to pebble there.
> > 
> > 
> > @cjdcordeiro do we want to support this case?
> 
> I think we need that. @linostar can you confirm that with this PR, `docker exec <rock> pebble services` doesn't work? If so, then we urgently need the binary in $PATH, and we also need a spread test to cover these cases.
> 
> (FYI @sergiusens )

@cjdcordeiro Confirmed:
image

cjdcordeiro:
Thanks. See https://warthogs.atlassian.net/browse/ROCKS-1088

PR #529, opened 2024-04-09
fix(flask): don't enforce definition of base, build-base and platforms

jdkandersson:

-----

Makes it an error for the any gunicorn based extension not to define a `base`, `build-base` and `platforms` key

### DIFF
--- a/rockcraft/commands/init.py
+++ b/rockcraft/commands/init.py
@@ -80,7 +80,6 @@ class InitCommand(AppCommand):
             """\
                 name: {name}
                 base: ubuntu@22.04 # the base environment for this Flask application
-                build-base: ubuntu@22.04 # the build base environment for creating the OCI image
                 version: '0.1' # just for humans. Semantic versioning is recommended
                 summary: A summary of your Flask application # 79 char long summary
                 description: |
--- a/rockcraft/commands/init.py
+++ b/rockcraft/commands/init.py
@@ -80,6 +80,7 @@ class InitCommand(AppCommand):
             """\
                 name: {name}
                 base: ubuntu@22.04 # the base environment for this Flask application
+                build-base: ubuntu@22.04 # the build base environment for creating the OCI image
                 version: '0.1' # just for humans. Semantic versioning is recommended
                 summary: A summary of your Flask application # 79 char long summary
                 description: |
--- a/rockcraft/extensions/gunicorn.py
+++ b/rockcraft/extensions/gunicorn.py
@@ -117,6 +117,12 @@ class _GunicornBase(Extension):
           - services: a service to run the Gunicorn server
           - parts: see _GunicornBase._gen_parts
         """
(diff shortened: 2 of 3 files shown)


rockcraft/commands/init.py
```diff
@@ -80,6 +80,7 @@ class InitCommand(AppCommand):
             """\
                 name: {name}
                 base: ubuntu@22.04 # the base environment for this Flask application
+                build-base: ubuntu@22.04 # the build base environment for creating the OCI image
```
SamirPS, line 83: "The build base environment for this Flask Application" no?

rockcraft/extensions/gunicorn.py
```diff
@@ -117,6 +117,12 @@ def get_root_snippet(self) -> Dict[str, Any]:
           - services: a service to run the Gunicorn server
           - parts: see _GunicornBase._gen_parts
         """
+        for required_key in ("base", "build-base", "platforms"):
```
tigarmo, line 120: The rules for `build-base` are [a bit more complicated](https://github.com/canonical/rockcraft/blob/main/rockcraft/models/project.py#L170) than this: if `build-base` is missing, it's the value of `base`, *unless* `base` is `"bare"` (in which case this is an error).
I'm thinking now that we probably want to drop this validation here altogether, because these fields [are already validated](https://github.com/canonical/rockcraft/blob/main/rockcraft/models/project.py#L136) before getting here. The new flow in the application is this:
- A `BuildPlanner` is built from the "unprocessed" yaml data to determine the base and what to build (the platforms). This validates the base, build-base and platforms entries;
- The original yaml is processed by applying extensions, global environment, grammar, etc. This step is where this method would be called;
- The processed yaml is then used to create a regular `Project` model.

Another argument for not adding the validation here is that we'd have to keep the two validations "in sync" - I think it's better to enforce "by contract" that the yaml will always have the things the project needs.

tigarmo (changes requested):

rockcraft/extensions/gunicorn.py
jdkandersson: so it sounds like we just need to remove the logic that adds the fields

rockcraft/commands/init.py
jdkandersson: You are right, probably I'll remove this again since it is only necessary if we had a bare base here or wanted it to be different, neither of which is needed

jdkandersson: Marking as resolved for now 🙂

rockcraft/extensions/gunicorn.py
jdkandersson: ok that should be fixed now, I looked through the docs and couldn't find any reference to our treatment of `base`, `build-base` and `platforms`

tigarmo (approved):

PR #535, opened 2024-04-11
feat(ci): add pull request labeler for documentation labels

zhijie-yang:

-----

Resolves ROCKS-1050 Create a workflow that tags documentation PRs with the proper label.

- Added labeller workflow to automatically add "documentation" labels to pull requests if there are changes in `docs/**` in the repo.
- This has to be a blind merge according to the [labeller documentation](https://github.com/actions/labeler?tab=readme-ov-file#initial-set-up-of-the-labeler-action)
- This feature is verified in a [standalone GitHub repo](https://github.com/zhijie-yang/pr-labeller-test).


.github/labeler.yml
```diff
@@ -0,0 +1,4 @@
+# Add 'cocumentation' label to any changes within 'docs' folder or any subfolders
```
lengau, line 1:
```suggestion
# Add 'documentation' label to any changes within 'docs' folder or any subfolders
```

.github/workflows/labeler.yml
```diff
@@ -0,0 +1,12 @@
+name: Pull Request Labeler
+on:
+- pull_request_target
```
lengau, line 3: If I'm understanding what you want to do correctly, I think you either need:

```suggestion
  pull_request_target:
   branches:
     - main 
```

or perhaps a general `pull_request`?

[check the docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target) as I don't think this is a valid workflow

zhijie-yang: This is a valid workflow, see [actions/labeler documentation](https://github.com/actions/labeler?tab=readme-ov-file#create-workflow). I have no further doubts about this comment if you mean to have the `main` branch be the only target branch of a PR that receives a `documentation` label.

cjdcordeiro (commented):
Cool. Do you have an example of this working in your fork?

BTW, I agree with this only targeting `main` PRs.

Additionally, you can use the `paths` [field](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore) to restrict when this workflow is going to run (in fact, you could completely get rid of the `labeler.yml` cause we only have one label for now, so that's an overhead...but it could be useful for other auto-labels in the future, so I'm fine with keeping it)

zhijie-yang:
Test PR using my fork of this repo can be found [here](https://github.com/zhijie-yang/rockcraft/pull/2)

tigarmo (approved):

lengau (approved):

PR #567, opened 2024-05-10
refactor: use default BuildPlanner

mr-cal:

-----

This is a test PR for https://github.com/canonical/craft-application/pull/326, to see if the default BuildPlanner will work for rockcraft.

There are a few pre-requisite commits to be compatible with the latest craft-application.

I wasn't able to get pydantic to use [this](https://github.com/canonical/craft-application/pull/326/files#diff-d49e733a1e99a8cc3fe98cb7d956b735159c60342405692603fe5fcf8a970edaR111) upstream validator but everything else seems to be in working order.

(CRAFT-2824)

### DIFF
--- a/rockcraft/models/project.py
+++ b/rockcraft/models/project.py
@@ -15,6 +15,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 """Project definition and helpers."""
+import copy
 import re
 import shlex
 from collections.abc import Mapping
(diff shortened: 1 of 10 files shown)


rockcraft/models/project.py
```diff
@@ -281,6 +254,7 @@ class Project(YamlModelMixin, BuildPlanner, BaseProject):  # type: ignore[misc]
     services: dict[str, Service] | None
     checks: dict[str, Check] | None
     entrypoint_service: str | None
+    platforms: dict[str, Platform | None]  # type: ignore[assignment]
```
mr-cal, line 257: The generated schema is the same, with or without ` | None`.  Either way, this job fails:
https://github.com/canonical/rockcraft/actions/runs/9289669392/job/25564121553?pr=567#step:5:11

tigarmo: fixed the schema

tigarmo:
failures are unrelated (I guess `devel` is broken maybe, it's also failing on `main`)

tigarmo (approved):

mr-cal (commented):
I can't approve my own PR but your updates look good to me!

https://github.com/canonical/rockcraft/pull/567/commits/8c4b15dad0788c0e645531504a86a3df558232d9 is an almost empty commit

rockcraft/models/project.py
mr-cal: Ah, I wasn't aware you could manually tweak the schema

sergiusens (approved):

Issue #334, opened 2023-07-26
Declaring ca-certificates in stage-packages doesn't populate the /etc/ssl/certs directory


weiiwang01:
### Bug Description

When declaring `ca-certificates` in the part's `stage-packages`, the `/etc/ssl/certs` directory is not populated as expected. The `/etc/ssl/certs` should contain symbolic links pointing to the certificates found in `/usr/share/ca-certificates/mozilla/`. It's worth noting that certificates are present in `/usr/share/ca-certificates/mozilla/` within the Rock image.

### To Reproduce

Build the rock image based on the `rockcraft.yaml` file below, and then inspect the content of `/etc/ssl/certs` directory in the image. The directory is empty.
```bash
rockcraft pack
sudo /snap/rockcraft/current/bin/skopeo --insecure-policy copy oci-archive:ca-certs_0.1_amd64.rock docker-daemon:ca-certs:latest
docker run -it --rm --entrypoint /bin/ls ca-certs:latest -lah /etc/ssl/certs
```

### Environment

ubuntu 22.04 LTS, LXD

### rockcraft.yaml

```shell
# Copyright 2023 Canonical Ltd.
# See LICENSE file for licensing details.
name: ca-certs
summary: ca-certs issue.
description: ca-certs issue.
version: "0.1"
base: ubuntu:20.04
license: Apache-2.0
platforms:
  amd64:

parts:
  apt-dependencies:
    plugin: nil
    source: .
    stage-packages:
      - ca-certificates
```

### Relevant log output

```shell
N/A
```

gregory-schiano:
Side comment, the chisel slice `ca-certificates_data` should work since it's aggregating all the `/usr/share/ca-certificates/mozilla/` certificate into a file in the mutate part
See: https://github.com/canonical/chisel-releases/blob/ubuntu-22.04/slices/ca-certificates.yaml

So it's not relying on symlinks

cjdcordeiro:
have you tried using `overlay-packages`?

jnsgruk:
> > have you tried using `overlay-packages`?
> 
> `overlay-packages` works, thanks! Is there a general rule to decide which packages should go in the `staging-packages`, and which should go in the `overlay-packages`? Since other packages works in the `staging-packages`.

Yeh, this might be good to include in the docs - I can see more people tripping over this.

gregory-schiano:
But is it intended that symlinks aren't copied from `stage-packages` ?
[...]

PR #576, opened 2024-05-21
Overlay ca-certificates in django/flask extensions

javierdelapuente:

-----

As packages declared in `stage-packages` do no run maintainer scripts it is necessary to use the `overlay` step for the `ca-certificates` package to work correctly  (see https://github.com/canonical/rockcraft/issues/334).


sergiusens (changes requested):
These are chisel packages, which do run scripts, something must be off on the chisel side if this is not working as expected. @cjdcordeiro any thoughts?

javierdelapuente:
Closing, as adding an overlay will not allow the usage of `base: bare`, which is a limitation. A possible solution maybe related to the chiseled `ca-certificates`. See https://github.com/canonical/chisel-releases/issues/257.

PR #577, opened 2024-05-21
Delete pyproject.toml in override-build in flask/django

javierdelapuente:

-----

File `pyproject.toml` is not necessary for django/flask extensions, as there is no need to create a Python package.

Besides, for some project like NetBox, the `pyproject.toml` breaks the build.

### DIFF
--- a/rockcraft/extensions/gunicorn.py
+++ b/rockcraft/extensions/gunicorn.py
@@ -22,6 +22,7 @@ import os.path
 import pathlib
 import posixpath
 import re
+import textwrap
 from typing import Any, Dict, Tuple
 
 from overrides import override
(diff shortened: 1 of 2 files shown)


sergiusens (changes requested):
Shouldn't the project remove this instead? If it is even broken, then why have it at all?

javierdelapuente:
> Shouldn't the project remove this instead? If it is even broken, then why have it at all?

If the file is removed from the upstream project it would be easier, however there are other uses for this file. 

My use case for this case is NetBox. In this case this file ([pyproject.toml](https://github.com/netbox-community/netbox/blob/develop/pyproject.toml)) is there for linting purposes. However, the `python` plugin in `rockcraft` tries to install the project and it fails. My alternative is to include it in the `rockcraft.yaml` (https://github.com/canonical/netbox/blob/b412d045ebac54664f41fa643c65d4568aa60c31/rockcraft.yaml#L129).

jdkandersson:
@sergiusens There are several use cases for `pyproject.toml`, such as providing metadata for the project. In python, you can manage dependencies using `pyproject.toml`, but you don't have to. In the case of the flask extension, we ask the user to use `requirements.txt` instead. So in this change we automatically delete `pyproject.toml` because if the file is present, `rockcraft` will try to use it for installing dependencies rather than use `requirements.txt`. Does that make sense?

javierdelapuente:
Closed in favor of this issue: https://github.com/canonical/craft-parts/issues/747

PR #588, opened 2024-06-19
feat: make 'license' an optional field

cjdcordeiro:

-----

`license` becomes an optional field, and as a result, it no longer shows up in the templated `rockcraft.yaml` file, for any profile.

### DIFF
--- a/docs/reference/rockcraft.yaml.rst
+++ b/docs/reference/rockcraft.yaml.rst
@@ -115,7 +115,7 @@ the value of ``base``.
 
 **Type**: string, in `SPDX format <https://spdx.org/licenses/>`_
 
-**Required**: Yes
+**Required**: No
 
 The license of the software packaged inside the rock. This must match the SPDX
 format, but is case insensitive (e.g. both ``MIT`` and ``mit`` are valid).
--- a/docs/reference/rockcraft.yaml.rst
+++ b/docs/reference/rockcraft.yaml.rst
@@ -117,8 +117,9 @@ the value of ``base``.
 
 **Required**: No
 
-The license of the software packaged inside the rock. This must match the SPDX
-format, but is case insensitive (e.g. both ``MIT`` and ``mit`` are valid).
+The license of the software packaged inside the rock. This must either be
+"proprietary" or match the SPDX format. It is case insensitive (e.g. both
+``MIT`` and ``mit`` are valid).
 
 ``run-user``
 ------------
--- a/rockcraft/commands/init.py
+++ b/rockcraft/commands/init.py
@@ -67,7 +67,6 @@ class InitCommand(AppCommand):
                     most important story about it. Keep it under 100 words though,
                     we live in tweetspace and your description wants to look good in the
                     container registries out there.
-                license: GPL-3.0 # your application's SPDX license
                 platforms: # The platforms this rock should be built on and run on
                     amd64:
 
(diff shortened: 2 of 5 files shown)


rockcraft/models/project.py
```diff
@@ -331,18 +331,23 @@ def _check_unsupported_options(cls, values: Mapping[str, Any]) -> Mapping[str, A
...
+            return license
 
-        lic: spdx_lookup.License | None = spdx_lookup.by_id(rock_license)  # type: ignore[reportUnknownMemberType]
+        lic: spdx_lookup.License | None = spdx_lookup.by_id(license)  # type: ignore[reportUnknownMemberType]
         if lic is None:
+            docs_url = "https://documentation.ubuntu.com/rockcraft/en/stable/reference/rockcraft.yaml/#license"
             raise CraftValidationError(
-                f"License {rock_license} not valid. It must be valid and in SPDX format."
+                f"License {license} not valid. It must be valid and in SPDX format.\n"
+                f"See {docs_url}"
```
sergiusens, line 350: there should be a docs_url field here

cjdcordeiro: cool. I just realized that. amended

rockcraft/commands/init.py
```diff
@@ -129,7 +127,6 @@ class InitCommand(AppCommand):
                     most important story about it. Keep it under 100 words though,
                     we live in tweetspace and your description wants to look good in the
                     container registries out there.
-                license: GPL-3.0 # your application's SPDX license
```
linostar, line 132: Is removing the license field from the init template necessary? I think it could be better to leave it there (maybe with a comment explaining that it's optional?) so one doesn't forget it if they want to make their rock public.

tigarmo: @cjdcordeiro could/should we add the license as a commented-out line?

docs/reference/rockcraft.yaml.rst
```diff
@@ -115,7 +115,7 @@ the value of ``base``.
 
 **Type**: string, in `SPDX format <https://spdx.org/licenses/>`_
```
tigarmo, line 116: can you update this to say that the string `"proprietary"` is also accepted?

rockcraft/models/project.py
```diff
@@ -331,18 +331,24 @@ def _check_unsupported_options(cls, values: Mapping[str, Any]) -> Mapping[str, A
...
             # This is the license name we use on our stores.
-            return rock_license
+            return license
 
-        lic: spdx_lookup.License | None = spdx_lookup.by_id(rock_license)  # type: ignore[reportUnknownMemberType]
+        lic: spdx_lookup.License | None = spdx_lookup.by_id(license)  # type: ignore[reportUnknownMemberType]
         if lic is None:
             raise CraftValidationError(
-                f"License {rock_license} not valid. It must be valid and in SPDX format."
+                f"License {license} not valid. It must be valid and in SPDX format.",
```
tigarmo, line 350: It can also be `"proprietary"` apparently, the error message should point that out I believe

rockcraft/commands/init.py
cjdcordeiro: that was a request from above

docs/reference/rockcraft.yaml.rst
cjdcordeiro: sure, done

tigarmo (approved):

linostar (approved):
LGTM, thanks.

sergiusens (approved):

Issue #587, opened 2024-06-19
doc: wrong header level for flask/django extension guides


cjdcordeiro:
The `Managing project files with the flask extension` and `Managing project files with the Django extension` headers should be at a lower level. Atm, they appear as top-level in the left menu sidebar, when in fact they are part of another guide.

---
Document: how-to/rocks/use-flask-extension.rst

cjdcordeiro:
fyi @gregory-schiano

PR #594, opened 2024-06-20
Flask docs

jdkandersson:

-----
Fixes https://github.com/canonical/rockcraft/issues/587

Improvements to the documentation for Flask support


docs/tutorials/getting-started-with-flask.rst
```diff
@@ -0,0 +1,114 @@
+Getting started with Rockcraft for Flask
+----------------------------------------
+
+In this tutorial you will learn how to create a Flask application, build a rock
+with the Flask application and run it.
+
+Prerequisites
+-------------
+- snap enabled system (https://snapcraft.io)
+- LXD installed (https://linuxcontainers.org/lxd/getting-started-cli/)
```
javierdelapuente, line 10: LXD is not longer under linuxcontainers.org

jdkandersson:
- Add more information to the reference docs about the extension
- Move most of the content out of how to except for prime and deb into the tutorial

jdkandersson:
For the rock size, talk about the rock on disk and include a note that says the size on docker is the uncompressed size

docs/tutorials/getting-started-with-flask.rst
```diff
@@ -0,0 +1,114 @@
...
+    :dedent: 2
+
+The ``curl`` command should print ``Hello, world!``.
+
+Install Rockcraft
+-----------------
+
+Install Rockcraft on your host:
+
+.. literalinclude:: code/hello-world/task.yaml
```
erinecon, line 51: Should this line read `.. literalinclude:: code/getting-started-with-flask/task.yaml`?

jdkandersson:
Add an explanation of project details and what we are trying to accomplish - nice to have

docs/tutorials/code/getting-started-with-flask/task.yaml
```diff
@@ -0,0 +1,57 @@
...
+  # [docs:create-venv]
+  sudo apt-get update && sudo apt-get install python3-venv -y
+  python3 -m venv .venv
+  source .venv/bin/activate
+  pip install -r requirements.txt
+  # [docs:create-venv-end]
+
+  # [docs:run-flask]
+  flask run &
+  retry -n 5 --wait 2 curl localhost:5000
```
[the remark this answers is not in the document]
cjdcordeiro: +1. Or maybe skip `retry` as it makes for a clunkier snippet. You can separate the snippets, paste the expected output, and then show the `curl` snippet. By then flask should already be up, and if not, you can have a note (or even a section at the end) for troubleshooting

docs/tutorials/getting-started-with-flask.rst
```diff
@@ -0,0 +1,114 @@
...
+
+.. literalinclude:: code/getting-started-with-flask/task.yaml
+    :language: bash
+    :start-after: [docs:run-flask]
+    :end-before: [docs:run-flask-end]
+    :dedent: 2
+
+The ``curl`` command should print ``Hello, world!``.
+
+Install Rockcraft
```
cjdcordeiro, line 46: some of the prerequisites are on top, and they are not needed for the previous section. Then there's also this requisite. The approach should be: either all in under a single step, or each one explained only when needed

docs/tutorials/getting-started-with-flask.rst
```diff
@@ -0,0 +1,114 @@
...
+
+Use the ``flask-framework`` profile to create ``rockcraft.yaml``:
+
+.. literalinclude:: code/getting-started-with-flask/task.yaml
+    :language: bash
+    :start-after: [docs:create-rockcraft-yaml]
+    :end-before: [docs:create-rockcraft-yaml-end]
+    :dedent: 2
+
+Open ``rockcraft.yaml`` in a text editor and customise the ``name``,
```
cjdcordeiro, line 68: I think these steps should be dumbed down a bit more. e.g. the `sed` above isn't as important as the `rockcraft init`. and it's only being done for the sake of this tutorial.

docs/tutorials/getting-started-with-flask.rst
```diff
@@ -0,0 +1,114 @@
...
+``summary``, ``description`` and ``license``. Ensure that ``platforms`` includes
+the architecture of your host. For example, if your host uses the ARM
+architecture, include ``arm64`` in ``platforms``. Pack the rock:
+
+.. literalinclude:: code/getting-started-with-flask/task.yaml
```
cjdcordeiro, line 73: why the iptables command? that shouldn't happen with the `docker` snap. And if it does, this is one more remark that belongs to a troubleshooting note/section

docs/how-to/rocks/index.rst
```diff
@@ -8,6 +8,5 @@
    Publish a rock to a registry <publish-a-rock>
    Migrate a Docker image to a chiselled rock <migrate-to-chiselled-rock>
    Use the flask extension <use-flask-extension>
-   Use the django extension <use-django-extension>
```
linostar, line 11: Why is django guide removed?

docs/tutorials/getting-started-with-flask.rst
```diff
@@ -0,0 +1,114 @@
...
+    :dedent: 2
+
+
+Run the rock in Docker
+----------------------
+
+First, import the rock into Docker:
+
+.. note::
+    If you changed the ``name`` in ``rockcraft.yaml`` or are not on an ``amd64``
```
cjdcordeiro, line 86: then it might be worth adding the STDOUT from the previous commands.

we use 
```
..  code-block:: text
    :class: log-snippets
```
for log snippets

docs/tutorials/getting-started-with-flask.rst
```diff
@@ -0,0 +1,114 @@
...
+
+.. literalinclude:: code/getting-started-with-flask/task.yaml
+    :language: bash
+    :start-after: [docs:skopeo-copy]
+    :end-before: [docs:skopeo-copy-end]
+    :dedent: 2
+
+Now run the rock:
+
+.. literalinclude:: code/getting-started-with-flask/task.yaml
```
cjdcordeiro, line 97: why 8000?

docs/tutorials/getting-started-with-flask.rst
```diff
@@ -0,0 +1,114 @@
...
+    :language: bash
+    :start-after: [docs:docker-run]
+    :end-before: [docs:docker-run-end]
+    :dedent: 2
+
+The ``curl`` command should print ``Hello, world!`` in response.
+
+Cleanup
```
cjdcordeiro, line 105: IMO I'd skip this

cjdcordeiro (changes requested):

docs/how-to/rocks/index.rst
cjdcordeiro: My request. These two extensions need a lot of work, so it's safer to remove this for now as it could lead people to the same ill-documented guides we're trying to fix for Flask.

The same work that is being done for flask must be done for Django, so as soon we we're done with this, a similar PR should be raised to revive the django docs

docs/tutorials/getting-started-with-flask.rst
cjdcordeiro: IMHO the prerequisites should be at the top, in a single steps, explained for what they are needed.

The reason for this is because readers might get confused with instructions like `lxd init --auto` and end up running that everytime they want to do a `rockcraft pack`

docs/tutorials/getting-started-with-flask.rst
cjdcordeiro: the fix is explained here: https://documentation.ubuntu.com/lxd/en/latest/howto/network_bridge_firewalld/#prevent-connectivity-issues-with-lxd-and-docker

I've also noticed that doing `lxd init` first and then install `snap install docker` might avoid the issue, so if you can reproduce that, I'd make sure that's the right order in the docs

docs/how-to/rocks/use-flask-extension.rst
```diff
@@ -1,25 +1,53 @@
 Use the flask-framework extension
 ---------------------------------
 
-.. note::
-    The Flask extension is compatible with the ``bare`` and ``ubuntu@22.04``
-    bases.
+The ``flask-framework`` extension requires a ``requirements.txt`` file and a
+``app.py`` file that exposes ``app`` as a WSGI app. For example, ``app.py``:
```
erinecon, line 5: Nit-pick: I think `app` should be explicitly defined. Maybe something like, "...that exposes the Flask application ``app`` as a WSGI app."? (Maybe `app` doesn't refer to a Flask application in this instance.)

docs/how-to/rocks/use-flask-extension.rst
```diff
@@ -28,15 +56,90 @@ be prefixed with ``flask/app``.
...
-.. literalinclude:: ../code/use-flask-extension/prime_exclude_example/rockcraft.yaml
+.. literalinclude:: ../code/use-flask-extension/prime-exclude-example/rockcraft.yaml
     :language: yaml
     :start-after: [docs:parts-start]
     :end-before: [docs:parts-end]
+
+Including additional debs in the OCI image
+------------------------------------------
+
+if your app requires debs, for example to connect to a database, add the
```
erinecon, line 75: Nit picks: Capitalise "if". I would also restructure slightly to say "If your app requires debs (for example, to connect to a database), add..."

docs/tutorials/getting-started-with-flask.rst
```diff
@@ -0,0 +1,114 @@
+Getting started with Rockcraft for Flask
```
erinecon, line 1: Would it be more appropriate to say, "Getting started with Rockcraft for Flask applications"? Or has "applications" been explicitly removed to reduce the title size?

docs/tutorials/getting-started-with-flask.rst
```diff
@@ -0,0 +1,114 @@
...
+----------------------------------------
+
+In this tutorial you will learn how to create a Flask application, build a rock
+with the Flask application and run it.
+
+Prerequisites
+-------------
+- snap enabled system (https://snapcraft.io)
```
erinecon, line 9: Nitpick: Capitalise "snap"

docs/tutorials/getting-started-with-flask.rst
```diff
@@ -0,0 +1,114 @@
...
+- LXD installed (https://linuxcontainers.org/lxd/getting-started-cli/)
+- Docker installed (https://snapcraft.io/docker)
+- a text editor
```
erinecon, line 12: Nitpick: Capitalise "a"

docs/tutorials/getting-started-with-flask.rst
```diff
@@ -0,0 +1,114 @@
...
+
+Create Flask application
+------------------------
+
+Create a new directory and write the following into a text editor and save it as
```
erinecon, line 17: Nitpick: Rewrite slightly to say "Create a new directory, write the following into a text editor and save it as..."

docs/tutorials/getting-started-with-flask.rst
```diff
@@ -0,0 +1,114 @@
...
+    :end-before: [docs:create-venv-end]
+    :dedent: 2
+
+In the same directory, write the following into a text editor and save it as
+``app.py``:
+
+.. literalinclude:: code/getting-started-with-flask/app.py
+    :language: python
+
+Run and test the flask application to verify everything is working:
```
erinecon, line 36: (Context: I don't have experience with Flask applications) 
When I tried the tutorial, the `flask run &` seemed to work fine, but the `retry` command did not work. 
First, I wasn't sure whether I needed to CTRL+C from the `flask run &` command before running the other two commands. 
Then, when trying to run `retry -n 5 --wait 2 curl localhost:5000` I got the error `invalid option 'n'` and, when I removed that flag, `unrecognized option '--wait'`

docs/tutorials/getting-started-with-flask.rst
```diff
@@ -0,0 +1,114 @@
...
+architecture, include ``arm64`` in ``platforms``. Pack the rock:
+
+.. literalinclude:: code/getting-started-with-flask/task.yaml
+    :language: bash
+    :start-after: [docs:pack]
+    :end-before: [docs:pack-end]
+    :dedent: 2
+
+
+Run the rock in Docker
```
erinecon, line 80: I had previously installed docker with `sudo apt install docker` but these commands didn't work unless I installed docker via `sudo snap install docker`

docs/tutorials/getting-started-with-flask.rst
erinecon: Personally I think having the cleanup section is a good thing. Less experienced users might not know to do this

docs/tutorials/getting-started-with-flask.rst
jdkandersson: Thats the port we use inside the container and it is the default for gunicorn. Would you suggest something else?

docs/tutorials/getting-started-with-flask.rst
jdkandersson: yeh I did the same an this wasn't needed, done

docs/tutorials/getting-started-with-flask.rst
jdkandersson: removed

docs/tutorials/getting-started-with-flask.rst
jdkandersson: Added

docs/tutorials/getting-started-with-flask.rst
jdkandersson: Moved to the top and ensured all pre-requesits are included with an explanation

docs/tutorials/code/getting-started-with-flask/task.yaml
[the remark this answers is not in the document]
jdkandersson: Removed

docs/tutorials/getting-started-with-flask.rst
jdkandersson: Fixed

docs/tutorials/getting-started-with-flask.rst
jdkandersson: Removed this anyway

docs/how-to/rocks/use-flask-extension.rst
jdkandersson: Added more explanation

docs/how-to/rocks/use-flask-extension.rst
```diff
@@ -1,25 +1,53 @@
...
+
+.. literalinclude:: ../code/use-flask-extension/example/app.py
+    :language: python
+
+And ``requirements.txt``:
+
+.. literalinclude:: ../code/use-flask-extension/example/requirements.txt
 
 To employ it, include ``extensions: [ flask-framework ]`` in your
 ``rockcraft.yaml`` file.
```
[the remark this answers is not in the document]
jdkandersson: Done

docs/how-to/rocks/use-flask-extension.rst
jdkandersson: Fixed

docs/tutorials/getting-started-with-flask.rst
jdkandersson: I'm happy to include it, but yeh it gets a bit long

docs/tutorials/getting-started-with-flask.rst
jdkandersson: Removed anyway

docs/tutorials/getting-started-with-flask.rst
jdkandersson: Removed anyway

docs/tutorials/getting-started-with-flask.rst
jdkandersson: Fixed

docs/tutorials/getting-started-with-flask.rst
jdkandersson: I removed the `retry`

jdkandersson: And added a note to explain what the `&` does

docs/tutorials/getting-started-with-flask.rst
jdkandersson: That should be fixed now - the instructions are to install docker with snap at the start

docs/tutorials/getting-started-with-flask.rst
javierdelapuente: I just tested this in a multipass VM. However, it is just a matter of restarting the VM, and the failure appears again...

docs/tutorials/getting-started-with-flask.rst
cjdcordeiro: it needs some form of consistency, as in the beginning the app will be available on 5000, and then on 8000, with no explanation why. So either make them both run on the same port, or explain why the port changes. Why not, at the beginning, do `flask run -p 8000 &`?

jdkandersson: ok, updated flask run command

docs/tutorials/getting-started-with-flask.rst
```diff
@@ -0,0 +1,454 @@
...
+containerise it in a rock, using Rockcraft's ``flask-framework``
+:ref:`extension <flask-framework-reference>`.
+
+Setup
+=====
+
+We recommend starting from a clean Ubuntu 22.04 installation. If you don't have
+one available, you can create one using Multipass_:
+
+.. collapse:: How to create an Ubuntu 22.04 VM with Multipass
```
javierdelapuente, line 14: I think the advantage of Multipass is that it can be installed in many platforms, like  Windows/mac/ other distribution. This is good, but clearly focused on another Ubuntu. Maybe at least a link/reference to the installation of Multipass?

docs/reference/extensions.rst
```diff
@@ -6,20 +6,82 @@ Rockcraft extensions are crafted to expand and modify the user-provided
...
+    The Flask extension is compatible with the ``bare`` and ``ubuntu@22.04``
+    bases.
+
+Project requirements
+====================
+
+There are 4 requirements to be able to use the ``flask-framework`` extension:
+
+1. There must be a ``requirements.txt`` file in the root of the project
+2. ``Flask`` must be declared as a dependency in the ``requirements.txt`` file
```
sergiusens, line 33: 1 and 2 are arguably 1 requirement

docs/reference/extensions.rst
```diff
@@ -6,20 +6,82 @@ Rockcraft extensions are crafted to expand and modify the user-provided
...
+3. There must be an ``app.py`` file in the root of the project
+4. The project must include a WSGI app with the path ``app:app`` (set the name
+   of the Flask object to ``app`` in ``app.py``)
```
sergiusens, line 36: same for this

docs/reference/extensions.rst
```diff
@@ -6,20 +6,82 @@ Rockcraft extensions are crafted to expand and modify the user-provided
 rockcraft project file, aiming to minimise the boilerplate code when
 initiating a new rock.
 
+.. _flask-framework-reference:
```
sergiusens, line 9: it might be better to make extensions a dir with an index and put flask into its own rst. This is cleaner when linking to documentation from tooling and concise in the sense that you can start reading from the top instead of visually searching for the anchor

docs/tutorials/getting-started-with-flask.rst
```diff
@@ -0,0 +1,454 @@
...
+As a final step, let's say you want to update your application. For example,
+you want to add a new ``/time`` endpoint which returns the current time.
+
+Start by opening the ``app.py`` file in a text editor and update the code to
+look like the following:
+
+.. literalinclude:: code/getting-started-with-flask/time-app.py
+    :language: python
+
+Since you are creating a new version of your application, open the
```
javierdelapuente, line 373: To add base and buid-base, a sed one liner is used. However to update the versions, the user is asked to do it. I think it should be the same for both cases (whatever is nicer).

docs/tutorials/getting-started-with-flask.rst
```diff
@@ -0,0 +1,454 @@
...
+        multipass shell flask-hello-world
+
+----
+
+LXD will be required for creating the OCI image. Make sure it is installed an
+initialised:
+
+.. code-block:: bash
+
+   sudo snap install lxd
```
javierdelapuente, line 50: At least in the multipass VM, it is there by default.
```
ubuntu@flask-hello-world:~$ sudo snap install lxd
snap "lxd" is already installed, see 'snap help refresh'
```

docs/tutorials/getting-started-with-flask.rst
cjdcordeiro: TBH I think this is ok like this. The reason is because the `sed` command is necessary because that change must be applied verbatim. All the other file editing changes are not as strict, i.e. if the user sets the version to 0.1.1, this will still work...we use 0.2. as an example. OTOH, the "base" modification **must** literally take those values.

docs/tutorials/getting-started-with-flask.rst
cjdcordeiro: that's alright. It's an info message. harmless. I think that's less error prone than adding text saying "if you're on VM .... then it might already be installed, but you still need to `init` it with ...etc."

docs/reference/extensions.rst
cjdcordeiro: makes sense. done

docs/reference/extensions.rst
cjdcordeiro: also changed

docs/reference/extensions.rst
cjdcordeiro: changed

docs/tutorials/getting-started-with-flask.rst
cjdcordeiro: Sure. Added near the multipass installation.

docs/tutorials/getting-started-with-flask.rst
```diff
@@ -0,0 +1,454 @@
...
+
+    Finally, once the VM is up, open a shell into it:
+
+    .. code-block:: bash
+
+        multipass shell flask-hello-world
+
+----
+
+LXD will be required for creating the OCI image. Make sure it is installed an
```
varshigupta12, line 45: Linking LXD docs here could be good

docs/tutorials/getting-started-with-flask.rst
```diff
@@ -0,0 +1,454 @@
...
+===============
+
+This is an optional but recommended step, especially if you're looking to
+deploy your rock into a production environment. With :ref:`chisel_explanation`
+you can produce lean and production-ready rocks by getting rid of all the
+contents that are not needed for your Flask application to run. This results
+in a much smaller rock with a reduced attack surface.
+
+The first step towards chiselling your rock is to ensure you are using a
+``bare`` base.
```
varshigupta12, line 299: It'll be nice to have some explanation or reference of bases linked here

docs/tutorials/getting-started-with-flask.rst
erinecon: I don't have a strong opinion either way. I think the title makes sense with or without "applications"; I mostly wanted to make sure that was the case

docs/tutorials/getting-started-with-flask.rst
cjdcordeiro: Yep.doing it now that the other PR has been merged

cjdcordeiro: done in d255831

docs/tutorials/getting-started-with-flask.rst
cjdcordeiro: done in 48992a5

tigarmo:
Thanks a lot everyone! I'm going to merge this as soon as it passes

cjdcordeiro (approved):

tigarmo (approved):

PR #597, opened 2024-06-20
doc: add explanation and guide for "bases"

cjdcordeiro:

-----

This PR adds a new explanation page for "Bases", which then refers to another new How-to guide on how to use `base` to go from a rock to a baseless and then chiselled rock.

This PR also fixes the `devel` reference in the reference docs and add doc links to the `rockcraft init` YAML template

### DIFF
--- a/docs/explanation/index.rst
+++ b/docs/explanation/index.rst
@@ -14,6 +14,7 @@ the motivations behind its development.
    chisel
    overlay-step
    rocks
+   bases
    pebble
    lifecycle-layer
    filesets
--- a/docs/explanation/rocks.rst
+++ b/docs/explanation/rocks.rst
@@ -15,6 +15,8 @@ Interoperability between rocks and other containers also extends to how
 container images are built, allowing rocks to be used as bases for existing
 build recipes, such as Dockerfiles, for further customisation and development.
 
+.. _what-sets-rocks-apart:
+
 What sets rocks apart?
 ~~~~~~~~~~~~~~~~~~~~~~
 
--- a/docs/how-to/chisel/index.rst
+++ b/docs/how-to/chisel/index.rst
@@ -7,3 +7,4 @@
    Create a package slice for Chisel <create-slice>
    Install a custom package slice <install-slice>
    Publish a slice definition <publish-slice>
+   Chisel an existing rock <chisel-existing-rock>
--- a/docs/reference/rockcraft.yaml.rst
+++ b/docs/reference/rockcraft.yaml.rst
@@ -67,6 +67,8 @@ A longer, possibly multi-line description of the rock.
 
 The rock version, used to tag the OCI image and name the rock file.
 
+.. _rockcraft_yaml_base:
+
 ``base``
 --------
 
@@ -84,14 +86,12 @@ which is typically used with static binaries or
    The notation "ubuntu:<channel>" is also supported for some channels, but this
    format is deprecated and should be avoided.
 
-.. note::
-   Base ``ubuntu@24.04`` is still unstable and under active development. To use
-   it, ``build-base`` *must* be ``devel``.
+.. _rockcraft_yaml_build_base:
 
 ``build-base``
 --------------
 
-**Type**: One of ``ubuntu@20.04 | ubuntu@22.04 | devel``
+**Type**: One of ``ubuntu@20.04 | ubuntu@22.04 | ubuntu@24.04 | devel``
 
 **Required**: Yes, if ``base`` is ``bare``
 
--- a/rockcraft/commands/init.py
+++ b/rockcraft/commands/init.py
@@ -59,6 +59,8 @@ class InitCommand(AppCommand):
         "simple": textwrap.dedent(
             """\
                 name: {name}
+                # see https://documentation.ubuntu.com/rockcraft/en/stable/explanation/bases/
+                # for more information about bases and using 'bare' bases for chiselled rocks
                 base: ubuntu@22.04 # the base environment for this rock
                 version: '0.1' # just for humans. Semantic versioning is recommended
                 summary: Single-line elevator pitch for your amazing rock # 79 char long summary
@@ -68,7 +70,7 @@ class InitCommand(AppCommand):
                     we live in tweetspace and your description wants to look good in the
                     container registries out there.
                 license: GPL-3.0 # your application's SPDX license
-                platforms: # The platforms this rock should be built on and run on
+                platforms: # the platforms this rock should be built on and run on
                     amd64:
 
                 parts:
@@ -79,6 +81,8 @@ class InitCommand(AppCommand):
         "flask-framework": textwrap.dedent(
             """\
                 name: {name}
+                # see https://documentation.ubuntu.com/rockcraft/en/stable/explanation/bases/
+                # for more information about bases and using 'bare' bases for chiselled rocks
                 base: ubuntu@22.04 # the base environment for this Flask application
                 version: '0.1' # just for humans. Semantic versioning is recommended
                 summary: A summary of your Flask application # 79 char long summary
@@ -88,16 +92,16 @@ class InitCommand(AppCommand):
                     we live in tweetspace and your description wants to look good in the
                     container registries out there.
                 license: GPL-3.0 # your application's SPDX license
-                platforms: # The platforms this rock should be built on and run on
+                platforms: # the platforms this rock should be built on and run on
                     amd64:
 
-                # To ensure the flask-framework extension works properly, your Flask application
+                # to ensure the flask-framework extension works properly, your Flask application
                 # should have an `app.py` file with an `app` object as the WSGI entrypoint.
                 extensions:
                     - flask-framework
 
 
-                # Uncomment the sections you need and adjust according to your requirements.
+                # uncomment the sections you need and adjust according to your requirements.
                 # parts:
                 #   flask-framework/dependencies:
                 #     stage-packages:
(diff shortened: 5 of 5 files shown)


linostar (changes requested):
Just few nitpicks.

docs/explanation/bases.rst
```diff
@@ -0,0 +1,67 @@
...
+Ubuntu bases
+------------
+
+When setting an Ubuntu
+:ref:`base <rockcraft_yaml_base>` for a rock, Rockcraft will pull the
+corresponding Ubuntu base container image from a container registry. This is
+exactly the same container one would get from places like
+`Docker Hub <docker_hub_>`_ or `Amazon ECR <ecr_>`_.
+
+The resulting rock will therefore also contain and software and utilities that
```
tigarmo, line 22:
```suggestion
The resulting rock will therefore also contain any software and utilities that
```

docs/explanation/bases.rst
```diff
@@ -0,0 +1,67 @@
...
+are present in the Ubuntu base container images.
+
+Setting an Ubuntu :ref:`base <rockcraft_yaml_base>` for a rock is
+especially useful when the goal is to build a rock that can be used for
+general-purpose environments such as a development working space.
+
+Bare bases
+----------
+
+As the name says, a ``bare`` :ref:`base <rockcraft_yaml_base>` indicates
```
tigarmo, line 32:
```suggestion
As the name suggests, a ``bare`` :ref:`base <rockcraft_yaml_base>` indicates
```

tigarmo (commented):
thanks! Just two easy comments

docs/explanation/bases.rst
```diff
@@ -0,0 +1,67 @@
...
+that the rock shall have no
+baseline system. This definition is similar to the "`scratch`_" Docker image,
+with the exception that a rock **can never be completely empty**, since it must
+always have :ref:`Pebble <pebble_explanation_page>` as its entrypoint, and
+include their own :ref:`metadata <what-sets-rocks-apart>` inside.
```
sergiusens, line 37: I think you are referring to the rock with _their_, there is too much indirection here as _their_ could be referring to Pebble for the unsuspecting victim and only figure it out by reading the metadata link.

```suggestion
baseline system. This definition is similar to the "`scratch`_" Docker image,
with the exception that every rocks itself always includes a minimum baseline
such as  :ref:`Pebble <pebble_explanation_page>`  and additional metadata:
:ref:`<what-sets-rocks-apart>`.
```
I would not alias the link for what sets the rock apart and let the original title render (I might have not gotten the syntax right)

docs/explanation/bases.rst
```diff
@@ -0,0 +1,67 @@
...
+When and how to use a ``bare`` base
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The ``bare`` :ref:`base <rockcraft_yaml_base>` is extremely useful when
+the goal is to build a lean **chiselled rock**. Such rocks are typically meant
+for production environments, and as such, the combination of the ``bare``
+:ref:`base <rockcraft_yaml_base>`
+with :ref:`Chisel <chisel_explanation>` slices will result in a minimalistic
+container image that has a reduced attack surface while retaining its
+functionality.
```
sergiusens, line 48: perhaps words on why these rocks are meant for production environments, or perhaps preferred is the better word.

I would go along the lines of reducing the surface area and the security vectors

docs/how-to/code/chisel-existing-rock/chiselled-rock/rockcraft.yaml
```diff
@@ -0,0 +1,15 @@
...
+summary: Chiselled Python3.11 rock
+description: Example of a chiselled Python rock
+platforms:
+  amd64:
+run-user: _daemon_  # setting a non-root user for the rock
+parts:
+  install-python-slices:
+    plugin: nil
+    stage-packages:
+      - base-files_release-info # we only want the release information
```
sergiusens, line 14: @cjdcordeiro doesn't python3.11_core depend on base-files_release-info. Do we need to explicitly declare it?

docs/how-to/rocks/chisel-existing-rock.rst
```diff
@@ -0,0 +1,132 @@
...
+    :end-before: [docs:run-bare-rock-end]
+    :dedent: 2
+
+
+The question then is: *how is rockcraft able to produce an equally functional
+Python rock with such a drastic reduction in size?* 
+
+And the answer is: *the rock no longer has the Ubuntu base as its first layer,
+and thus no longer has utilities like ``bash`` and ``apt`` (which aren't needed
+anyway)!*
```
sergiusens, line 87: I might add something along the lines: _for this use case_

sergiusens (approved):
well written, some comments to address. The struggle of howto vs tutorial is real!

docs/how-to/code/chisel-existing-rock/chiselled-rock/rockcraft.yaml
cjdcordeiro: `base-files` is one of those that due to being an essential, it isn't always listed as a pkg dependency. in this case, it's not even needed for python to run. the `release-info` slices only brings in stuff like `/etc/os-release` which is a must for allowing 3rd tools to audit and scan the resulting image

docs/explanation/bases.rst
cjdcordeiro: Sure. Applied the suggestion with some re-wording.

docs/explanation/bases.rst
cjdcordeiro: Sure. Re-worded it

docs/how-to/rocks/chisel-existing-rock.rst
cjdcordeiro: done

docs/how-to/rocks/chisel-existing-rock.rst
```diff
@@ -0,0 +1,132 @@
...
+
+.. literalinclude:: ../code/chisel-existing-rock/task.yaml
+    :language: bash
+    :start-after: [docs:run-chiselled-rock]
+    :end-before: [docs:run-chiselled-rock-end]
+    :dedent: 2
+
+**To conclude**, you've just created a general-purpose Python rock with just a
+few YAML lines and no code whatsoever! Then, by changing a couple of YAML
+fields (the ``base``), you've achieving a **~33% size reduction** while
```
tigarmo, line 123:
```suggestion
fields (the ``base``), you've achieved a **~33% size reduction** while
```

docs/how-to/rocks/chisel-existing-rock.rst
```diff
@@ -0,0 +1,132 @@
...
+maintaining functionality. Finally, by appending two words (literally, just
+the slice names) to the *rockcraft.yaml*, you were able to reduce the rock's
+size even further by an **additional ~37%** of its original size! In short:
+
++---------------+------------------+-----------+
```
tigarmo, line 128: This table is very compelling!

docs/how-to/rocks/chisel-existing-rock.rst
```diff
@@ -0,0 +1,132 @@
...
+
+This rock will also have other utilities like ``bash`` and ``apt`` which come
+from the rock's underlying Ubuntu 22.04 base.
+
+Rebuild the rock with a ``bare`` base
+-------------------------------------
+
+When starting to prepare the rock for production, the main goal is to get rid
+of all the software that is not necessary at runtime, and the first step
+towards achieving that goal is to use a ``bare``.
```
tigarmo, line 52:
```suggestion
towards achieving that goal is to use a ``bare`` base.
```

docs/how-to/rocks/chisel-existing-rock.rst
```diff
@@ -0,0 +1,132 @@
...
+
+In this example, the Python application is a very simple "Hello, world", which
+means we need nothing else but the core Python modules. For that, copy the
+previous *rockcraft.yaml* file to a new directory and simply append the desired
+package slice names to the list of ``stage-packages``, like this:
+
+.. literalinclude:: ../code/chisel-existing-rock/chiselled-rock/rockcraft.yaml
+    :language: yaml
+
+Pack it with
```
tigarmo, line 104:
```suggestion
Pack it with:
```
(for consistency with the other instances of this construct in this document)

docs/how-to/rocks/chisel-existing-rock.rst
```diff
@@ -0,0 +1,132 @@
...
+
+.. literalinclude:: ../code/chisel-existing-rock/task.yaml
+    :language: bash
+    :start-after: [docs:pack-chiselled-rock]
+    :end-before: [docs:pack-chiselled-rock-end]
+    :dedent: 2
+
+and the end result will be an astoundingly small Python rock with **13MB**!
```
tigarmo, line 112:
```suggestion
And the end result will be an astoundingly small Python rock with **13MB**!
```
(same reason)

requirements-doc.txt
```diff
@@ -81,6 +81,7 @@ sphinxcontrib-jquery==4.1
 sphinxcontrib-jsmath==1.0.1
 sphinxcontrib-qthelp==1.0.7
 sphinxcontrib-serializinghtml==1.1.10
+sphinx-toolbox==3.5.0
```
tigarmo, line 84: Since there are some fixes to be done anyway, please also add `sphinx-toolbox` [here](https://github.com/canonical/rockcraft/blob/main/setup.cfg#L47) (we use those deps when freezing the requirements).

tigarmo (changes requested):

docs/how-to/rocks/chisel-existing-rock.rst
cjdcordeiro: this is on purpose. I have a love & hate relationship with code blocks, cause in theory, there's no period  and thus no end of sentence. The scheme I was using was: _if the code block is preceded by `:`, then the next sentence is capitalized. otherwise it shall just continue in lower case_

For the sake of moving this PR, I've just accepted all changes

requirements-doc.txt
cjdcordeiro: noted. done in cb6e68a

tigarmo (approved):

PR #602, opened 2024-06-21
fix: enable overlays in _enable_craft_parts_features()

tigarmo:

-----

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -248,3 +248,20 @@ def run_lifecycle(mocker, default_build_plan):
         )
 
     return _inner
+
+
+@pytest.fixture(autouse=True)
+def reset_features():
+    from craft_parts import Features
+
+    Features.reset()
+    yield
+    Features.reset()
+
+
+@pytest.fixture()
+def enable_overlay_feature(reset_features):
+    """Enable the overlay feature."""
+    from craft_parts import Features
+
+    Features(enable_overlay=True)
--- a/rockcraft/application.py
+++ b/rockcraft/application.py
@@ -71,3 +71,12 @@ class Rockcraft(Application):
         Should be overridden by applications that need to register plugins at startup.
         """
         return plugins.get_plugins()
+
+    @override
+    def _enable_craft_parts_features(self) -> None:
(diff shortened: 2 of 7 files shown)


mr-cal (approved):

sergiusens:
```
sergiusens@cuivienen:~/Source/canonical/rockcraft$ snap refresh --amend --classic rockcraft --channel edge/pr-602
rockcraft (edge/pr-602) 1.4.0.post11+g64fe731 from Canonical✓ refreshed
sergiusens@cuivienen:~/Source/canonical/rockcraft$ mkdir foo
sergiusens@cuivienen:~/Source/canonical/rockcraft$ cd foo/
sergiusens@cuivienen:~/Source/canonical/rockcraft/foo$ rockcraft init
deleting current features configuration
sergiusens@cuivienen:~/Source/canonical/rockcraft/foo$ ls
rockcraft.yaml

**Comment by sergiusens** [comment]
nevermind! I was using my venv by accident :/

**Review by sergiusens — APPROVED** [review]

File: tests/conftest.py [hunk 1]
```diff
@@ -248,3 +248,20 @@ def _inner(**kwargs):
         )
 
     return _inner
+
+
+@pytest.fixture(autouse=True)
+def reset_features():
```
sergiusens, line 254: there is no need for this to be an independent fixture, is there?

Could this not be cleaner

```python
@pytest.fixture()
def enable_overlay_feature(reset_features):
    """Enable the overlay feature."""
    from craft_parts import Features

    Features.reset()
    Features(enable_overlay=True)
    yield
    Features.reset()
```

?

tigarmo: > there is no need for this to be an independent fixture, is there?

There is, actually. If this feature is not used in *all* tests, then two consecutive tests that call `Application.run()` will try to enable the overlay feature twice and fail. It's a side-effect of the `Features` singleton.
I add it as a parameter explicitly in the `enable_overlay_feature()` fixture to ensure correct order: first `reset_features` is setup, then `enable_overlay_feature`, then the test runs, then teardown in the reverse order

PR #604, opened 2024-06-22
fix: switch from progress to message for init

sergiusens:
- Snapcraft uses message, progress hides the docs url as it is "uni-line"
- Use a similar message to what is done in Snapcraft
- Improve the test for initializing flask-framework
- Use the correct documentation slug

-----

### DIFF
--- a/tests/unit/test_cli.py
+++ b/tests/unit/test_cli.py
@@ -16,6 +16,7 @@
 import pathlib
 import sys
 from pathlib import Path
+import textwrap
 from unittest.mock import DEFAULT, call
 
 import pytest
@@ -117,11 +118,10 @@ def test_run_init_fallback_name(mocker):
     assert rock_project.name == "my-rock-name"
 
 
-@pytest.mark.usefixtures("new_dir")
-def test_run_init_flask(mocker, tmp_path, monkeypatch):
-    (tmp_path / "requirements.txt").write_text("flask")
-    (tmp_path / "app.py").write_text("app = object()")
-    mock_ended_ok = mocker.patch.object(emit, "ended_ok")
+def test_run_init_flask(mocker, emitter, monkeypatch, new_dir, tmp_path):
+    (new_dir / "requirements.txt").write_text("Flask", encoding="utf-8")
+    (new_dir / "app.py").write_text("app = object()", encoding="utf-8")
+
     mocker.patch.object(sys, "argv", ["rockcraft", "init", "--profile=flask-framework"])
 
     cli.run()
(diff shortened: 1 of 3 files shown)


cmatsuoka (approved):

cjdcordeiro (changes requested):
ty.

maybe you could use the new doc_slug variable here as well: https://github.com/canonical/rockcraft/pull/604/files#diff-a065a85026852eac577f63c3df7d7502eb85371b5ced140674a9e9b66d6bdaadR110 so we don't have the boilerplate text.

Another question, probably a bit more controversial, is about `stable` vs `latest`. Whoever is using `latest/edge` will be getting doc links that might not exist :/

Should the doc_slugs be based on the rockcraft version?

tigarmo:
I'm fine with the idea in principle (though we'd probably do it in craft-application instead). The problem is that by default new versions on readthedocs are private - that is, if I'm logged out I can access https://documentation.ubuntu.com/rockcraft/en/stable/ but not https://documentation.ubuntu.com/rockcraft/en/1.4.0/

And I can't seem to find a setting to change this default visibility.

tests/unit/test_cli.py
```diff
@@ -131,7 +131,12 @@ def test_run_init_flask(mocker, tmp_path, monkeypatch):
 
     assert len(rock_project_yaml["summary"]) < 80
     assert len(rock_project_yaml["description"].split()) < 100
-    assert mock_ended_ok.mock_calls == [call()]
-
+    emitter.assert_message(
+        textwrap.dedent(
+            """\
+        Created 'rockcraft.yaml'.
```
tigarmo, line 137: do you specifically want the single quotes around rockcraft.yaml?

sergiusens:
@cjdcordeiro yes to what you say, but not now. We discussed this and we decided to do it after the rush. Also, not part of this PR per se.

tests/unit/test_cli.py
sergiusens: I did say I am following `snapcraft init`, which produces:

    Created 'snap/snapcraft.yaml'.
    Go to https://docs.snapcraft.io/the-snapcraft-format/8337 for more information about the snapcraft.yaml format.

sergiusens:
@cjdcordeiro fwiw, I do want versioned slugs :-)

cjdcordeiro:
@sergiusens fair enough. Let's then at least parameterize the doc_url in the YAML text. 

Do you already have an issue somewhere with the "versioned slugs" task? (ack @tigarmo , I didn't know that either. can you please raise that issue in the `#documentation` channel? Btw, I can manually edit the 1.4.0 visibility and make it public, but indeed, I see that some versions are public and some are private...and I don't know if there's a way to automatically set them all to public)

sergiusens:
@cjdcordeiro just did :-)

tigarmo:
> can you please raise that issue in the #documentation channel?

Sure, I'll do that on monday. Since we have readthedocs for business we can probably raise a support ticket

tigarmo (approved):

PR #606, opened 2024-06-22
Use new python3.10-venv slices when bare in flask-framework extension

gregory-schiano:

-----

Using the new `python3.11-venv` slice to reduce the image of the generated rock when base is `bare` using `flask-framework` and `django-framework` extensions.

How to test:
Install a build of rockcraft using this branch, you need to have `docker` installed too
```bash
git pull https://github.com/canonical/sample-flask.git
cd sample-flask
rockcraft init --profile flask-framework
# Edit the required fields in rockcraft.yaml
# 1. We ensure that when using ubuntu@22.04 base it still works
rockcraft pack
rockcraft.skopeo copy --insecure-policy oci-archive:sample-flask_0.1_amd64.rock docker-daemon:sample-flask:ubuntu
docker run -d --name flask-ubuntu -p 8000:8000 sample-flask:ubuntu
curl localhost:8000
# Should return <h1>Hello</h1>
docker exec -ti flask-ubuntu bash
/bin/python3 --version
# Should return "Python 3.10.12"
exit
docker stop flask-ubuntu
# Edit rockcraft.yaml, change base to "bare" and add "build-base: ubuntu@22.04"
rockcraft clean
rockcraft pack
rockcraft.skopeo copy --insecure-policy oci-archive:sample-flask_0.1_amd64.rock docker-daemon:sample-flask:bare
docker run -d --name flask-bare -p 8000:8000 sample-flask:bare
curl localhost:8000
# Should return <h1>Hello</h1>
docker exec -ti flask-bare bash
/bin/python3 --version
# Should return "Python 3.10.12"
exit
docker stop flask-bare
```

You can also test using the tutorial (follow the chiseling part of the tutorial):
https://documentation.ubuntu.com/rockcraft/en/latest/tutorials/getting-started-with-flask/


rockcraft/extensions/gunicorn.py
```diff
@@ -66,13 +66,20 @@ def gen_install_app_part(self) -> Dict[str, Any]:
     def _gen_parts(self) -> dict:
         """Generate the parts associated with this extension."""
         data_dir = get_extensions_data_dir()
+        stage_packages = ["python3-venv"]
```
sergiusens, line 69: you can still use chisel with a non bare base, there is file dedup in any case, so not as important, but there is more _scripting_ with chisel

gregory-schiano: That's true, but doing so will change python version for `python3.10` in `ubuntu@22.04` base to `python3.11`, which might be a breaking change in the extension.
I'd rather keep the default python version of jammy in the `ubuntu@22.04` base

cjdcordeiro: why do you say it's a breaking change?

jdkandersson: Going to `python3.11` is a breaking change to a degree because Python usually does some minor backwards incompatible things even within minor versions. They usually don't impact many people (I haven't checked the specifics for 3.10 -> 3.11).

The more convincing reason not to go to `python3.11` for Ubuntu 22.04 is that people would expect Python 3.10 if they are building on top of Ubuntu 22.04 since that is what ships with it

jdkandersson:
Why is it `Python 3.11.0rc1`? The latest version of Python 3.11 is 3.11.9, 0rc1 I think is pre-release so probably doesn't contain a bunch of security fixes etc

rockcraft/extensions/gunicorn.py
jdkandersson: And it also looks like for some reason we would be shipping `3.11.0rc1` which doesn't seem like what we want to do when `3.11.9` is out

cjdcordeiro (changes requested):
Btw, the docs (https://documentation.ubuntu.com/rockcraft/en/stable/tutorials/getting-started-with-flask/#chisel-the-rock) should be updated as I'm expecting the resulting rock to now be more than 15% smaller

rockcraft/extensions/gunicorn.py
gregory-schiano: So few more elements:
Ubuntu Main python package is 3.10
The sliced package in python3.11 which comes from universe

I think for base `ubuntu@22.04` we should either use the apt package or the python3.10 sliced (which doesn't exists yet).

gregory-schiano:
> Btw, the docs (https://documentation.ubuntu.com/rockcraft/en/stable/tutorials/getting-started-with-flask/#chisel-the-rock) should be updated as I'm expecting the resulting rock to now be more than 15% smaller

Indeed, but not that this %age (like the size also provided in this tutorial) might change at any time due to new packages, and dependencies.

cjdcordeiro:
> > Btw, the docs (https://documentation.ubuntu.com/rockcraft/en/stable/tutorials/getting-started-with-flask/#chisel-the-rock) should be updated as I'm expecting the resulting rock to now be more than 15% smaller
> 
> Indeed, but not that this %age (like the size also provided in this tutorial) might change at any time due to new packages, and dependencies.

I was my impression, from this PR, that if you use a `base: bare`, the extension will use a slice instead of a deb (which is the case today). Thus, the resulting rock will be smaller.

gregory-schiano:
> > > Btw, the docs (https://documentation.ubuntu.com/rockcraft/en/stable/tutorials/getting-started-with-flask/#chisel-the-rock) should be updated as I'm expecting the resulting rock to now be more than 15% smaller
> > 
> > 
> > Indeed, but not that this %age (like the size also provided in this tutorial) might change at any time due to new packages, and dependencies.
> 
> It was my impression, from this PR, that if you use a `base: bare`, the extension will use a slice instead of a deb (which is the case today). Thus, the resulting rock will be smaller.

You understood properly, but the size of the images depends on the size of the dependencies installed and the size of the files the slices are setting. This can change overtime.

cjdcordeiro:
> > > > Btw, the docs (https://documentation.ubuntu.com/rockcraft/en/stable/tutorials/getting-started-with-flask/#chisel-the-rock) should be updated as I'm expecting the resulting rock to now be more than 15% smaller
> > > 
> > > 
> > > Indeed, but not that this %age (like the size also provided in this tutorial) might change at any time due to new packages, and dependencies.
> > 
> > 
> > It was my impression, from this PR, that if you use a `base: bare`, the extension will use a slice instead of a deb (which is the case today). Thus, the resulting rock will be smaller.
> 
> You understood properly, but the size of the images depends on the size of the dependencies installed and the size of the files the slices are setting. This can change overtime.

Ofc it can, but in that case, we wouldn't be stating any sizes anywhere, and we are (64MB , and then 15% reduction). You do agree that, with your changes, the size reduction will **no longer** be 15% right? If that's true, then the doc needs fixing

gregory-schiano:
> > > > > Btw, the docs (https://documentation.ubuntu.com/rockcraft/en/stable/tutorials/getting-started-with-flask/#chisel-the-rock) should be updated as I'm expecting the resulting rock to now be more than 15% smaller
> > > > 
> > > > 
> > > > Indeed, but not that this %age (like the size also provided in this tutorial) might change at any time due to new packages, and dependencies.
> > > 
> > > 
> > > It was my impression, from this PR, that if you use a `base: bare`, the extension will use a slice instead of a deb (which is the case today). Thus, the resulting rock will be smaller.
> > 
> > 
> > You understood properly, but the size of the images depends on the size of the dependencies installed and the size of the files the slices are setting. This can change overtime.
> 
> Ofc it can, but in that case, we wouldn't be stating any sizes anywhere, and we are (64MB , and then 15% reduction). You do agree that, with your changes, the size reduction will **no longer** be 15% right? If that's true, then the doc needs fixing

Sure agree, that was just a side note.
I've already updated the doc, we went from 65MB to 46MB, so 30% less

cjdcordeiro:
Waiting on https://github.com/canonical/chisel-releases/pull/271

cjdcordeiro (commented):
the spread test needs an easy fix (remove the rockcraft.yaml)

rockcraft/extensions/gunicorn.py
gregory-schiano: We decided to slice python3.10 in Ubuntu 22.04 and use it. I close the thread.

gregory-schiano:
For reference, the ubuntu 22.04 base leads to a 65MB image, while the bare base image (with python3.10) leads to a 44MB image size

cjdcordeiro (approved):

rockcraft/extensions/gunicorn.py
```diff
@@ -66,13 +66,20 @@ def gen_install_app_part(self) -> Dict[str, Any]:
...
+        build_environment = []
+        if self.yaml_data["base"] == "bare":
+            stage_packages = ["python3.10-venv_ensurepip"]
+            build_environment = [{"PARTS_PYTHON_INTERPRETER": "python3.10"}]
```
tigarmo, line 73: is there no slice that provides the `python3` symlink? (just out of curiosity)

tests/spread/rockcraft/extension-flask-uncomment-blocks/task.yaml
```diff
@@ -0,0 +1,82 @@
...
+  BLOCK/bare: runtime-slices
+  BLOCK/base_2204: runtime-debs
+execute: |
+  NAME="flask-${SCENARIO//./-}"
+  ROCK_FILE="${NAME}_0.1_amd64.rock"
+  IMAGE="${NAME}:0.1"
+
+  run_rockcraft init --name flask-extension --profile flask-framework
+  sed -i "s/name: .*/name: ${NAME}/g" rockcraft.yaml
+  if [ "${SCENARIO}" = "bare" ]; then
```
tigarmo, line 14: I wonder if this would be easier to read/maintain with regular `diff` and `patch`? As in, do the base `init`, update the project to use a bare-base, save that diff as `scenario-bare.patch`
It's a though for short-term improvements, we don't need to do this today

tests/spread/rockcraft/extension-flask/task.yaml
```diff
@@ -1,33 +1,47 @@
...
+  SCENARIO/bare: bare
+  SCENARIO/base_2204: ubuntu-22.04
 execute: |
+  NAME="flask-${SCENARIO//./-}"
+  ROCK_FILE="${NAME}_0.1_amd64.rock"
+  IMAGE="${NAME}:0.1"
+
   run_rockcraft init --name flask-extension --profile flask-framework
+  sed -i "s/name: .*/name: ${NAME}/g" rockcraft.yaml
+  if [ "${SCENARIO}" = "bare" ]; then
```
tigarmo, line 12: same thought here re: using diffs

tigarmo (approved):

tigarmo:
Thanks a lot! Does anyone else want to look into this before I sync it with main & merge?

cjdcordeiro:
@tigarmo please wait for a check on the documentation. Is https://documentation.ubuntu.com/rockcraft/en/stable/reference/extensions/flask-framework/ up to date after this merge?

If the rockcraft.yaml accurate?

rockcraft/extensions/gunicorn.py
gregory-schiano: Python3 is available in the final image, but during this phase python3-venv apt package is automatically installed by the python plugin, and it gets in conflict with the slice I think. Couldn't make it work without the build-environment

tests/spread/rockcraft/extension-flask-uncomment-blocks/task.yaml
gregory-schiano: I made this if/else to allow flexibility in the test. Supporting ubuntu@24.04 will be just a matter of adding a variant. Managing it with patches might become tricky

gregory-schiano:
> @tigarmo please wait for a check on the documentation. Is https://documentation.ubuntu.com/rockcraft/en/stable/reference/extensions/flask-framework/ up to date after this merge?
> 
> If the rockcraft.yaml accurate?

I didn't change anything about the user experience. The documentation update needed was about the size reduction of the produced chiseled rock.
I've done a review of the existing documentation to ensure something needs to be updated, and didn't find anything else.
But could be good to have a second pair of eyes

PR #607, opened 2024-06-24
small fixes

jdkandersson:

-----

Found a couple of small mistakes:

- The instructions for a code editor were already in the generalised setup instructions so I removed it from the tutorial
- The rock size was based on an arm64 rock, now I fixed it for the amd64 rock
- Minor grammar/ styling

### DIFF
--- a/docs/tutorials/getting-started-with-flask.rst
+++ b/docs/tutorials/getting-started-with-flask.rst
@@ -19,10 +19,6 @@ Finally, create a new directory for this tutorial and go inside it:
    mkdir flask-hello-world
    cd flask-hello-world
 
-Note that you'll also need a text editor. You can either install one of your
-choice or simply use one of the already existing editors in your Ubuntu
-environment (like ``vi``).
-
 Create the Flask application
 ============================
 
@@ -127,7 +123,7 @@ extension:
     :end-before: [docs:ls-rock-end]
     :dedent: 2
 
-The created rock is about 62MB in size. We will reduce its size later in this
+The created rock is about 65MB in size. We will reduce its size later in this
 tutorial.
 
 .. note::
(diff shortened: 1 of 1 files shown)


docs/tutorials/getting-started-with-flask.rst
```diff
@@ -19,10 +19,6 @@ Finally, create a new directory for this tutorial and go inside it:
    mkdir flask-hello-world
    cd flask-hello-world
 
-Note that you'll also need a text editor. You can either install one of your
-choice or simply use one of the already existing editors in your Ubuntu
-environment (like ``vi``).
```
jdkandersson, line 24: Already included in the literalinclude above

docs/tutorials/getting-started-with-flask.rst
```diff
@@ -127,7 +123,7 @@ extension:
     :end-before: [docs:ls-rock-end]
     :dedent: 2
 
-The created rock is about 62MB in size. We will reduce its size later in this
+The created rock is about 65MB in size. We will reduce its size later in this
```
jdkandersson, line 126: This is the amd64 rock size (also checked the reduction size is still roughly 15%)

docs/tutorials/getting-started-with-flask.rst
cjdcordeiro: true

cjdcordeiro (changes requested):
just need to fix the linters and should be good

jdkandersson:
> just need to fix the linters and should be good

None of the failures are related to changes in this PR as far as i can tell - I probably don't have context to easily fix those issues

cjdcordeiro (approved):
re-ran the tests - all good now.

docs/tutorials/getting-started-with-flask.rst
sergiusens: If this image uses ubuntu as the base, that base image changes size constantly, so it might be better to speak in relative terms and mention that it got smaller in the following iteration instead and here mention only that it will

tigarmo (approved):

PR #615, opened 2024-06-25
fix(flask): drop 'flask-framework/' prefix in template

tigarmo:
runtime-slices and runtime-debs aren't parts that the extension is aware of, so prefixing them with 'flask-framework/' is misleading and a conceptual error.

-----


tigarmo:
Linter failure is linkcheck complaining about https://www.gnu.org/software/make/manual/make.html#DESTDIR

cjdcordeiro (approved):

sergiusens (approved):

PR #620, opened 2024-07-01
docs: tutorials->tutorial and rename flask page

cjdcordeiro:

-----

 - rename Tutorials to Tutorial
  >  Tutorial because there should really be just one tutorial. One learning experience, even if in multiple sections.

 - rename the Flask tutorial .rst file in order to achieve a smaller URL
 - the docs are also versioned and without language, so this PR also removes the `/en/` from the links in the codebase


tigarmo:
Did you already change the readthedocs versioning scheme? Because now eg https://documentation.ubuntu.com/rockcraft/en/stable/reference/rockcraft.yaml is a 404 and that breaks links printed by stable Rockcraft.

cjdcordeiro:
@tigarmo thanks for reverting it.

.github/ISSUE_TEMPLATE/docs.yaml
```diff
@@ -32,7 +32,7 @@ body:
     attributes:
       label: Documentation location
       description: Location of the corresponding documentation page (e.g., file, section, URL), if applicable
-      placeholder: ex. https://canonical-rockcraft.readthedocs-hosted.com/en/latest/tutorials/hello-world/
+      placeholder: ex. https://canonical-rockcraft.readthedocs-hosted.com/latest/tutorial/hello-world/
```
sergiusens, line 35: why don't you use the friendly ubuntu link here too?

cjdcordeiro, line 35:
```suggestion
      placeholder: ex. https://documentation.ubuntu.com/rockcraft/stable/tutorial/hello-world/
```

cjdcordeiro: yep

tigarmo (changes requested):
I'm fine with the changes in principle, but unless this PR also add redirects it'll break the links on older versions of Rockcraft

cjdcordeiro:
Which older versions have tutorial links in the codebase? This current version doesn't have any

tigarmo (approved):
oh you're right! Thanks

mr-cal (approved):

PR #631, opened 2024-07-08
Add go init profile and go-extension

javierdelapuente:

-----

This PR implements the spec ISD153, "12-Factor Go Support" for rockcraft.

For that, a new init profile called "go-framework" and a new go extension, called "go-framework" are needed. These additions are similar to the flask and django init profiles and extensions, with several differences. Some of the main differences are:
 - Preference for base bare.
 - No dependency on gunicorn. Go generates binaries that are directly executable.
 - By default only one binary is installed  (by convention the one with the project name). Including other binaries in the rock will require to change the rockcraft.yaml file.
 - No statsd.

Also, new platforms are added to django, flask extension, so they are the same as in Go.


rockcraft/extensions/go.py
```diff
@@ -0,0 +1,213 @@
...
+
+
+class GoFramework(Extension):
+    """An extension class for Go applications."""
+
+    @staticmethod
+    @override
+    def get_supported_bases() -> Tuple[str, ...]:
+        """Return supported bases."""
+        return "bare", "ubuntu@22.04", "ubuntu:22.04"
```
javierdelapuente, line 36: Add also 24.04

tests/unit/extensions/test_go.py
```diff
@@ -0,0 +1,236 @@
...
+        },
+    }
+
+
+@pytest.mark.usefixtures("go_extension")
+def test_go_extension_no_go_mod_file_error(tmp_path, go_input_yaml):
+    (tmp_path / "somefile").write_text("random text")
+    with pytest.raises(ExtensionError) as exc:
+        extensions.apply_extensions(tmp_path, go_input_yaml)
+    assert str(exc.value) == "missing go.mod file"
```
jdkandersson, line 84: Check for the doc link

javierdelapuente: Done!

cjdcordeiro:
@rebornplusplus could be nice for you to have a look at this PR, as it might be one of the future use cases for the user research activities

jdkandersson (approved):

javierdelapuente:
Hi @tigarmo, can you have a look at this PR?. Thank you!

rockcraft/extensions/go.py
```diff
@@ -0,0 +1,213 @@
...
+    def project_name(self):
+        """Return the normalized name of the rockcraft project."""
+        return self.yaml_data["name"]
+
+    def _check_project(self):
+        """Check go.mod file exist in project."""
+        if not (self.project_root / "go.mod").exists():
+            raise ExtensionError(
+                "missing go.mod file",
+                docs_url="https://documentation.ubuntu.com/rockcraft/en/stable/reference/extensions/go-framework",
```
tigarmo, line 101:
```suggestion
                doc_slug="/reference/extensions/go-framework",
```

rockcraft/extensions/go.py
```diff
@@ -0,0 +1,213 @@
...
+                    "migrate",
+                    "migrate.sh",
+                    "static",
+                    "templates",
+                )
+                if (self.project_root / f).exists()
+            ]
+        return user_stage
+
+    def _get_nested(self, obj: dict, path: str) -> dict:
```
tigarmo, line 209: two comments here: 
1) I believe dots are valid in part names, so this is fragile. How about changing it to a list of names instead? E.g. the signature could be `def _get_nested(self, obj, *parts: str)`
2) This says it returns `dict` but `build-snaps` is a list. Seems like this should return `Any`

rockcraft/extensions/go.py
```diff
@@ -0,0 +1,213 @@
...
+        """Return the assets stage list for the Go project."""
+        user_stage = self._get_nested(self.yaml_data, "parts.go-framework/assets").get(
+            "stage", []
+        )
+
+        if not all(re.match("-? *app/", p) for p in user_stage):
+            raise ExtensionError(
+                "go-framework extension requires the 'stage' entry in the "
+                "go-framework/assets part to start with app",
+                docs_url="https://documentation.ubuntu.com/rockcraft/en/stable/reference/extensions/go-framework",
```
tigarmo, line 193: same doc_slug as the other comment

tigarmo (changes requested):
thanks!

rockcraft/commands/init.py
```diff
@@ -103,6 +103,9 @@ class InitCommand(AppCommand):
                     container registries out there.
                 platforms: # the platforms this rock should be built on and run on
                     amd64:
+                    arm64:
```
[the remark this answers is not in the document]
javierdelapuente: I put all of them in the `charmcraft` [PR for go](https://github.com/canonical/charmcraft/pull/1800#discussion_r1719891244), and I was suggested to remove armhf and riscv64, as they are not supported by the juju snap.

rockcraft/extensions/go.py
javierdelapuente: Updated to be a list instead of a string so dots could be used in part names.

I changed to not get `build-snaps` directly, as it can be a list, and instead get the object. Otherwise I would need to complicate the _get_nested function to check for the type (and as it is a very simple helper, I preferred not to handle all cases unless more needs arise).

rockcraft/extensions/go.py
```diff
@@ -0,0 +1,214 @@
...
+
+
+class GoFramework(Extension):
+    """An extension class for Go applications."""
+
+    @staticmethod
+    @override
+    def get_supported_bases() -> Tuple[str, ...]:
+        """Return supported bases."""
+        return "bare", "ubuntu@22.04", "ubuntu:22.04", "ubuntu@24.04"
```
jdkandersson, line 36: Should we remove 22.04 here?

javierdelapuente: Removed

rockcraft/extensions/go.py
```diff
@@ -206,8 +207,8 @@ def _assets_stage(self):
             ]
         return user_stage
 
-    def _get_nested(self, obj: dict, path: str) -> dict:
-        """Get a nested object using a path (a dot-separated list of keys)."""
-        for key in path.split("."):
+    def _get_nested(self, obj: dict, paths: list[str]) -> dict:
```
tigarmo, line 210: I did mean `*paths: str` instead of `paths: list[str]` so that callers can do `self._get_nested(data, "a", "b", "c")` instead of `self._get_nested(data, ["a", "b", "c"])`, but I suppose it doesn't make much of a difference

tigarmo (approved):
nice! thanks

javierdelapuente:
Commented all bases except amd64 as discussed.

rockcraft/commands/init.py
```diff
@@ -103,9 +103,9 @@ class InitCommand(AppCommand):
                     container registries out there.
                 platforms: # the platforms this rock should be built on and run on
                     amd64:
-                    arm64:
-                    ppc64el:
-                    s390x:
+                    # arm64:
```
jdkandersson, line 106: Also here, perhaps include an explanatory comment?

jdkandersson (approved):

rockcraft/commands/init.py
javierdelapuente: Done!

tests/unit/extensions/test_go.py
```diff
@@ -0,0 +1,237 @@
...
+            "migrate": "app/migrate",
+            "migrate.sh": "app/migrate.sh",
+            "templates": "app/templates",
+        },
+        "stage": ["app/migrate", "app/migrate.sh", "app/static", "app/templates"],
+    }
+
+
+@pytest.mark.usefixtures("go_extension")
+def test_go_extension_extra_assets_overriden(tmp_path, go_input_yaml):
```
lengau, line 219:
```suggestion
def test_go_extension_extra_assets_overridden(tmp_path, go_input_yaml):
```

tests/unit/extensions/test_go.py
```diff
@@ -0,0 +1,237 @@
...
+        ),
+        (
+            {
+                "bin/anotherbinary": "usr/local/bin/goprojectname",
+            },
+            {"bin/anotherbinary": "usr/local/bin/goprojectname"},
+        ),
+    ],
+)
+def test_go_extension_override_organize(
```
lengau, line 128:
```suggestion
def test_go_extension_overrides_organize(
```

rockcraft/extensions/go.py
```diff
@@ -0,0 +1,214 @@
...
+                if snap.startswith("go"):
+                    return True
+        build_packages = install_app.get("build-packages", [])
+        if build_packages:
+            for package in build_packages:
+                if package in ["gccgo-go", "golang-go"]:
+                    return True
+        return False
+
+    def _gen_install_assets_part(self) -> Dict[str, Any] | None:
```
lengau, line 162:
```suggestion
    def _get_install_assets_part(self) -> Dict[str, Any] | None:
```

In general I think `gen` at the start of a method implies that it's a generator

rockcraft/extensions/go.py
```diff
@@ -0,0 +1,214 @@
...
+    def _check_project(self) -> None:
+        """Check go.mod file exist in project."""
+        if not (self.project_root / "go.mod").exists():
+            raise ExtensionError(
+                "missing go.mod file",
+                doc_slug="/reference/extensions/go-framework",
+                logpath_report=False,
+            )
+
+    def _gen_install_app_part(self) -> Dict[str, Any]:
```
lengau, line 105:
```suggestion
    def _get_install_app_part(self) -> Dict[str, Any]:
```

lengau (approved):
LGTM, with some minor naming suggestions

rockcraft/extensions/go.py
```diff
@@ -0,0 +1,214 @@
...
+            },
+        }
+
+        snippet["parts"] = {
+            # This is needed in case there is no assets part, as the working directory is /app
+            "go-framework/base-layout": {
+                "plugin": "nil",
+                "override-build": "mkdir -p ${CRAFT_PART_INSTALL}/app",
+            },
+            "go-framework/install-app": self._gen_install_app_part(),
```
tigarmo, line 68:
```suggestion
            "go-framework/install-app": self._get_install_app_part(),
```

rockcraft/extensions/go.py
```diff
@@ -0,0 +1,214 @@
...
+            "go-framework/runtime": {
+                "plugin": "nil",
+                "stage-packages": ["ca-certificates_data"],
+            },
+        }
+
+        assets_part = self._gen_install_assets_part()
```
tigarmo, line 75:
```suggestion
        assets_part = self._get_install_assets_part()
```

PR #648, opened 2024-07-23
Fixed tutorial how to create a package slice for chisel

clay-lake:
Updated tutorial "How to create a package slice for Chisel" to reflect updates in chisel-releases effecting our docs tests.

-----

### DIFF
--- a/docs/how-to/code/create-slice/openssl.yaml
+++ b/docs/how-to/code/create-slice/openssl.yaml
@@ -1,19 +1,32 @@
 package: openssl
-slices:
-    bins:
-        essential:
-          - libc6_libs
-          - libc6_config
-          - libssl3_libs
-          - openssl_config
-        contents:
-            /usr/bin/c_rehash:
-            /usr/bin/openssl:
 
-    config:
-        contents:
-            /etc/ssl/private/:
-            /etc/ssl/openssl.cnf:
-            /usr/lib/ssl/certs:
-            /usr/lib/ssl/openssl.cnf:
-            /usr/lib/ssl/private:
+essential:
+  - openssl_copyright
+
+slices:
+  bins:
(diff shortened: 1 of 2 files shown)


cjdcordeiro (approved):

tigarmo (approved):

docs/how-to/code/create-slice/openssl.yaml
```diff
@@ -1,19 +1,32 @@
...
+      /etc/ssl/certs/:
+      /etc/ssl/openssl.cnf:
+      /etc/ssl/private/:
+      /usr/lib/ssl/certs:
+      /usr/lib/ssl/openssl.cnf:
+      /usr/lib/ssl/private:
+
+  data:
+    contents:
+      /usr/lib/ssl/cert.pem: {symlink: /etc/ssl/certs/ca-certificates.crt}
```
zhijie-yang, line 28: For Ubuntu >= 24.04, the `/usr/lib/ssl/cert.pem -> /etc/ssl/certs/ca-certificates.crt` exists in the openssl package by default. So leave here a comment that the `{symlink: /etc/...}` is only needed for focal and jammy.

zhijie-yang (changes requested):
Thanks for the changes, just a nitpick. Please also observe the changes that caused the linting issue.

docs/how-to/code/create-slice/openssl.yaml
clay-lake: Thanks. I think I surpassed the line limit with the last suggestion I applied.

Re the comment, I removed it from the original openssl.yaml I copied because I thought the comment would be distracting in the context of demoing slices.

linostar (approved):
LGTM, thanks!

PR #653, opened 2024-07-29
docs: add a guide for adding internal users to rocks

linostar:

-----

Adding a new how-to guide that explains how to add an internal user or group (aside from `_daemon_`) inside a rock.

The method selected for that is by adding the user/group in override-build instead of doing it in overlay. This is because overlay won't work on rocks with bare bases.


mr-cal (approved):

docs/how-to/rocks/add-internal-user-in-rocks.rst
cjdcordeiro, line 1: it would also be nice to show set a rock's pebble service to run as this user

cjdcordeiro (changes requested):

docs/how-to/rocks/add-internal-user-in-rocks.rst
```diff
@@ -0,0 +1,45 @@
+.. _add_internal_user_in_rocks:
+
+How to add an internal user to a rock
+*************************************
+
+Rockcraft allows you to use ``run-user`` to specify the user you want to the
```
lengau, line 6:
```suggestion
Rockcraft allows you to use ``run-user`` to specify the user you want to run the
```

docs/how-to/rocks/add-internal-user-in-rocks.rst
```diff
@@ -0,0 +1,45 @@
...
+rock's service(s) with. Without using that field in your rockcraft.yaml, the
+service(s) will be run by the root user. The downside, however, of using
+``run-user`` is that only supports one user called ``_daemon_`` with a UID/GID
```
lengau, line 9:
```suggestion
``run-user`` is that it only supports one user called ``_daemon_`` with a UID/GID
```

docs/how-to/rocks/add-internal-user-in-rocks.rst
```diff
@@ -0,0 +1,45 @@
...
+of 84792. For some daemon services that will be enough, but some other services
```
lengau, line 10:
```suggestion
of 84792. For most daemon services that's sufficient, but some services
```

docs/how-to/rocks/add-internal-user-in-rocks.rst
```diff
@@ -0,0 +1,45 @@
...
+require that the running user should have a specific username and/or UID.
+
+In order to add new internal user or group to a rock, three packages (or their
+respective slices) are needed:
+- ``base-files``: to create the base directories such as ``/etc`` and ``/bin``
```
lengau, line 15: Why do these need to be staged? Is it insufficient to simply use `useradd` and `groupadd` to modify the relevant `/etc` files in the overlay?

docs/how-to/rocks/add-internal-user-in-rocks.rst
```diff
@@ -0,0 +1,45 @@
...
+- ``base-passwd``: to produce basic ``/etc/passwd`` and ``/etc/group`` files,
+- ``passwd``: to provide ``useradd`` and ``groupadd`` commands.
```
lengau, line 17: I don't understand why this needs to be staged at all. The `useradd` and `groupadd` commands being used in `override-build` come from the build system, don't they?

docs/how-to/rocks/add-internal-user-in-rocks.rst
```diff
@@ -0,0 +1,45 @@
...
+
+Invoking ``useradd`` and ``groupadd`` command can take place in the
+:ref:`build <lifecycle>` step, by appending the commands to the build flow.
+However, the changes made by those commands should apply to the base system and
```
lengau, line 21:
```suggestion
However, the changes made by those commands normally apply to the base system,
```

docs/how-to/rocks/add-internal-user-in-rocks.rst
```diff
@@ -0,0 +1,45 @@
...
+not the build system. Hence, ``$CRAFT_PART_INSTALL`` should be passed as the
+root system to those two commands.
```
lengau, line 23:
```suggestion
root directory to those two commands.
```

lengau (changes requested):

docs/how-to/rocks/add-internal-user-in-rocks.rst
linostar: this is needed for /root and /home directories. Without /root, the default root user (that has /root as its home) won't be able to run any commands (an error will show because /root does not exist). Similarly, /home is needed to become the base dir of the home dir of the created user(s).

docs/how-to/rocks/add-internal-user-in-rocks.rst
linostar: see comment below.

docs/how-to/rocks/add-internal-user-in-rocks.rst
linostar: That will be tricky, since run-user doesn't support manually-added internal users, and su/sudo don't have available slice definitions yet.

docs/how-to/code/internal-user/rockcraft.yaml
```diff
@@ -0,0 +1,20 @@
...
+parts:
+    my-part:
+        plugin: nil
+        stage-packages:
+            - base-passwd_data
+            - base-files_base
+            - coreutils_bins
+            - passwd_bins
+        override-build: |
+            groupadd -R ${CRAFT_PART_INSTALL} ubuntu
```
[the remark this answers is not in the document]
linostar: done

docs/how-to/rocks/add-internal-user-in-rocks.rst
tigarmo: right but I think @lengau is asking specifically about `passwd` - the build system already has `useradd` and `groupadd`

docs/how-to/rocks/add-internal-user-in-rocks.rst
```diff
@@ -0,0 +1,46 @@
...
+Invoking ``useradd`` and ``groupadd`` command can take place in the
+:ref:`build <lifecycle>` step, by appending the commands to the build flow.
+However, the changes made by those commands normally apply to the base system,
+not the build system. Hence, ``$CRAFT_PART_INSTALL`` should be passed as the
+root directory to those two commands.
+
+The following *rockcraft.yaml* illustrates how to use ``useradd`` command in
+``override-build`` field.
+
+.. literalinclude:: ../code/internal-user/rockcraft.yaml
```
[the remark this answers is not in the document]
linostar: done

docs/how-to/code/internal-user/rockcraft.yaml
```diff
@@ -0,0 +1,20 @@
+name: internal-user
+base: bare
+build-base: ubuntu@24.04
+version: 'latest'
+summary: Add an internal user inside a rock
+description: Add an internal user inside a rock, and run a service with it.
```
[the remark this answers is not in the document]
linostar: added a service to the example

docs/how-to/rocks/add-internal-user-in-rocks.rst
linostar: added a service that uses the field `user`.

docs/how-to/rocks/add-internal-user-in-rocks.rst
linostar: removed passwd_bins slice.

tigarmo (approved):
thanks!

docs/how-to/rocks/add-internal-user-in-rocks.rst
```diff
@@ -0,0 +1,55 @@
...
+the rock's service(s) with. Without using that field in your rockcraft.yaml,
+the service(s) will be run by the root user. The downside, however, of using
+``run-user`` is that it only supports one user called ``_daemon_`` with a
+UID/GID of 84792. For most daemon services that's sufficient, but other
+services require that the running user should have a specific username and/or
+UID.
+
+In order to add a new internal user or group to a rock, two packages (or their
+respective slices) are needed:
+- ``base-files``: to create the base dirs such as ``/root`` and ``/home``,
```
cjdcordeiro, line 16: missing a newline (see the rendered docs)

docs/how-to/rocks/add-internal-user-in-rocks.rst
```diff
@@ -0,0 +1,55 @@
...
+- ``base-passwd``: to produce basic ``/etc/passwd`` and ``/etc/group`` files.
+
+Invoking the ``useradd`` and ``groupadd`` commands can take place in the
+:ref:`build <lifecycle>` step, by appending the commands to the build flow.
```
cjdcordeiro, line 20:
```suggestion
:ref:`build <lifecycle>` step.
```

docs/how-to/rocks/add-internal-user-in-rocks.rst
```diff
@@ -0,0 +1,55 @@
...
+However, the changes made by those commands normally apply to the base system,
```
cjdcordeiro, line 21: This is not clear. The way I read it is:
"if i use `adduser` in the build step, they will affect my base system"

1- that's not the case
2- what's the "base system"?

docs/how-to/rocks/add-internal-user-in-rocks.rst
```diff
@@ -0,0 +1,55 @@
...
+not the build system. Hence, ``$CRAFT_PART_INSTALL`` should be passed as the
+root directory to those two commands.
+
+The following *rockcraft.yaml* illustrates how to use ``useradd`` command in
```
cjdcordeiro, line 25:
```suggestion
The following ``rockcraft.yaml`` illustrates how to use ``useradd`` command in
```

docs/how-to/rocks/add-internal-user-in-rocks.rst
```diff
@@ -0,0 +1,55 @@
...
+``override-build`` field inside a part.
```
cjdcordeiro, line 26:
```suggestion
the ``override-build`` field inside a part.
```

docs/how-to/code/internal-user/rockcraft.yaml
```diff
@@ -0,0 +1,32 @@
...
+    my-part:
+        plugin: nil
+        stage-packages:
+            - base-passwd_data
+            - base-files_base
+            - bash_bins
+            - coreutils_bins
+        override-build: |
+            groupadd --root ${CRAFT_PART_INSTALL} ubuntu
+            useradd -d /home/ubuntu -s /bin/bash --root ${CRAFT_PART_INSTALL} -g ubuntu ubuntu
```
cjdcordeiro, line 31: I think there's too much indentation, when compared to other guides

docs/how-to/rocks/add-internal-user-in-rocks.rst
```diff
@@ -0,0 +1,55 @@
...
+
+.. literalinclude:: ../code/internal-user/rockcraft.yaml
+    :language: yaml
+    :start-after: [docs:rock-parts]
+    :end-before: [docs:rock-parts-end]
+
+Furthermore, let's add a service that tests the running user by executing
```
cjdcordeiro, line 33: This is a very fast context switch.
Also, it's doing more than that. It's **running a service as that user**

docs/how-to/code/internal-user/rockcraft.yaml
cjdcordeiro, line 31: hummm, I have my doubts about this.

Can you please try to run this with an ubuntu base? What happens? Can you print the resulting rock's `/etc/passwd`?

Having said that, I'd probably recommend using a different, nonconflicting, username

docs/how-to/code/internal-user/rockcraft.yaml
```diff
@@ -0,0 +1,32 @@
...
+description: Add an internal user inside a rock, and run a service with it.
+platforms:
+    amd64:
+
+# [docs:rock-services]
+services:
+    whoami:
+        override: replace
+        startup: enabled
+        command: bash -c "whoami; sleep 5"
```
cjdcordeiro, line 15: the sleep is an ugly trick in this case. I suggest running something useful. Can be something that sleeps, intentionally (rather than as a trick to avoid a bad exit code).

And then, as the last step in the doc (which is also way too cumbersome), you can do a `pebble exec --context` to run `whoami`

cjdcordeiro (changes requested):

docs/how-to/rocks/add-internal-user-in-rocks.rst
linostar: Reworded.

docs/how-to/code/internal-user/rockcraft.yaml
linostar: your doubts are about the ubuntu user? it doesn't come with the ubuntu base by default. But I'm changing it to something else anyway.

cjdcordeiro: It does

```bash
$ docker run --rm ubuntu:noble cat /etc/passwd | grep ubuntu
ubuntu:x:1000:1000:Ubuntu:/home/ubuntu:/bin/bash
```

docs/how-to/rocks/add-internal-user-in-rocks.rst
linostar: What do you mean? the context here is about adding a user, then testing that it exists. If we keep it simple, there shouldn't be a step in between.

The service runs whoami to prove that pebble services can be run with the created user as well. There are other form of tests can be done, but this one was like hitting two birds with one stone (or rock) imo.

cjdcordeiro: I mean that the text jumps from the actual "adding a new user" to "running a Pebble service that user", without much context. There's no bridge between the two sections, so it's a very abrupt context switch.

> The service runs whoami to prove that pebble services can be run with the created user as well. There are other form of tests can be done, but this one was like hitting two birds with one stone (or rock) IMO.

Yes, but again, it's lacking the details of what's happening. The important bit is not the `whoami` but the fact that you can enforce the execution of a service to be owned by your user, **regardless of the OCI user**

docs/how-to/code/internal-user/rockcraft.yaml
```diff
@@ -0,0 +1,34 @@
...
+  web-server:
+    override: replace
+    startup: enabled
+    command: python3.12 -m http.server --cgi
+    user: myuser 
+# [docs:rock-services-end]
+
+# [docs:rock-parts]
+parts:
+  my-part:
```
lengau, line 21: I'm wary of using the `nil` plugin in examples just because it doesn't need `craftctl default` somewhere in the overrides.

What about using the `dump` plugin and filtering the staged files to just include `cgi-bin`?

lengau (approved):

docs/how-to/rocks/add-internal-user-in-rocks.rst
```diff
@@ -0,0 +1,65 @@
...
+
+.. literalinclude:: ../code/internal-user/task.yaml
+    :language: bash
+    :start-after: [docs:pack-rock]
+    :end-before: [docs:pack-rock-end]
+    :dedent: 2
+
+We can test the existence of the new user by running the image and accessing
+the url ``http://127.0.0.1/cgi-bin/getuser.py`` with curl. The returned
+response should contain our username:
```
medubelko, line 59:
```suggestion
You can now check which internal user is running the part by running the image
and querying ``http://127.0.0.1/cgi-bin/getuser.py`` with curl:

.. literalinclude:: ../code/internal-user/task.yaml
    :language: bash
    :start-after: [docs:check-user]
    :end-before: [docs:check-user-end]
    :dedent: 2

The response should contain the new user name:

// response here
```

docs/how-to/rocks/add-internal-user-in-rocks.rst
```diff
@@ -0,0 +1,65 @@
...
+
+.. literalinclude:: ../code/internal-user/task.yaml
+    :language: bash
+    :start-after: [docs:check-user]
+    :end-before: [docs:check-user-end]
+    :dedent: 2
```
medubelko, line 65: Per the style guide, we should split inputs and outputs. Would it be possible for you to modify the script so we can read the output as a separate code block?

medubelko (changes requested):
Hi @linostar, I'm the Technical Author assigned to the oversee the documentation efforts of Rockcraft. Thanks for your contribution and for adding to the docs. :) I've left some comments on how to improve the new how-to guide.

docs/how-to/code/internal-user/rockcraft.yaml
linostar: I actually tried the dump plugin first. Not sure if it's a regression but it failed to copy anything (or it copied them but they weren't staged/primed). I am going to investigate that separately.

docs/how-to/rocks/add-internal-user-in-rocks.rst
```diff
@@ -0,0 +1,65 @@
+.. _add_internal_user_in_rocks:
+
+How to add an internal user to a rock
+*************************************
+
+Rockcraft allows you to use :ref:`run-user <rockcraft_yaml_run_user>` to
+specify the user you want to run the rock's service(s) with. Without using that
+field in the ``rockcraft.yaml`` file, the service(s) will be run by the root
+user. However, ``run-user`` only accepts a limited amount of well-known users,
+which could become a constraint for certain container applications.
```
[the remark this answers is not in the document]
linostar: Hi Michael and thank you for chipping in.
I'll be incorporating your changes and responding to some of your comments and suggestions shortly.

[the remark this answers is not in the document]
linostar: "Key" is more suitable to a dictionary or a map object IMHO. "Field" on the other hand a document attributes, like in the case of the rockcraft.yaml file. Furthermore, "field" was the one used throughout the rest of the doc pages, and changing it here would require changing it everywhere else.

docs/how-to/rocks/add-internal-user-in-rocks.rst
```diff
@@ -0,0 +1,65 @@
+.. _add_internal_user_in_rocks:
+
+How to add an internal user to a rock
+*************************************
+
+Rockcraft allows you to use :ref:`run-user <rockcraft_yaml_run_user>` to
+specify the user you want to run the rock's service(s) with. Without using that
+field in the ``rockcraft.yaml`` file, the service(s) will be run by the root
+user. However, ``run-user`` only accepts a limited amount of well-known users,
```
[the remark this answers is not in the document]
linostar: The supported users are listed in the `run-user` section which was referenced in the first line of this paragraph.

docs/how-to/rocks/add-internal-user-in-rocks.rst
```diff
@@ -0,0 +1,65 @@
...
+which could become a constraint for certain container applications.
+
+In order to add a new internal user or group to a rock, two packages (or their
+respective slices) are needed:
+
+- ``base-files``: to create the base dirs such as ``/root`` and ``/home``,
+- ``base-passwd``: to produce basic ``/etc/passwd`` and ``/etc/group`` files.
+
+Invoking the ``useradd`` and ``groupadd`` commands can take place in the
+:ref:`build <lifecycle>` step. However, the changes made by those commands will
```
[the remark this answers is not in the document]
linostar: As per the other doc pages, no. Check https://documentation.ubuntu.com/rockcraft/en/latest/explanation/parts/#build-artefacts for reference.

docs/how-to/rocks/add-internal-user-in-rocks.rst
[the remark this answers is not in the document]
medubelko: OK, let's leave it as "field", then.

docs/how-to/rocks/add-internal-user-in-rocks.rst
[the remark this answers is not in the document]
medubelko: Missed that, my bad. Since `_daemon_` is a [Canonical convention](https://discourse.ubuntu.com/t/unifying-user-identity-across-snaps-and-rocks/36469#naming-convention), let's remove "well-known".

docs/how-to/rocks/index.rst
```diff
@@ -5,6 +5,7 @@
 
    Use Rockcraft's GitHub Action <rockcraft-pack-action>
    Convert an entrypoint to a Pebble layer <convert-to-pebble-layer>
+   Add a new internal user in a rock <add-internal-user-in-rocks>
```
medubelko, line 8:
```suggestion
   Add a new internal user to a rock <add-internal-user-in-rocks>
```

Missed this on my first review. Any chance we can match the file name to the title, too?

docs/how-to/code/internal-user/rockcraft.yaml
```diff
@@ -0,0 +1,34 @@
...
+    source: ./
+    stage-packages:
+      - base-passwd_data
+      - base-files_base
+      - bash_bins
+      - python3.12_standard
+    override-build: |
+      groupadd --root ${CRAFT_PART_INSTALL} mygroup
+      useradd -d /home/myuser -s /bin/bash --root ${CRAFT_PART_INSTALL} -g mygroup myuser
+      cp -r cgi-bin/ ${CRAFT_PART_INSTALL}/cgi-bin
```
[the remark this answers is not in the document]
linostar: Yes.

docs/how-to/rocks/add-internal-user-in-rocks.rst
linostar: the output can be added in task.yal, but I added it explicitly in the tutorial file.

docs/how-to/rocks/add-internal-user-in-rocks.rst
linostar: I changed both the flow and the contents of that part. Furthermore, the service was changed to a python http service.

docs/how-to/code/internal-user/rockcraft.yaml
linostar: The service was changed to a python http service.

docs/how-to/code/internal-user/rockcraft.yaml
```diff
@@ -1,32 +1,36 @@
 name: internal-user
-base: bare
-build-base: ubuntu@24.04
+base: ubuntu@22.04
```
cjdcordeiro, line 2: Why has this changed?

docs/how-to/code/internal-user/rockcraft.yaml
```diff
@@ -1,32 +1,36 @@
...
-        user: ubuntu
-        on-success: ignore 
+  web-service:
+    override: replace
+    command: node server.js
+    startup: enabled
+    on-success: shutdown
+    on-failure: shutdown
+    working-dir: /lib/node_modules/node_web_app
+    user: myuser
```
cjdcordeiro, line 18: Whilst I understand that this is a more proper service to exemplify, is complicates the guide a lot, making it more about Node than about the adding of a user. We need a simpler test service. Or even no service at all, and then just test the rock with `pebble exec --user=muser whoami`. If you have a simple proposal of a basic service that could print this username, without adding too much complexity to this guide, pls share

docs/how-to/code/internal-user/task.yaml
```diff
@@ -12,11 +12,13 @@ execute: |
...
-  id=$(docker run -d internal-user:latest --verbose)
-  docker exec "$id" pebble logs | grep ubuntu
+  id=$(docker run -d -p 8080:8080 internal-user:latest --verbose)
+  sleep 5
+  docker logs "$id" 2> &1 | grep "Serving by myuser"
   docker rm -f "$id"
-  # [docs:check-user-end]
- 
\ No newline at end of file
+  # [docs:check-user-end]
```
cjdcordeiro, line 24: missing newline

cjdcordeiro (changes requested):

docs/how-to/code/internal-user/task.yaml
linostar: you previously requested that change [here](https://github.com/canonical/rockcraft/pull/653/#pullrequestreview-2216211213)

docs/how-to/code/internal-user/rockcraft.yaml
linostar: to accomodate for using the npm plugin (see next conversation for details).

docs/how-to/code/internal-user/task.yaml
cjdcordeiro: ah yeah, that's a diff glitch I guess...I think I suggested removing empty spaces - you had "    ". But I wanted to preserver the newline. GH assumed line deletion though

docs/how-to/code/internal-user/rockcraft.yaml
linostar: Let me recap what I tried so far before picking the node server example above:

1. I tried in `base: bare` a service just "python3.12 -m http.server --cgi" with a simple python script in `cgi-bin` folder that returns the username in a HTTP response. Result: it worked locally, but in spread runner, it seems that the python http server wasn't able to listen on the specified port (tried different ports >= 8000). When curl tried to access the url (something like http://127.0.0.1:8000/cgi-bin/getuser.py) it got "connection reset by peer" error. The docker container was run with `-p 8000:8000`).

2. Next, I tried a python script that use the `http.server` module to launch an HTTP server, but instead of sending the username in an HTTP response which couldn't be tested in a spread runner apparently, I just printed out the username in stdout. Result: nothing was printed out in stdout. Apparently, the bare base with just base-files and base-passwd is missing something that makes python unable to print to stdout in certain contexts (the context of a http service).

3. So as not to overcomplicate the example of the service, I switched to the node example we have in our tutorials. However, after trying it, it doesn't work with bare base neither, so I changed the base accordingly.

cjdcordeiro: (discussed offline - let's try2)

linostar: changed the test to a python http server.

cjdcordeiro (approved):
much cleaner. thanks

medubelko (approved):
Dropped a small change to remove the breakpoint comments in rockcraft.yaml from the rendered page. LGTM.

linostar:
@tigarmo This should be ready for merging.

PR #663, opened 2024-08-05
chore!: migrate to pydantic 2

### LINKED ISSUE #657: Migrate to Pydantic 2
### What needs to get done

Update the models, schema etc to Pydantic 2

### Why it needs to get done

Pydantic 1 is going to EOL eventually

tigarmo:
Many of the changes are just straightforward updates and cleanups. I had to do things like finally dropping the Rockcraft `Platform` model because otherwise I'd have to re-do the work that craft-application already did.

There are some commits here that address things that aren't really pydantic-2 related, unfortunately, but they should be small. Although the commits are reasonably split I would actually suggest reviewing the final diff because some of the commits change things done by earlier ones, so there's a bit of "waste" there.

Fixes #657


rockcraft/models/project.py
```diff
@@ -49,59 +53,17 @@
...
-            raise ValueError("'build-for' expects 'build-on' to also be provided.")
-
-        return values
-
-
-NAME_REGEX = r"^([a-z](?:-?[a-z0-9]){2,})$"
-"""
-The regex for valid names for rocks. It matches the accepted values for pebble
-layer files:
+PROJECT_NAME_REGEX = r"^([a-z](?:-?[a-z0-9]){2,})$"
```
lengau, line 56: Since `max_length` is 40 below, could we do
```suggestion
PROJECT_NAME_REGEX = r"^([a-z](?:-?[a-z0-9]){2,39})$"
```

rockcraft/models/project.py
```diff
@@ -49,59 +53,17 @@
...
+_PROJECT_NAME_DESCRIPTION = """\
+Valid names for rocks. It matches the accepted values for pebble layer files:
 
 - must start with a lowercase letter [a-z]
 - must contain only lowercase letters [a-z], numbers [0-9] or hyphens
 - must not end with a hyphen, and must not contain two or more consecutive hyphens
```
lengau, line 62:
```suggestion
- must not end with a hyphen, and must not contain two or more consecutive hyphens
- between 3 and 40 characters in length
```

rockcraft/models/project.py
```diff
@@ -110,41 +72,56 @@ def _validate_platform_set(cls, values: Mapping[str, Any]) -> Mapping[str, Any]:
...
+]
 
 
 class BuildPlanner(BaseBuildPlanner):
     """BuildPlanner for Rockcraft projects."""
 
     platforms: dict[str, Platform]  # type: ignore[assignment]
     base: Literal["bare", "ubuntu@20.04", "ubuntu@22.04", "ubuntu@24.04"]  # type: ignore[reportIncompatibleVariableOverride]
-    build_base: Literal["ubuntu@20.04", "ubuntu@22.04", "ubuntu@24.04", "devel"] | None  # type: ignore[reportIncompatibleVariableOverride]
+    build_base: BuildBaseT = None  # type: ignore[reportIncompatibleVariableOverride]
```
lengau, line 101: This could also be:

```suggestion
    build_base: Literal["ubuntu@20.04", "ubuntu@22.04", "ubuntu@24.04", "devel"] = pydantic.Field(default=None, validate_default=True)  # type: ignore[reportIncompatibleVariableOverride]
```

I prefer having it separate if we can have a `BaseT` near it or de-duplicate the `ubuntu@` bases

rockcraft/models/project.py
```diff
@@ -110,41 +72,56 @@ def _validate_platform_set(cls, values: Mapping[str, Any]) -> Mapping[str, Any]:
...
         """Build-base defaults to the base value if not specified.
 
         :raises CraftValidationError: If base validation fails.
         """
-        if not build_base:
-            base_value = values.get("base")
+        if not value:
+            base_value = info.data.get("base")
             if base_value == "bare":
                 raise ValueError('When "base" is bare, a build-base must be specified!')
```
lengau, line 115: I know this line is unchanged, but it's perhaps a bit excitable 🙂

rockcraft/models/project.py
```diff
@@ -110,41 +72,56 @@ def _validate_platform_set(cls, values: Mapping[str, Any]) -> Mapping[str, Any]:
...
+    @pydantic.field_validator("base", mode="before")
     @classmethod
     def _validate_deprecated_base(cls, base_value: str | None) -> str | None:
         return cls._check_deprecated_base(base_value, "base")
 
-    @pydantic.validator("build_base", pre=True)
+    @pydantic.field_validator("build_base", mode="before")
     @classmethod
     def _validate_deprecated_build_base(cls, base_value: str | None) -> str | None:
         return cls._check_deprecated_base(base_value, "build_base")
```
lengau, line 127:
```suggestion
        return cls._check_deprecated_base(base_value, "build-base")
```

rockcraft/models/project.py
```diff
@@ -306,7 +280,7 @@ def _check_unsupported_options(cls, values: Mapping[str, Any]) -> Mapping[str, A
 
         return values
 
-    @pydantic.validator("license", always=True)
+    @pydantic.field_validator("license")
```
lengau, line 283: Thoughts on using https://github.com/canonical/craft-application/pull/402 for this?

rockcraft/models/project.py
```diff
@@ -326,30 +300,34 @@ def _validate_license(
...
-        return item
+        if info.data.get("base") != "bare":
+            return parts
 
-    @pydantic.validator("entrypoint_service")
+        for part in parts.values():
+            if part_has_overlay(part):
+                raise ValueError(
+                    'Overlays cannot be used with "bare" bases (there is no system to overlay).'
+                )
```
lengau, line 324: It might be worth mentioning the specific part to the user:

```suggestion
        for name, part in parts.items():
            if part_has_overlay(part):
                raise ValueError(
                    f"Overlays cannot be used with "bare" bases (there is no system to overlay). (in part {name})"
                )
```

Or better, make this a functional validator on one part and make a `Part` annotated type that calls this validator.

mr-cal (approved):

rockcraft/models/project.py
tigarmo: actually I think I copied the `max_length` from craft-application - rocks don't have this restriction afaik. I'll remove it

rockcraft/models/project.py
tigarmo: I think it's a good idea, but as follow-up work

rockcraft/models/project.py
tigarmo: >  and make a Part annotated type that calls this validator

That was my original thought but I couldn't see how to do it without overriding the `parts` declaration in the craft-app Project. Maybe this validator should be upstreamed entirely?

tigarmo: Added the part name as suggested, thanks!

lengau (approved):
Nice work!

sergiusens (approved):

PR #674, opened 2024-08-16
Add fastapi init profile and fastapi-extension

javierdelapuente:

-----

This PR implements the spec ISD160 - 12-Factor FastAPI Support" for rockcraft.

For that, a new init profile called "fastapi-framework" and a new fastapi extension, called "fastapi-framework" are needed. These additions are similar to the flask and django init profiles and extensions, with several differences. Some of the main differences are:

FastAPI and starlette supported.
The ASGI entrypoint location can be in serveral places.
No dependency on gunicorn. FastAPI works with uvicorn without a configuration file.
No statsd.


rockcraft/extensions/fastapi.py
```diff
@@ -0,0 +1,255 @@
...
+
+
+class FastAPIFramework(Extension):
+    """An extension base class for Python FastAPI/starlette framework extension."""
+
+    @staticmethod
+    @override
+    def get_supported_bases() -> Tuple[str, ...]:
+        """Return supported bases."""
+        return "bare", "ubuntu@22.04", "ubuntu:22.04", "ubuntu@24.04"
```
jdkandersson, line 40: perhaps let's not support 22.04, unless you can think of a benefit to the user?

PR #675, opened 2024-08-16
style(lint): replace pylint with ruff

lengau:

-----


linostar (approved):
LGTM, thanks!

rockcraft/commands/init.py
```diff
@@ -16,12 +16,12 @@
 
 """Creation of minimalist rockcraft projects."""
 
+import argparse
```
cmatsuoka, line 19: Ruff doesn't like `TYPE_CHECKING` anymore?

cmatsuoka (approved):

rockcraft/commands/init.py
lengau: I'm using it directly in a method signature, so it needs importing.

lengau:
Moving back to draft so I can:

1. Add ruff in CI
2. Remove pylint
3. Ensure all the pylint linters are enabled in ruff

.github/workflows/tests.yaml
```diff
@@ -17,6 +17,11 @@ jobs:
           submodules: true
       - name: Install packages
         run: |
+          sudo snap install --no-wait ruff
```
mr-cal, line 20: Are you leaving this on 5 lines on purpose?
I thought you could do:
```
sudo snap install --no-wait ruff shellcheck woke
sudo snap install --no-wait --classic node pyright
```

mr-cal (approved):

Makefile
```diff
@@ -89,7 +89,7 @@ test-flake8:
 
 .PHONY: test-ruff
 test-ruff:
-	ruff rockcraft tests
+	ruff check rockcraft tests
```
tigarmo, line 92: since we're dropping `ruff` from the requirements, can we check whether ruff is installed here? With a message on how to install it

setup.cfg
```diff
@@ -63,14 +63,10 @@ test =
...
     mypy
     pydocstyle
-    pylint
-    pylint-fixme-info
-    pylint-pytest>=1.1.3
     pytest
     pytest-check>=2.0
     pytest-mock
     pytest-subprocess
-    ruff==0.3.5
```
tigarmo, line 73: can you please, on a separate commit, re-generate the requirements-* files?

tigarmo (changes requested):
Thanks, this is great ;)

.github/workflows/tests.yaml
lengau: It works for the non-classic ones, but the classic ones need their own lines, so I opted for consistency.

rockcraft/extensions/gunicorn.py
```diff
@@ -201,6 +201,7 @@ def is_experimental(base: str | None) -> bool:
         """Check if the extension is in an experimental state."""
         return False
 
+    # Silencing ruff because of https://github.com/astral-sh/ruff/issues/12942
```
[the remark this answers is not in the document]
lengau: Doubly so as I fixed the issue and there's no ruff silencing :-)

setup.cfg
lengau: A separate commit or a separate PR?

lengau: https://github.com/canonical/rockcraft/pull/678

tigarmo: I meant a separate commit but I don't remember why I asked that 😆 
I think putting that change on this PR would be fine and more correct

lengau:
@tigarmo I was planning on doing #678 once this is merged, but if you'd prefer I can put the requirements file here.

tigarmo:
Just do it here if you can please :)

tigarmo (approved):
a thousand blessings upon you and your house

PR #688, opened 2024-08-29
feat: JLink plugin 

vpa1977:

-----

This PR adds jlink plugin that allows to generate Java runtime tailored for the Java application. 
When ran without supplying jar file property it will search the staging area for any jars and generate an image for them. 
If no jar files are found, then java.base image will be generated. 

This PR includes unit and integration tests. 
The plugin documentation is split into separate PR: https://github.com/canonical/rockcraft/pull/689

Sample ROCK using plugin: https://github.com/canonical/openjdk-rocks/tree/prototype/demos/java-runtime


vpa1977:
Note: for some reason build-the-docs CI fails (404)

vpa1977:
@cjdcordeiro @jdkandersson Hi, I've started on PRs for spring-boot extensions. This one is a prerequisite for it - it adds jlink plugin that builds the java runtime.

jdkandersson:
Perhaps @gregory-schiano would be the best to review this since he has Java knowledge

rockcraft/plugins/jlink_plugin.py
```diff
@@ -0,0 +1,110 @@
...
+        commands.append("mkdir -p ${CRAFT_PART_BUILD}/tmp")
+        # extract jar files into temp folder
+        commands.append(
+            "(cd ${CRAFT_PART_BUILD}/tmp && for jar in ${PROCESS_JARS}; do jar xvf ${jar}; done;)"
+        )
+        commands.append("cpath=$(find ${CRAFT_PART_BUILD}/tmp -type f -name *.jar)")
+        commands.append("cpath=$(echo ${cpath}:. | sed s'/[[:space:]]/:/'g)")
+        commands.append("echo ${cpath}")
+        commands.append(
+            'if [ "x${PROCESS_JARS}" != "x" ]; then deps=$(jdeps --class-path=${cpath} -q --recursive  --ignore-missing-deps --print-module-deps --multi-release 21 ${PROCESS_JARS}); else deps=java.base; fi'
```
gregory-schiano, line 85: Shouldn't `multi-release` use the version specified in the options ?

rockcraft/plugins/jlink_plugin.py
```diff
@@ -0,0 +1,110 @@
...
+        )
+        commands.append(
+            "install_root=${CRAFT_PART_INSTALL}/usr/lib/jvm/java-21-openjdk-${CRAFT_TARGET_ARCH}/"
```
gregory-schiano, line 88: Shouldn't the jdk version match the one specified in the options ?

rockcraft/plugins/jlink_plugin.py
```diff
@@ -0,0 +1,110 @@
...
+        else:
+            commands.append("PROCESS_JARS=$(find ${CRAFT_STAGE} -type f -name *.jar)")
+
+        # create temp folder
+        commands.append("mkdir -p ${CRAFT_PART_BUILD}/tmp")
+        # extract jar files into temp folder
+        commands.append(
+            "(cd ${CRAFT_PART_BUILD}/tmp && for jar in ${PROCESS_JARS}; do jar xvf ${jar}; done;)"
+        )
+        commands.append("cpath=$(find ${CRAFT_PART_BUILD}/tmp -type f -name *.jar)")
```
gregory-schiano, line 81: For coherence shouldn't `cpath` be capitalized ?

rockcraft/plugins/jlink_plugin.py
gregory-schiano: Also shouldn't `install_root` be capitalized ?

rockcraft/plugins/jlink_plugin.py
```diff
@@ -0,0 +1,110 @@
...
+        commands.append(
+            "install_root=${CRAFT_PART_INSTALL}/usr/lib/jvm/java-21-openjdk-${CRAFT_TARGET_ARCH}/"
+        )
+
+        commands.append(
+            "rm -rf ${install_root} && jlink --add-modules ${deps} --output ${install_root}"
+        )
+        # create /usr/bin/java link
+        commands.append(
+            "(cd ${CRAFT_PART_INSTALL} && mkdir -p usr/bin && ln -s --relative usr/lib/jvm/java-21-openjdk-${CRAFT_TARGET_ARCH}/bin/java usr/bin/)"
```
gregory-schiano, line 96: Shouldn't the jdk version match the one specified in the options ?

rockcraft/plugins/jlink_plugin.py
```diff
@@ -0,0 +1,110 @@
...
+        )
+        commands.append("mkdir -p ${CRAFT_PART_INSTALL}/etc/ssl/certs/java/")
+        # link cacerts
+        commands.append(
+            "cp /etc/ssl/certs/java/cacerts ${CRAFT_PART_INSTALL}/etc/ssl/certs/java/cacerts"
+        )
+        commands.append("cd ${CRAFT_PART_INSTALL}")
+        commands.append(
+            "rm -f usr/lib/jvm/java-21-openjdk-${CRAFT_TARGET_ARCH}/lib/security/cacerts"
```
gregory-schiano, line 105: Shouldn't the jdk version match the one specified in the options ?

rockcraft/plugins/jlink_plugin.py
```diff
@@ -0,0 +1,110 @@
...
+        )
+        commands.append(
+            "ln -s --relative etc/ssl/certs/java/cacerts usr/lib/jvm/java-21-openjdk-${CRAFT_TARGET_ARCH}/lib/security/cacerts"
```
gregory-schiano, line 108: Shouldn't the jdk version match the one specified in the options ?

.github/workflows/tests.yaml
```diff
@@ -77,11 +77,14 @@ jobs:
...
         run: |
+          echo "::group::snap install"
+          sudo snap install chisel --edge
+          echo "::endgroup::"
           echo "::group::apt-get update"
           sudo apt-get update
           echo "::endgroup::"
           echo "::group::apt-get install..."
-          sudo apt-get install -y python3 python3-dev libapt-pkg-dev libyaml-dev umoci
+          sudo apt-get install -y python3 python3-dev libapt-pkg-dev libyaml-dev umoci git openjdk-21-jdk
```
gregory-schiano, line 87: Why is openjdk needed since everything related to jdk seems to use chisel ?

gregory-schiano (commented):
Had a first look, didn't review the tests yet

.github/workflows/tests.yaml
vpa1977: OpenJDK is needed on host to create jlink image. 
Chisel is used to install minimal set of dependencies for openjdk.

vpa1977:
Failing spread tests:
```
    - google:ubuntu-22.04-64:tests/spread/rockcraft/overlay-logs
    - google:ubuntu-22.04-64:tests/spread/rockcraft/usrmerge-file-collision
```

rockcraft/plugins/jlink_plugin.py
```diff
@@ -0,0 +1,123 @@
...
+from overrides import override
+
+
+class JLinkPluginProperties(PluginProperties, frozen=True):
+    """The part properties used by the JLink plugin."""
+
+    plugin: Literal["jlink"] = "jlink"
+    jlink_java_version: int = 21
+    jlink_jars: list[str] = []
+    jlink_dep_slices: list[str] = []
```
lengau, line 31: I'm not sure I like this, as it duplicates `stage-packages`. None of the other plugins stage packages like this, and it's probably better to address this in documentation and wait for the resolution of https://github.com/canonical/craft-parts/issues/833

vpa1977: We can not use stage-packages here as the plugin needs to use custom chisel release. 
This can be dropped after Java slices are merged. 

Alternatively, and this is probably a better solution - allow specifying custom chisel-releases for the stage-packages.

vpa1977: I have dropped the property and updated doc MR to include part about explicitly setting up dependencies for the runtime.

lengau:
This is probably worth having a `plugin-jlink` spread test like the [`plugin-python` ones](https://github.com/canonical/rockcraft/tree/main/tests/spread/rockcraft/plugin-python).

rockcraft/plugins/jlink_plugin.py
```diff
@@ -0,0 +1,112 @@
...
+        # create /usr/bin/java link
+        commands.append(
+            "(cd ${CRAFT_PART_INSTALL} && mkdir -p usr/bin && ln -s --relative usr/lib/jvm/java-"
+            + str(options.jlink_java_version)
+            + "-openjdk-${CRAFT_TARGET_ARCH}/bin/java usr/bin/)"
+        )
+        commands.append("mkdir -p ${CRAFT_PART_INSTALL}/etc/ssl/certs/java/")
+        # link cacerts
+        commands.append(
+            "cp /etc/ssl/certs/java/cacerts ${CRAFT_PART_INSTALL}/etc/ssl/certs/java/cacerts"
```
lengau, line 99: (Relevant reading for others coming in: https://github.com/canonical/chisel-releases/pull/308/files)

I'd rather have the docs tell the user to stage `ca-certificates-java_data`, and once we have https://github.com/canonical/craft-parts/issues/833 we can have it automatically incorporate that. What do you think?

rockcraft/plugins/jlink_plugin.py
```diff
@@ -0,0 +1,112 @@
...
+                deps=$(jdeps --class-path=${CPATH} -q --recursive  --ignore-missing-deps \
+                    --print-module-deps --multi-release """
+            + str(options.jlink_java_version)
+            + """ ${PROCESS_JARS}); else deps=java.base; fi
+            """
+        )
+        commands.append(
+            "INSTALL_ROOT=${CRAFT_PART_INSTALL}/usr/lib/jvm/java-"
+            + str(options.jlink_java_version)
+            + "-openjdk-${CRAFT_TARGET_ARCH}/"
```
lengau, line 84:
```suggestion
            f"INSTALL_ROOT=${CRAFT_PART_INSTALL}/usr/lib/jvm/java-{options.jlink_java_version}-openjdk-${CRAFT_TARGET_ARCH}/"
```

rockcraft/plugins/jlink_plugin.py
```diff
@@ -0,0 +1,112 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""The Java runtime plugin."""
```
lengau, line 17:
```suggestion
"""The JLink plugin."""
```

rockcraft/plugins/jlink_plugin.py
```diff
@@ -0,0 +1,112 @@
...
+        commands.append("mkdir -p ${CRAFT_PART_INSTALL}/etc/ssl/certs/java/")
+        # link cacerts
+        commands.append(
+            "cp /etc/ssl/certs/java/cacerts ${CRAFT_PART_INSTALL}/etc/ssl/certs/java/cacerts"
+        )
+        commands.append("cd ${CRAFT_PART_INSTALL}")
+        commands.append(
+            "rm -f usr/lib/jvm/java-"
+            + str(options.jlink_java_version)
+            + "-openjdk-${CRAFT_TARGET_ARCH}/lib/security/cacerts"
```
lengau, line 105:
```suggestion
            f"rm -f usr/lib/jvm/java-{options.jlink_java_version}-openjdk-${CRAFT_TARGET_ARCH}/lib/security/cacerts"
```

lengau: Or given that we use this several times, maybe set

```python
jdk_path = f"usr/lib/jvm/java-{options.jlink_java_version}-openjdk-${{CRAFT_TARGET_ARCH}}"
```

rockcraft/plugins/jlink_plugin.py
vpa1977: Hmm, we can stage ca-certificates-java_data but should avoid priming it. 
ca-certificates-java_data pulls in ca-certificates_data which is not necessary for Java keystore. 
We also need an extra step to generate Java keystore from the staged packages. 

Would not it be easier to use the build environment Java keystore ?

rockcraft/plugins/jlink_plugin.py
vpa1977: We can not do this because we use ${} to escape the environment variable.

vpa1977:
Added PR for the linkcheck linter failure: https://github.com/canonical/rockcraft/pull/700

vpa1977:
> This is probably worth having a `plugin-jlink` spread test like the [`plugin-python` ones](https://github.com/canonical/rockcraft/tree/main/tests/spread/rockcraft/plugin-python).
Thanks!!!  I have added one.

rockcraft/plugins/jlink_plugin.py
lengau: We should try as much as possible not to copy things from the build environment. If the data is different on one architecture for some reason, cross-building to/from that architecture could cause issues. I'm okay with generating the Java keystore from those staged packages though.

vpa1977: Actually this step is unnecessary as jlink already provides correct cacerts.

rockcraft/plugins/jlink_plugin.py
lengau: Sorry, I mistyped this version - put a double-bracket around the environment variables and it'll work properly.

```suggestion
        f"INSTALL_ROOT=${{CRAFT_PART_INSTALL}}/usr/lib/jvm/java-{options.jlink_java_version}-openjdk-${{CRAFT_TARGET_ARCH}}/"
```

vpa1977: Thank you, fixed. I have updated a few other places with string concatenation.

lengau:
I discussed this with my team this week, and the conclusion was that the way we should do this is as an option for the base Java plugin.

So the default behaviour would be the current behaviour (which I *think* is to not stage a JRE?), but the jlink options here would be changed to `<plugin>-jlink-<option>`, and if any of them are set, it would use jlink to stage the JRE. Do you think that's reasonable/feasible?

tests/spread/rockcraft/plugin-jlink/parts.yaml
```diff
@@ -0,0 +1,49 @@
...
+parts:
+  helloworld:
+    source: helloworld
+    source-type: local
+    build-packages:
+      - openjdk-21-jdk
+      - maven
+    plugin: nil
+    build-environment:
+      - JAVA_HOME: "/usr/lib/jvm/java-21-openjdk-${CRAFT_TARGET_ARCH}"
```
lengau, line 22: Under the suggested change about putting this into the java plugins, this would be something we could add in `get_build_environment` (which is still override-able by users if need be)

vpa1977:
> I discussed this with my team this week, and the conclusion was that the way we should do this is as an option for the base Java plugin.
> 
> So the default behaviour would be the current behaviour (which I _think_ is to not stage a JRE?), but the jlink options here would be changed to `<plugin>-jlink-<option>`, and if any of them are set, it would use jlink to stage the JRE. Do you think that's reasonable/feasible?
Hmm, so we will have to enumerate plugin properties in java plugin and retrieve ones with `jlink` key in it?

tests/spread/rockcraft/plugin-jlink/parts.yaml
vpa1977: This is a band-aid for maven. Gradle searches for Java in path. 
The solution in the case would be excluding stage directory from the search path in Java plugin.

vpa1977:
Closing PR - we will implement this functionality in the base java plugin

PR #690, opened 2024-08-29
Implement ux research for flask-framework tutorial

jdkandersson:

-----

Implements the findings from the UX research

### DIFF
--- a/docs/reuse/tutorial/setup.rst
+++ b/docs/reuse/tutorial/setup.rst
@@ -9,7 +9,7 @@ one available, you can create one using `Multipass`_:
 
         .. code-block:: bash
 
-            sudo snap services multipass
+            snap services multipass
 
         If you see the ``multipass`` service but it isn't "active", then you'll
         need to run ``sudo snap start multipass``. On the other hand, if you get
--- a/docs/reuse/tutorial/setup.rst
+++ b/docs/reuse/tutorial/setup.rst
@@ -1,34 +1,32 @@
 We recommend starting from a clean Ubuntu 22.04 installation. If you don't have
 one available, you can create one using `Multipass`_:
 
-.. collapse:: How to create an Ubuntu 22.04 VM with Multipass
+Is Multipass_ already installed and active? Check by running
 
-    Is Multipass_ already installed and active? Check by running
+.. code-block:: bash
 
-    .. code-block:: bash
+    sudo snap services multipass
 
-        sudo snap services multipass
+If you see the ``multipass`` service but it isn't "active", then you'll
+need to run ``sudo snap start multipass``. On the other hand, if you get
+an error saying ``snap "multipass" not found``, then you must install
+`Multipass <install-multipass_>`_:
 
-    If you see the ``multipass`` service but it isn't "active", then you'll
-    need to run ``sudo snap start multipass``. On the other hand, if you get
-    an error saying ``snap "multipass" not found``, then you must install
-    `Multipass <install-multipass_>`_:
+.. code-block:: bash
 
-    .. code-block:: bash
+    sudo snap install multipass
 
-        sudo snap install multipass
+Then you can create the VM with the following command:
 
-    Then you can create the VM with the following command:
+.. code-block:: text
 
-    .. code-block:: text
+    multipass launch --disk 10G --name rock-dev 22.04
 
-        multipass launch --disk 10G --name rock-dev 22.04
+Finally, once the VM is up, open a shell into it:
 
-    Finally, once the VM is up, open a shell into it:
+.. code-block:: bash
 
-    .. code-block:: bash
-
-        multipass shell rock-dev
+    multipass shell rock-dev
 
 ----
 
--- a/docs/reuse/tutorial/setup.rst
+++ b/docs/reuse/tutorial/setup.rst
@@ -1,20 +1,34 @@
 We recommend starting from a clean Ubuntu 22.04 installation. If you don't have
 one available, you can create one using `Multipass`_:
 
-Is Multipass_ already installed and active? Check by running
+.. tabs::
 
-.. code-block:: bash
+    .. tab:: Ubuntu
 
-    sudo snap services multipass
+        Is Multipass_ already installed and active? Check by running
 
-If you see the ``multipass`` service but it isn't "active", then you'll
-need to run ``sudo snap start multipass``. On the other hand, if you get
-an error saying ``snap "multipass" not found``, then you must install
-`Multipass <install-multipass_>`_:
+        .. code-block:: bash
 
-.. code-block:: bash
+            sudo snap services multipass
 
-    sudo snap install multipass
+        If you see the ``multipass`` service but it isn't "active", then you'll
+        need to run ``sudo snap start multipass``. On the other hand, if you get
+        an error saying ``snap "multipass" not found``, then you must install
+        `Multipass <install-multipass_>`_:
+
+        .. code-block:: bash
+
+            sudo snap install multipass
+
+    .. tab:: Windows
+
+        See `Multipass installation instructions <install-multipass_>`_, switch
+        to Windows in the drop down.
+
+    .. tab:: macOS
+
+        See `Multipass installation instructions <install-multipass_>`_, switch
+        to macOS in the drop down.
 
 Then you can create the VM with the following command:
 
--- a/docs/tutorial/flask.rst
+++ b/docs/tutorial/flask.rst
@@ -138,7 +138,7 @@ need to load it into Docker:
     :end-before: [docs:skopeo-copy-end]
     :dedent: 2
 
-Check that the image was successfully imported into Docker:
+Check that the image was successfully loaded into Docker:
 
 .. literalinclude:: code/flask/task.yaml
     :language: bash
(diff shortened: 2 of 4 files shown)


jdkandersson:
@tigarmo could you take a look at the read the docs issue please? It seems that the build isn't working again. It also seems there is a broken link which is why the linting is failing

docs/reuse/tutorial/setup.rst
```diff
@@ -1,34 +1,46 @@
...
+        .. code-block:: bash
 
-    If you see the ``multipass`` service but it isn't "active", then you'll
-    need to run ``sudo snap start multipass``. On the other hand, if you get
-    an error saying ``snap "multipass" not found``, then you must install
-    `Multipass <install-multipass_>`_:
+            sudo snap services multipass
 
-    .. code-block:: bash
+        If you see the ``multipass`` service but it isn't "active", then you'll
```
clay-lake, line 14: I wonder if we can assume that multipass is active and skip to the multipass installation sentence. A user that manually disabled their multipass snap may already know how to correct this issue.

docs/reuse/tutorial/setup.rst
```diff
@@ -1,34 +1,46 @@
...
+        `Multipass <install-multipass_>`_:
 
-        sudo snap install multipass
+        .. code-block:: bash
 
-    Then you can create the VM with the following command:
+            sudo snap install multipass
 
-    .. code-block:: text
+    .. tab:: Windows
```
clay-lake, line 23: The tabs are nice! Out of scope for this PR, but it would be neat if we could select the relevant tab "onload" based on the user agent. This could also work for the multipass docs.

docs/tutorial/flask.rst
```diff
@@ -329,6 +326,12 @@ look like the following:
 Since you are creating a new version of your application, **open the**
 ``rockcraft.yaml`` file and change the ``version`` (e.g. to ``0.2``).
 
+.. note::
+
+    ``rockcraft pack`` will create a new image with the updated code even if you
+    don't change the version. It is recommended to change the version whenever
+    you make changes to the application in the image.
```
clay-lake, line 333: Bumping the version number in the tutorial is good because it demos how software releases integrate with rockcraft, and demos the chisel feature, but I think some may interpret this as "whenever changes are made during development" and not just when software releases are made.

docs/tutorial/flask.rst
```diff
@@ -52,22 +52,11 @@ In the same directory, copy and save the following into a text file called
...
-.. note::
-    The ``&`` at the end of the command runs the Flask application in the
-    background. You can continue to use your terminal as normal and will see all
-    the output from the Flask application in your terminal. To stop the Flask
-    application, you can use the ``kill`` command shown below.
+Run the Flask application using ``flask run -p 8000`` to verify that it works.
 
 Test the Flask application by using ``curl`` to send a request to the root
-endpoint:
+endpoint. You may need a new terminasl for this, if you are using Multipass use
```
erinecon, line 58: Nitpick: "terminasl" --> "terminal"

docs/tutorial/flask.rst
```diff
@@ -99,14 +82,19 @@ creation and tailoring for a Flask application by using the
...
 architecture, include ``arm64`` in ``platforms``.
 
 .. note::
-    For this tutorial, we'll use the ``name`` "flask-hello-world" and build
-    the rock on an ``amd64`` platform.
+    For this tutorial, we'll use the ``name`` ``flask-hello-world`` and assume
+    you are running on an ``amd64`` platform. Check the architecture of your
+    system using ``dpkg --print-architecture``.
+
+.. note::
```
erinecon, line 95: Could these two notes be combined into one note, instead of having them back-to-back?

I do understand that they discuss slightly different things, so I'm not totally opposed to leaving them separated. I'm personally not a huge fan of the back-to-back notes as I think it devalues them a bit.

docs/tutorial/flask.rst
lengau:
```suggestion
endpoint. You may need a new terminal for this, if you are using Multipass use
```

docs/reuse/tutorial/setup.rst
jdkandersson: There is the deliberately disabled part where we could skip it, there is the accidentally disabled and the something went wrong on the machine where it could be useful to have this as a check

docs/reuse/tutorial/setup.rst
jdkandersson: yeh I couldn't find a way to have the link switch correctly to the OS

docs/tutorial/flask.rst
jdkandersson: Fixed

docs/tutorial/flask.rst
jdkandersson: Fixed

docs/reuse/tutorial/setup.rst
```diff
@@ -1,34 +1,46 @@
...
+        Is Multipass_ already installed and active? Check by running
 
-        sudo snap services multipass
+        .. code-block:: bash
 
-    If you see the ``multipass`` service but it isn't "active", then you'll
-    need to run ``sudo snap start multipass``. On the other hand, if you get
-    an error saying ``snap "multipass" not found``, then you must install
-    `Multipass <install-multipass_>`_:
+            sudo snap services multipass
```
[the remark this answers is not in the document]
jdkandersson: Fixed

docs/reuse/tutorial/setup.rst
clay-lake: That is true. In my opinion it comes down to how many people are affected by this issue to keep the tutorial concise. 

If it is a small, but significant fraction of users who have accidentally disabled multi pass, we could add a row to the troubleshooting guide at the bottom. Unfortunately we haven't haven't had many people use multipass in the UX sessions to get a indication how common this problem is.

I just disabled my multipass snap to see what happens:

``` bash

:~$ sudo snap disable multipass 
2024-08-30T09:13:30+02:00 INFO Waiting for "snap.multipass.multipassd.service" to stop.
multipass disabled
:~$ snap services
Service                                              Startup   Current   Notes
...
lxd.user-daemon                                      enabled   inactive  socket-activated
multipass.multipassd                                 disabled  inactive  -
snapd-desktop-integration.snapd-desktop-integration  enabled   active    user
:~$ multipass list
Command 'multipass' not found, but can be installed with:
sudo snap install multipass
:~$ sudo snap install multipass
snap "multipass" is already installed, see 'snap help refresh'
```

docs/tutorial/flask.rst
clay-lake: Perhaps something like this

``` markdown
    ``rockcraft pack`` will overwrite any existing image with the same filename. Updating the version field in the `rockcraft.yaml` will automatically update the filename upon packing, preventing conflicts with prior image versions.
```

jdkandersson: Hmm, that makes it sound like the motivation is around keeping files around, do you think it is more related to releasing?

erinecon (approved):
Thanks, LGTM

docs/reuse/tutorial/setup.rst
jdkandersson: Ok, let's remove this part for now and see if we need it

jdkandersson: I'm looking at the sentence and it seems to just be explaining the output of the `snap services multipass` command. If that sentence is removed, it might not make as much sense. @cjdcordeiro any thoughts please?

clay-lake: @jdkandersson  Yeah, that is true. Is multipass pre-installed on any releases? We could direct the user to install multipass like we do in the `Windows` and `macOS` tabs. 

If multipass was installed on the system beforehand, and the user knows about it, they would probably skip the installation. On the otherhand if multipass is installed and the user does not know it is, `snap install` would return that multipass is already installed and the user would continue forward.

jdkandersson: I don't think multipass is pre-installed at this point

docs/tutorial/flask.rst
clay-lake: I would argue in the tutorial its more about keeping the images so we can compare the chiseled and unchiseled images, but in the context of software development I would only expect the software version to be bumped occasionally.

docs/tutorial/flask.rst
```diff
@@ -77,13 +66,7 @@ endpoint:
...
 The Flask application should respond with ``Hello, world!``.
 
-The Flask application looks good, so you can stop for now:
-
-.. literalinclude:: code/flask/task.yaml
-    :language: bash
-    :start-after: [docs:stop-flask]
-    :end-before: [docs:stop-flask-end]
-    :dedent: 2
+The Flask application looks good, so you can stop for now using ``ctrl+C``.
```
tigarmo, line 69:
```suggestion
The Flask application looks good, so you can stop it for now using ``ctrl+C``.
```

tigarmo (approved):

docs/tutorial/flask.rst
jdkandersson: fixed

PR #692, opened 2024-08-30
Add fastapi init profile and fastapi-framework extension

javierdelapuente:

-----

This PR implements the spec ISD160 - 12-Factor FastAPI Support" for rockcraft.

For that, a new init profile called "fastapi-framework" and a new fastapi extension, called "fastapi-framework" are needed. These additions are similar to the flask and django init profiles and extensions, with several differences. Some of the main differences are:

FastAPI and starlette supported.
The ASGI entrypoint location can be in serveral places (and it is ASGI instead of WSGI).
No dependency on gunicorn. FastAPI works with uvicorn without a configuration file.
No statsd.
Ubuntu 24.04 by default (supported bases 24.04 and bare). 

All current extensions (go, django, flask and now fastapi), have similarities (at the end they extend/change the rockcraft.yaml file), but also many subtle different details, like how they service command is overridden, extra tools or frameworks, what can be overridden and how, extra env variables (only Go).... As so, I have decided to not try to use a common base class for all of them (allowing them to easily evolve independently without a coupled inheritance structure). This is a tradeoff, as code is similar between frameworks. This is open for discussion.


rockcraft/commands/init.py
```diff
@@ -276,6 +276,68 @@ class InitCommand(AppCommand):
...
+                extensions:
+                - fastapi-framework
+
+                # uncomment the sections you need and adjust according to your requirements.
+                # parts:  # you need to uncomment this line to add or update any part.
+
+                #   fastapi-framework/install-app:
+                #     prime:
+                #       # by default, only the files in app.py, app/, src/, "project name/" templates/,
+                #       # static/, migrate, migrate.sh and main.py are copied into the image.
```
jdkandersson, line 323: Also check for `migrate.py`, please check the Flask one as well

rockcraft/extensions/fastapi.py
```diff
@@ -0,0 +1,260 @@
...
+        """Return the normalized name of the rockcraft project."""
+        return self.yaml_data["name"].replace("-", "_").lower()
+
+    def _get_parts(self) -> dict:
+        """Generate the parts associated with this extension."""
+        stage_packages = ["python3-venv"]
+        build_environment = []
+        if self.yaml_data["base"] == "bare":
+            stage_packages = ["python3.12-venv_ensurepip"]
+            build_environment = [{"PARTS_PYTHON_INTERPRETER": "python3.12"}]
```
jdkandersson, line 95: Check whether this is needed

rockcraft/extensions/fastapi.py
```diff
@@ -0,0 +1,260 @@
...
+
+    def _check_project(self) -> None:
+        """Ensure this extension can apply to the current rockcraft project."""
+        error_messages = self._requirements_txt_error_messages()
+        if not self.yaml_data.get("services", {}).get("fastapi", {}).get("command"):
+            error_messages += self._asgi_entrypoint_error_messages()
+        if error_messages:
+            raise ExtensionError(
+                "\n".join("- " + message for message in error_messages),
+                doc_slug="/reference/extensions/go-framework",
```
jdkandersson, line 229: Check this

tests/unit/extensions/test_fastapi.py
jdkandersson, line 1: Could we add a spread test please?

rockcraft/extensions/fastapi.py
javierdelapuente: It is not necessary. Removed

tests/unit/extensions/test_fastapi.py
javierdelapuente: Added

javierdelapuente:
Hi @tigarmo can you have a look at this PR? Thanks!

jdkandersson (approved):

lengau (approved):

PR #694, opened 2024-09-02
tests(spread): use LXD from `5.21/candidate`

tigarmo:
This serves two purposes:

a) It un-breaks `main` as `5.21/candidate` has the fixes for the regressions
   when mounting the overlay filesystems;
b) It lets the spread tests act as a line of defense against further LTS
   regressions, without being too exposed to breakage by using something like
   `latest/edge`.

-----

### DIFF
--- a/spread.yaml
+++ b/spread.yaml
@@ -62,8 +62,9 @@ prepare: |
     tests.pkgs remove lxd
   fi
 
-  # 5.21/stable is the latest stable LTS
-  snap install lxd --channel=5.21/stable
+  # 5.21/stable is the latest stable LTS. Use 5.21/candidate as a "first line of
+  # defense" against LXD regressions.
+  snap install lxd --channel=5.21/candidate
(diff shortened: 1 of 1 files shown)


mr-cal (approved):

lengau (approved):

spread.yaml
```diff
@@ -62,8 +62,9 @@ prepare: |
     tests.pkgs remove lxd
   fi
 
-  # 5.21/stable is the latest stable LTS
-  snap install lxd --channel=5.21/stable
+  # 5.21/stable is the latest stable LTS. Use 5.21/candidate as a "first line of
+  # defense" against LXD regressions.
+  snap install lxd --channel=5.21/candidate
```
lengau, line 67: Does lxd use their beta channel?

tigarmo: looks like it doesn't:
image

PR #702, opened 2024-09-09
adjust the guide "convert entrypoint to pebble layer"

linostar:

-----

This introduces some minor adjustements to the "convert entrypoint to pebble layer" so it is easier to understand and allow it to reproduce the results of the reference image more accurately.


docs/how-to/code/convert-to-pebble-layer/rockcraft.yaml
```diff
@@ -38,14 +38,14 @@ services:
   nginx:
     override: replace
     startup: disabled
-    command: nginx [ -h ]
+    command: nginx -g 'daemon off;'
```
cjdcordeiro, line 41: if these are not optional, they cannot be changed from the CMD

cjdcordeiro (changes requested):

.readthedocs.yaml
```diff
@@ -12,7 +12,7 @@ build:
   jobs:
     post_checkout:
       - git fetch --tags --depth 1 # Also fetch tags
-      - git describe               # Useful for debugging
+      - git describe --always      # Useful for debugging
```
tigarmo, line 15: what's this?

docs/how-to/rocks/convert-to-pebble-layer.rst
```diff
@@ -125,16 +125,22 @@ And finally, run the container:
...
 
-Notice the given command ``start nginx``, as this is Pebble's client syntax to
-let the Pebble daemon know that the ``nginx`` service defined in
-``rockcraft.yaml`` (which is disabled by default) should be enabled at startup.
-Otherwise, the Pebble daemon would start without any NGINX service, although
-you could still later on ask for that service to be started (via something like
-``docker exec <container-name> start nginx``).
+The Pebble daemon will start without any NGINX service, although you could
+still later on ask for either service to be started. For instance, you can
+start ``nginx`` service by typing:
```
tigarmo, line 130:
```suggestion
start the ``nginx`` service by typing:
```

tigarmo (changes requested):

.readthedocs.yaml
linostar: Few recent PRs, like this one and PR #692, were failing to build docs because `git describe` command is not finding a tag to describe. This is a workaround.

tigarmo: But this is a problem that we need to fix some other way; if the `git describe` failed then the version of rockcraft on the docs will be wrong

linostar: But what is causing it to fail all of a sudden?

tigarmo: I don't know, nothing changed on our side as far as I can tell

tigarmo: I fixed this on `main`, you can undo this change now

linostar: done

tigarmo (approved):

cjdcordeiro (approved):

PR #703, opened 2024-09-10
Exclude db.sqlite3 file in Django framework extension

javierdelapuente:

-----

This file can be created locally if in the host `python manage.py migrate` is run.

The local sqlite3 database is not needed not desired for the rock.


rockcraft/extensions/gunicorn.py
```diff
@@ -339,6 +339,7 @@ def gen_install_app_part(self) -> dict[str, Any]:
                 "plugin": "dump",
                 "source": self.name,
                 "organize": {"*": "django/app/", ".*": "django/app/"},
+                "stage": ["-django/app/db.sqlite3"],
```
tigarmo, line 342: @javierdelapuente can you tell me what this file is and why we're excluding it?

javierdelapuente: By default, when you run `django-admin startproject` , it will create a project using as a database sqlite3. Something like this will be generated in the initial templates:
```
DATABASES = {
    'default': {
        'ENGINE': 'django.db.backends.sqlite3',
        'NAME': BASE_DIR / 'db.sqlite3',
    }
}
```

If the user runs the `python3 manage.py migrate` in the host, it will create a `db.sqlite3` file with the database, and may be filled with data by the application. That file will then be included in the rock, which is something undesired.

This is also related to this [PR](https://github.com/canonical/paas-app-charmer/pull/42), as we do not support sqlite3 in 12 factor apps, because it is in ephemeral storage and because it will break HA applications (different sqlite3 database for each unit)

tigarmo: I see, thanks!

tigarmo (approved):

PR #712, opened 2024-09-24
build(deps): update deps with freeze-requirements.sh

tigarmo:
This PR brings in the most recent craft-* versions, with some changes and schema updates. Each change is in its own commit.


schema/rockcraft.json
```diff
@@ -3795,6 +3806,7 @@
                   "meson",
                   "nil",
                   "npm",
+                  "poetry",
```
lengau, line 3809: Since we're disabling poetry do we want to temporarily remove it from the JSON? Or is that too much work for something we're about to re-enable?

tigarmo: it's not disabled on rockcraft, so adding it to the schema here seems 'less wrong' heh

lengau: Ohh, I thought it was. My bad. Thanks!

mattculler (approved):

PR #715, opened 2024-09-25
refactor: use craft-platform for the build plans

### LINKED ISSUE #698: Use craft-platforms for build plans
### What needs to get done

Use craft-platforms for creating build plans

See prototype here: https://github.com/canonical/rockcraft/pull/662

### Why it needs to get done

* To deduplicate similar build planning logic
* To ensure consistent build plans between rockcraft and consumers of craft-platforms, like Launchpad

tigarmo:
The semantics are the same; the goal is to have a single source of build plans that is used both by Rockcraft and by whatever other code (e.g. remote builds) might need it.

Fixes #698

-----

### DIFF
--- a/rockcraft/models/project.py
+++ b/rockcraft/models/project.py
@@ -24,19 +24,22 @@ from pathlib import Path
 from typing import TYPE_CHECKING, Any, Literal
 
 import craft_cli
+import craft_platforms
 import pydantic
 import spdx_lookup  # type: ignore
 import yaml
 from craft_application.errors import CraftValidationError
 from craft_application.models import (
-    BuildPlanner as BaseBuildPlanner,
-)
-from craft_application.models import (
+    BuildInfo,
     Platform,
     get_validator_by_regex,
 )
+from craft_application.models import (
+    BuildPlanner as BaseBuildPlanner,
+)
 from craft_application.models import Project as BaseProject
 from craft_application.models.base import alias_generator
+from craft_platforms import rock
 from craft_providers import bases
 from craft_providers.errors import BaseConfigurationError
 from typing_extensions import Annotated, override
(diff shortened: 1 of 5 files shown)


rockcraft/models/project.py
```diff
@@ -219,6 +222,28 @@ def effective_base(self) -> bases.BaseName:
     def model_reference_slug(cls) -> str | None:
         return "/reference/rockcraft.yaml"
 
+    @override
+    def get_build_plan(self) -> list[BuildInfo]:
+        platforms = typing.cast(
+            # https://github.com/canonical/craft-platforms/issues/43
```
mr-cal, line 228: Since this is fixed, I think we can drop this with a release of craft-platforms.

rockcraft/models/project.py
```diff
@@ -219,6 +222,28 @@ def effective_base(self) -> bases.BaseName:
...
+        )
+        build_infos = rock.get_rock_build_plan(
+            base=self.base,
+            build_base=self.build_base,
+            platforms=platforms,
+        )
+        return [
+            BuildInfo(
+                platform=info.platform,
+                build_on=str(info.build_on),
```
mr-cal, line 240: If we update craft-application's `build-on` and `build-for` to `str | DebianArchitecture`, we can drop this, right?

mr-cal (approved):
Looks good - you can ignore my open-ended questions

rockcraft/models/project.py
tigarmo: Either that, or we could have a function in craft-application that converts the buildinfos

lengau: My WIP craft-application change will obviate this with a `from_platforms()` constructor on its BuildInfo. If we want these QoL changes quickly for building the rockcraft/snapcraft versions on I could chop that out into its own PR and you can test from edge?

(The other thing it does is marks `craft_application.BuildInfo` as deprecated, pointing to `craft_platforms`, but that wouldn't make it into this PR since craft-application still uses its own `BuildInfo` everywhere and that'll be a breaking change.)

tigarmo: > If we want these QoL changes quickly

Nah I don't think that's necessary, we can take it slow

lengau (approved):

PR #717, opened 2024-09-26
feat: add Rockcraft-specific Poetry plugin

### LINKED ISSUE #701: Enable the poetry plugin
### What needs to get done

Enable the poetry plugin with the quirks as of the python plugin

### Why it needs to get done

To be able to use poetry based projects as a source

tigarmo:
The plugin has the same behavior and restrictions as the Python plugin with regards to base-dependent behavior, symlink handling, sitecustomize, etc.

Therefore, this common behavior is extracted into a new "python_common" module, used by both plugins. This approach is also taken for the reference docs - the requirement of staging a Python interpreter is the same for both plugins.

Fixes #701

-----

### DIFF
--- a/docs/reference/plugins/_python_common.rst
+++ b/docs/reference/plugins/_python_common.rst
@@ -1,0 +1,28 @@
+
+Dependencies
+------------
+
+Since none of the bases that are available for rocks contain a default Python
+installation, including a Python interpreter in Rockcraft projects is mandatory.
+Both the ``python`` and the ``poetry`` plugins also require the ``venv`` module
+to create the virtual environment where Python packages are installed at build
+time.
+
+The easiest way to do this is to include the ``python3-venv`` package in the
+``stage-packages`` of the part that uses the Python-based plugin. This will pull
+in the default Python interpreter for the ``build-base``, like Python 3.10 for
+Ubuntu 22.04. However, other versions can be used by explicitly declaring them -
+here's an example that uses ``python3.12-venv`` from the Deadsnakes ppa:
+
+.. code-block:: yaml
+
+   package-repositories:
+     - type: apt
+       ppa: deadsnakes/ppa
+       priority: always
+
+   parts:
+     my-part:
+       plugin: <python or poetry>
+       source: .
+       stage-packages: [python3.12-venv]
--- a/rockcraft/plugins/python_common.py
+++ b/rockcraft/plugins/python_common.py
@@ -1,0 +1,131 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright 2024 Canonical Ltd.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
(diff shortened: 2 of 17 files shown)


docs/reference/plugins/_python_common.rst
```diff
@@ -0,0 +1,28 @@
+
+Dependencies
```
tigarmo, line 2: This part is mostly unchanged, some minor bits to talk about both plugins instead of just the `python` one

rockcraft/plugins/python_common.py
```diff
@@ -0,0 +1,131 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Common functionality for Python-based plugins.
```
tigarmo, line 17: This is taken from `python_plugin.py`, with no semantic changes

mr-cal (approved):

tests/integration/plugins/test_python_plugin.py
```diff
@@ -106,9 +109,10 @@ class ExpectedValues:
     VALUES_FOR_HOST = RELEASE_TO_VALUES["22.04"]
 
 
+@pytest.mark.parametrize("plugin_name", ["python", "poetry"])
 @pytest.mark.parametrize("base", tuple(UBUNTU_BASES))
-def test_python_plugin_ubuntu(base, tmp_path, run_lifecycle):
-    project = create_python_project(base=base)
+def test_python_plugin_ubuntu(base, tmp_path, run_lifecycle, plugin_name):
```
lengau, line 114: This is very creative, but not the dangerous sort of "clever code" I fear. Very nice!

tests/spread/rockcraft/plugin-poetry/task.yaml
```diff
@@ -0,0 +1,26 @@
...
+  docker run --rm $IMAGE exec /usr/bin/python3 -m test_poetry | MATCH "Hello from Poetry"
+  docker run --rm $IMAGE exec python3 -m test_poetry | MATCH "Hello from Poetry"
+  
+  # Run the extra Python package, installed as a dependency, to make sure it's found
+  docker run --rm $IMAGE exec black --version
+  docker run --rm $IMAGE exec /bin/python3 -m black --version
+  docker run --rm $IMAGE exec /usr/bin/python3 -m black --version
+  docker run --rm $IMAGE exec python3 -m black --version
+
+  docker system prune -a -f
```
lengau, line 26: Doesn't this really belong in the `restore` part (in case one of the Docker commands fails)?

Or perhaps even in the `spread.yaml`'s `restore-each`?

lengau (approved):
Looks good! The suggestion doesn't need to be done here, and can be rejected altogether if you don't like it.

PR #725, opened 2024-10-04
fix: set PATH on bare-based rocks

### LINKED ISSUE #711: Set PATH on the rock for bare bases
### What needs to get done

When packing a rock with a `bare` base, set the `PATH` environment variable.

### Why it needs to get done

For security - an empty PATH will mean the implicitly addition of the current working directory.

tigarmo:
If the PATH is empty, Pebble will be default use the standard Ubuntu value
of "/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin" (see [0]).
Follow this lead and set this value on the image's PATH.

This PATH setting on the image itself has no bearing on most cases, as the
PATH that prevails is the one defined by Pebble and its services, but an
empty (or unset) PATH is a potential security issue in cases where the pebble
entrypoint is bypassed.

0: https://github.com/canonical/pebble/blob/master/internals/overlord/cmdstate/request.go#L91

Fixes #711

### DIFF
--- a/rockcraft/oci.py
+++ b/rockcraft/oci.py
@@ -389,6 +389,21 @@ class Image:
         _config_image(image_path, cmd_params)
         emit.progress(f"CMD set to {opt_args}")
 
+    def set_default_path(self, base: str) -> None:
+        """Set the default PATH on the image (only for bare rocks)."""
+        if base != "bare":
+            emit.debug(f"Not setting a PATH on the image as base is {base!r}")
+            return
+
+        # Follow Pebble's lead here: if PATH is empty, use the standard one.
+        # This means that containers that bypass the pebble entrypoint will
+        # have the same behavior as PATH-less pebble services.
+        pebble_path = Pebble.DEFAULT_ENV_PATH
+        image_path = self.path / self.image_name
+
+        emit.debug(f"Setting bare-based rock PATH to {pebble_path!r}")
+        _config_image(image_path, ["--config.env", f"PATH={pebble_path}"])
+
     def set_pebble_layer(
         self,
         services: dict[str, Any],
--- a/rockcraft/pebble.py
+++ b/rockcraft/pebble.py
@@ -151,6 +151,9 @@ class Pebble:
         **_BASE_PART_SPEC,
         "stage": [PEBBLE_BINARY_PATH_PREVIOUS],
     }
+    # This is the value that Pebble sets to the PATH env var if it's empty.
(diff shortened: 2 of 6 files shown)


lengau (approved):

rockcraft/oci.py
```diff
@@ -389,6 +389,22 @@ def set_cmd(self, command: str | None = None) -> None:
         _config_image(image_path, cmd_params)
         emit.progress(f"CMD set to {opt_args}")
 
+    def set_path(self, base: str, build_base: str | None) -> None:
+        """Set the default PATH on the image (only for bare rocks)."""
+        if base != "bare":
+            emit.debug(f"Not setting a PATH on the image as base is {base!r}")
+            return
```
cjdcordeiro, line 396: I think we could make this inference a bit more sustainable, as it could break if one day we have other bases without a `PATH` or we redefine the meaning of `bare` and actually serve `bare` as an existing OCI artefact, with Pebble and other settings (like `PATH`) already inside.

What about getting the base's existing `env` from `{image_path}/config.json`?

rockcraft/oci.py
```diff
@@ -389,6 +389,22 @@ def set_cmd(self, command: str | None = None) -> None:
...
+
+        # Follow Pebble's lead here: if PATH is empty, use the standard one.
+        # This means that containers that bypass the pebble entrypoint will
+        # have the same behavior as PATH-less pebble services.
+        # (https://github.com/canonical/pebble/blob/master/internals/overlord/cmdstate/request.go#L91)
+        pebble_path = "/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
```
cjdcordeiro, line 402: nit: should this be defined in the `pebble` module?

cjdcordeiro (changes requested):
Nice idea with the default Pebble path. And nice parameterization of the tests.

Just left a couple of small comments.

rockcraft/oci.py
cjdcordeiro: Sure. Please add it as a good-first-issue. I'd love for it to be done by new joiners.

cjdcordeiro (approved):

rockcraft/oci.py
tigarmo: done, thanks!

mr-cal (approved):

rockcraft/oci.py
tigarmo: https://github.com/canonical/rockcraft/issues/729

PR #728, opened 2024-10-08
feat: maven rockcraft plugin

vpa1977:
Add overrides for craft-part maven plugin to allow building Maven projects in rockcraft.

Namely:
 - Do not link /bin/java as it conflicts with the base-files_base slice

The PR requires https://github.com/canonical/craft-parts/pull/895

-----


vpa1977:
Needs ci retry for lint task: `/home/runner/work/rockcraft/rockcraft/docs/reuse/tutorial/setup.rst:1: WARNING: broken link: https://multipass.run/docs/how-to-install-multipass (500 Server Error:  for url: https://multipass.run/docs/how-to-install-multipass)
`

tests/spread/rockcraft/plugin-maven/rockcraft.yaml
lengau, line 1: Could you update the [maven plugin docs](https://documentation.ubuntu.com/rockcraft/en/stable/common/craft-parts/reference/plugins/maven_plugin/) for this and include the jlink example?

vpa1977: I have updated the docs, but I do not think the example is a good one - namely it does not take application's imported modules into account. Adding jdeps steps will result in a complex shell script. 
I would prefer to reference JLink plugin https://github.com/canonical/rockcraft/pull/688 once it becomes available.

vpa1977:
Note: linter exception in the logs is resolved by https://github.com/canonical/rockcraft/pull/732

lengau (approved):

tests/spread/rockcraft/plugin-maven/rockcraft.yaml
```diff
@@ -0,0 +1,32 @@
...
+    stage-packages:
+     - openjdk-21-jre-headless_core
+     - base-files_base
+    build-packages:
+      - maven
+      - openjdk-21-jdk-headless
+    # replace chiselled image modules with jlink output
+    # in order to ensure that only java.base module is present in
+    # runtime
+    override-stage: |
```
tigarmo, line 26: how come this is done in `stage`? Can you do it in `build` instead?

rockcraft/plugins/maven_plugin.py
```diff
@@ -0,0 +1,76 @@
...
+).strip()
+
+
+class MavenPlugin(maven_plugin.MavenPlugin):
+    """A MavenPlugin plugin for Rockcraft.
+
+    This plugin extends Craft-parts' vanilla Maven plugin to properly
+    use and install the Java VM. Specifically:
+
+    - Do not link java executable to /bin/java
```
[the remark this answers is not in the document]
vpa1977: The /bin itself is a link to /usr/bin. 
Maven plugin in craft-parts tries to create /bin/java before /bin -> /usr/bin link is established creating a conflict.

rockcraft/plugins/maven_plugin.py
```diff
@@ -0,0 +1,76 @@
...
+"""The Rockcraft Maven plugin."""
+import logging
+from textwrap import dedent
+
+from craft_parts.plugins import maven_plugin
+from overrides import override  # type: ignore[reportUnknownVariableType]
+
+logger = logging.getLogger(__name__)
+
+# Template for the sitecustomize module that we'll add to the payload so that
```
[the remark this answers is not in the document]
vpa1977: This is a boilerplate in rockcraft plugins which I have copied.

tests/spread/rockcraft/plugin-maven/rockcraft.yaml
vpa1977: Notice 
```
    stage-packages:
     - openjdk-21-jre-headless_core
```
lib/modules have to be replaced after `openjdk-21-jre-headless_core` is staged. Hence this is done in stage.

tests/spread/rockcraft/plugin-maven/rockcraft.yaml
```diff
@@ -0,0 +1,32 @@
+name: plugin-maven
+base: bare
+build-base: ubuntu@24.04
+
+version: '0.1'
+summary: A rock that bundles a Maven project.
+description: A rock that bundles a Python project.
```
tigarmo, line 7:
```suggestion
description: A rock that bundles a Maven project.
```

tests/spread/rockcraft/plugin-maven/rockcraft.yaml
```diff
@@ -0,0 +1,32 @@
+name: plugin-maven
```
tigarmo, line 1: Thanks for the test. I'm trying it out here with Rockcraft 1.5.3 and this project builds fine without the changes from this PR. Do you know what's going on? Ideally we'd test a case that can only work with this fix.

rockcraft/plugins/maven_plugin.py
```diff
@@ -0,0 +1,76 @@
...
+
+    @override
+    def _get_java_link_commands(self) -> list[str]:
+        """Get the bash commands to provide /bin/java symlink."""
+        return []
+
+    @override
+    def get_build_environment(self) -> dict[str, str]:
+        env = super().get_build_environment()
+        env["PATH"] = "/usr/bin"
```
tigarmo, line 75: This is a pretty big change that we should only do if there's absolutely no other way around it. The `PATH` that craft-parts builds prioritizes the install and stage dirs on purpose, to let you do things like using a custom jdk in your build.

This is needed because we need Maven to pick the build-package jdk instead of the staged jre, right? Could we configure `JAVA_HOME` somehow? Like, to `"$(dirname $(dirname $(readlink -f $(which javac))))"` (the directory that contains `bin/javac`)?

tigarmo (changes requested):
@vpa1977 thanks for the PR. The main thing that's blocking me here is that the spread test project builds fine for me without the changes, which is weird because I've seen the issues that your code aims to fix in the past too.

Since you've worked with this plugin more recently, do you know of a project that would definitely not work without this rockcraft-custom plugin?

tests/spread/rockcraft/plugin-maven/rockcraft.yaml
vpa1977: This will override jlink-generated module with the staged one:
```
spring-jl17      latest          56cb9c14956a   About an hour ago   348MB
spring           latest          b2ad2b9fe0e1   About an hour ago   186MB
```
Notice 348MB image ...

tigarmo: But how can that be? The `stage-packages` are extracted into `CRAFT_PART_INSTALL`  *before*  the build step is executed

vpa1977: Is there any chance of non-deterministic behaviour there? I will try to build with --shell-after and check the contents of `CRAFT_PART_INSTALL`

tigarmo: no, this part should be pretty well-established; many plugins rely on the fact that stage-packages are already extracted when the build is executed.

vpa1977:
> @vpa1977 thanks for the PR. The main thing that's blocking me here is that the spread test project builds fine for me without the changes, which is weird because I've seen the issues that your code aims to fix in the past too.
> 
> Since you've worked with this plugin more recently, do you know of a project that would definitely not work without this rockcraft-custom plugin?

Please refer to CI run https://github.com/canonical/rockcraft/actions/runs/11447589661/job/31850127130?pr=738 of https://github.com/canonical/rockcraft/pull/738. This draft PR adds only test and it fails building maven project second time.

tests/spread/rockcraft/plugin-maven/rockcraft.yaml
vpa1977: See https://github.com/canonical/rockcraft/pull/728#issuecomment-2428209279

rockcraft/plugins/maven_plugin.py
vpa1977: I think there are two different problems:
 - maven project does not build because selected toolchain is wrong. Auto detected java home solves this.  
 - part build is not reproducible and the build output is not what I expected as a user, because the the build pick up different toolchains. The edge case - project fails to build at all. Limiting path to /usr/bin solves this, but not in line with other part plugins environment settings.   Another approach would be to sandbox staging directories to each part and ensure that it can not access output of its own staging phase.

tests/spread/rockcraft/plugin-maven/rockcraft.yaml
vpa1977: Ah, ok, this one is my fault - somehow in the image I've got a full install of openjdk-17-jre-headless and slices from openjdk-21-jre-headless.
I guess the issue was "not enough sleep".

rockcraft/plugins/maven_plugin.py
vpa1977: Maybe as a bandaid just put /usr/bin in front of the PATH?

vpa1977: Maybe as a bandaid just put /usr/bin in front of the PATH?

tigarmo:
> > @vpa1977 thanks for the PR. The main thing that's blocking me here is that the spread test project builds fine for me without the changes, which is weird because I've seen the issues that your code aims to fix in the past too.
> > Since you've worked with this plugin more recently, do you know of a project that would definitely not work without this rockcraft-custom plugin?
> 
> Please refer to CI run https://github.com/canonical/rockcraft/actions/runs/11447589661/job/31850127130?pr=738 of #738. This draft PR adds only test and it fails building maven project second time.

Ah I see, I missed the part where the second execution of `rockcraft pack` was the one that failed. Thanks

tests/spread/rockcraft/plugin-maven/task.yaml
```diff
@@ -0,0 +1,23 @@
...
+  run_rockcraft pack
+  ls
+  echo expecting ${ROCK_FILE}
+  test -f ${ROCK_FILE}
+  rockcraft.skopeo copy --insecure-policy  oci-archive:${ROCK_FILE} docker-daemon:${IMAGE}
+  docker images
+  rm ${ROCK_FILE}
+
+  # Run the packaged project
+  docker run --rm $IMAGE exec /usr/lib/jvm/java-21-openjdk-amd64/bin/java -jar /jar/sample-0.0.1-SNAPSHOT.jar | MATCH "hello world"
```
tigarmo, line 17:
```suggestion
  docker run --rm $IMAGE exec /usr/bin/java -jar /jar/sample-0.0.1-SNAPSHOT.jar | MATCH "hello world"
```

rockcraft/plugins/maven_plugin.py
```diff
@@ -0,0 +1,44 @@
...
+
+    @override
+    def _get_java_link_commands(self) -> list[str]:
+        """Get the bash commands to provide /bin/java symlink."""
+        return []
+
+    @override
+    def get_build_environment(self) -> dict[str, str]:
+        env = super().get_build_environment()
+        env["PATH"] = "/usr/bin:$PATH"
```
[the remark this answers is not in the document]
vpa1977: The JAVA_HOME variable in this case points to the first JDK found on PATH. 
Consider the following use-case:
- My application is a legacy Java 8 application, but the test harness is modern Java 21. 
- My application requires JDK to run (e.g. compiles code). 
- I deploy JDK 8 to the container to run my application

Should we automatically detect Java home, the second time the build runs it will return Java 8 and my tests will fail to compile.

vpa1977:
As discussed during sprint, I am dropping part that restricts the path and will provide the fix as PR in for `craft-parts` to set the JAVA_HOME. 
This PR will remove hardlink part which will make the maven plugin usable with rockcraft.

rockcraft/plugins/maven_plugin.py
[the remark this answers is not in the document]
vpa1977: As discussed, it is handled by https://github.com/canonical/craft-parts/issues/892

vpa1977:
The PR depends on https://github.com/canonical/craft-parts/pull/895

tests/spread/rockcraft/plugin-maven/rockcraft.yaml
```diff
@@ -0,0 +1,32 @@
...
+    # replace chiselled image modules with jlink output
+    # in order to ensure that only java.base module is present in
+    # runtime
+    override-build: |
+      craftctl default
+      rm -rf out
+      jlink --no-header-files --no-man-pages --strip-debug \
+        --add-modules java.base \
+        --output out
+      cp out/lib/modules ${CRAFT_PART_INSTALL}/usr/lib/jvm/java-21-openjdk-${CRAFT_ARCH_BUILD_FOR}/lib
```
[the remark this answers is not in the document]
vpa1977: This is a file

tests/spread/rockcraft/plugin-maven/rockcraft.yaml
```diff
@@ -0,0 +1,32 @@
...
+    plugin: maven
+    source: sample
+    maven-parameters: [ "-B" ]
+    stage-packages:
+     - openjdk-21-jre-headless_core
+     - base-files_base
+    build-packages:
+      - maven
+      - openjdk-21-jdk-headless
+    # replace chiselled image modules with jlink output
```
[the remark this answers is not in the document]
vpa1977: Notice:
```
    stage-packages:
     - openjdk-21-jre-headless_core
 ```
This adds `${CRAFT_PART_INSTALL}/usr/lib/jvm/java-21-openjdk-${CRAFT_ARCH_BUILD_FOR}` as far as I understand. 
`cp out/lib/modules ${CRAFT_PART_INSTALL}/usr/lib/jvm/java-21-openjdk-${CRAFT_ARCH_BUILD_FOR}/lib`
replaces the module file with the one generated by jlink.

**Comment by vpa1977** [comment]
Moving this one to draft as it depends on JAVA_HOME PR in craft parts

File: tests/spread/rockcraft/plugin-maven/rockcraft.yaml (hunk 10 above)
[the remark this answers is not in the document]
Comment (tigarmo) [reply tests/spread/rockcraft/plugin-maven/rockcraft.yaml:23]: ah I see, yes you're right
```
❯ ls -lah root/usr/lib/jvm/java-21-openjdk-amd64/lib/
total 135M
drwxr-xr-x 3 tiago tiago 4,0K nov 14 08:51 .
drwxr-xr-x 5 tiago tiago 4,0K nov 14 08:51 ..
-rwxr-xr-x 1 tiago tiago  15K nov 14 08:51 jexec
-rwxr-xr-x 1 tiago tiago  19K nov 14 08:51 jspawnhelper
lrwxrwxrwx 1 tiago tiago   34 nov 14 08:51 jvm.cfg -> /etc/java-21-openjdk/jvm-amd64.cfg
-rw-r--r-- 1 tiago tiago  17K nov 14 08:51 libextnet.so
-rw-r--r-- 1 tiago tiago 178K nov 14 08:51 libjava.so
-rw-r--r-- 1 tiago tiago  39K nov 14 08:51 libjimage.so
-rw-r--r-- 1 tiago tiago  77K nov 14 08:51 libjli.so
-rw-r--r-- 1 tiago tiago  17K nov 14 08:51 libjsig.so
-rw-r--r-- 1 tiago tiago  65K nov 14 08:51 libnet.so
-rw-r--r-- 1 tiago tiago 108K nov 14 08:51 libnio.so
-rw-r--r-- 1 tiago tiago  63K nov 14 08:51 libverify.so
-rw-r--r-- 1 tiago tiago  45K nov 14 08:51 libzip.so
-rw-r--r-- 1 tiago tiago 135M nov 14 08:51 modules
drwxr-xr-x 2 tiago tiago 4,0K nov 14 08:51 server
-rw-r--r-- 1 tiago tiago 102K nov 14 08:51 tzdb.dat
```

tigarmo:
> Moving this one to draft as it depends on JAVA_HOME PR in craft parts

since this is going back to Draft, can you update the `requirements*txt` files to point to the craft-parts branch? Then when we merge and release that PR we can update this one with the new craft-parts version

PR #744, opened 2024-11-01
feat: make requirements.txt for extensions optional

toabctl:
There are multiple ways within python to declare dependencies (eg. via pyproject.toml, setup.py or setup.cfg). Requiring a requirements.txt file in a specific path doesn't work for a lot of projects out there so make the requirements.txt file optional.


toabctl:
Is that something you would consider to be merged? I'm happy to fix CI & linters if this is wanted.

toabctl:
friendly ping...

tigarmo:
The 12-factor extensions are fairly opinionated by design. We discussed this internally and the decision at this moment is to keep the extensions as they are, with the current set of restrictions.

rpocase:
@tigarmo what is the reasoning for only supporting a more legacy style of dependency file? [PEP631](https://peps.python.org/pep-0631/) defines a standard for defining these in pyproject.toml and is increasingly the defacto way to handle this.

tigarmo:
`requirements.txt` is an older way of specifying dependencies but it's still (as far as I'm aware) the preferred way to declare (freeze) *all* dependencies, direct and transitive, at a specific version. This is important for applications (as opposed to libraries), which is what the 12f extensions target.

The reason that we, at this moment, require that the requirements.txt file exists and declares Django/Flask/etc is to keep the extension's story compact and cohesive. The more different use cases that it tries to support, the harder to reason about it, greater chance of unexpected behavior, etc.

That said, nothing is set in stone. If there was enough of a "critical mass" of users for which the extension is useless with the mandatory requirements.txt file then we could certainly revisit the topic.

PR #745, opened 2024-11-04
Adds run_once_command rock build option

claudiubelu:
By default, rockcraft sets Pebble as the default entrypoint. However, if the defined service finishes faster than 1 second (either successfully or not), Pebble will ignore the ``on-success`` / ``on-failure`` settings and still continue running. [1]

This can be problematic for a few types of rocks meant to be used in Kubernetes. There are some workload types which are expected to eventually finish and stop, entering a Completed state, and Pebble can be preventing that, leading to some workloads / operators not able to work as intended. A service manager like Pebble is not needed in these cases. Some of these cases are:

- ``Jobs`` / ``CronJobs``: These are meant as tasks with a finite execution time. They will spawn Pods which are not meant to be endlessly running.
- ``InitContainers``: Pods may have these in order to initialize various things before the main workload containers start. The workload containers are not started until the InitContainers have finished executing and entered a Completed state.
- ``.*ctl`` type of images (e.g.: kubectl, falcoctl, longhornctl): These can be used in previously mentioned cases, but they may also be used standalone by the user in ephemeral Pods (kubectl run).

At the moment, the workaround is to forcefully add a 1.1 sleep to the Pebble service. However, that may not be an option for bare-based rocks, as they wouldn't have the sleep command (unless it would be sliced in, needlessly increasing the image size and rock building complexity). Plus, it is an unpleasant user experience for peple switching to ``rockcraft`` and hitting this issue mostly through trial and error (building the rock, uploading it, deploy in a Kubernetes env if you have one, see if you encounter issues with it or not, notice issue, investigate, find issue, fix, start over).

Alternatively, some Helm charts may have the option to completely override the image command / entrypoint, bypassing this issue. But a user would have to *know* about this issue (mostly through trial and error) and amend the Helm chart values accordingly, which is not great, considering that rocks should be drop-in replacements for regular images. But not all Helm charts have this option, in which case, the Helm chart would have to be patched, which is not great either.

[1] https://github.com/canonical/pebble/issues/240#issuecomment-1599722443

-----

### DIFF
--- a/rockcraft/models/project.py
+++ b/rockcraft/models/project.py
@@ -257,6 +257,7 @@ class Project(BuildPlanner, BaseProject):  # type: ignore[misc]
     services: dict[str, Service] | None = None
     checks: dict[str, Check] | None = None
     entrypoint_service: str | None = None
+    run_once_command: str | None = None
     platforms: dict[str, Platform | None]  # type: ignore[assignment]
 
     model_config = pydantic.ConfigDict(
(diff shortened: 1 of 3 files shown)


claudiubelu:
Note: will add unit / integration tests if the proposal is acceptable for everyone.

cjdcordeiro:
@claudiubelu tyvm for your contribution.

This is clearly something we need to address.
We've been discussing it in the scope of Pebble, as we believe that is the first place where such support should be added. See 
 - https://github.com/canonical/pebble/issues/240
 - https://github.com/canonical/pebble/issues/495

Then we'll also address the ability to run one-shot commands for rocks. We'll still need to design it though, but you'd be more than welcome to take part in those discussions, in case you want to join our [regular monthly community meetings or office hours](https://ubuntu.com/community/governance/teams/rocks)

PR #747, opened 2024-11-08
Feat(extension): Enable Async Gunicorn workers for Flask and Django e…

alithethird:
# Enable Async Gunicorn workers for Flask and Django extensions

Reviewers:
@jdkandersson
@javierdelapuente
@tigarmo
@gregory-schiano

-----


tests/unit/test_cli.py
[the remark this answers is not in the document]
alithethird: I have these tests in the Paas-charm repo. I have added a Django init test and updated the Flask init test here.

docs/reference/extensions/django-framework.rst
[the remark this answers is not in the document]
alithethird: Putting that spread test in the Charmcraft repo since it involves using the charm configs.

tests/unit/test_cli.py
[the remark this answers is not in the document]
jdkandersson: Wouldn't these tests have to be here since they relate to extension logic?

docs/reference/extensions/django-framework.rst
```diff
@@ -43,21 +43,26 @@ application. In the following example we use it to specify ``libpq-dev``:
         # list required packages or slices for your Django application below.
         - libpq-dev
 
+.. warning::
+  You can only use 1 of the dependencies parts at a time.
```
jdkandersson, line 47: Perhaps I missed it, shouldn't we specific explicitly which one we are referring to here?

docs/reference/extensions/django-framework.rst
```diff
@@ -43,21 +43,26 @@ application. In the following example we use it to specify ``libpq-dev``:
...
+
 ``parts`` > ``django-framework/async-dependencies``
 =================================================================
 
 You can use this key to specify that you want to use async gunicorn workers in
-you Django application.
-It also works just like ``django-framework/dependencies``.
-In the following example we use it to specify ``libpq-dev``:
+your Django application. It also works just like
+``django-framework/dependencies``.
```
jdkandersson, line 54: Perhaps include a link to the heading here

docs/reference/extensions/django-framework.rst
```diff
@@ -43,21 +43,26 @@ application. In the following example we use it to specify ``libpq-dev``:
...
   parts:
     django-framework/async-dependencies:
-      stage-packages:
-        # list required packages or slices for your django app below.
-        - libpq-dev
+      python-packages:
+        - gunicorn[gevent]
+
+If your project needs additional debs to run, you can add them to
+``stage-packages`` just like it is done in ``django-framework/dependencies``.
```
jdkandersson, line 65: Perhaps also include a link to the heading here

docs/reference/extensions/flask-framework.rst
```diff
@@ -41,20 +41,26 @@ application. In the following example we use it to specify ``libpq-dev``:
...
   parts:
     flask-framework/async-dependencies:
-      stage-packages:
-        # list required packages or slices for your flask app below.
-        - libpq-dev
+      python-packages:
+        - gunicorn[gevent]
+
+If your project needs additional debs to run, you can add them to
+``stage-packages`` just like it is done in ``flask-framework/dependencies``.
```
jdkandersson, line 63: Same comments as for Django

rockcraft/commands/init.py
```diff
@@ -132,8 +132,8 @@ class InitCommand(AppCommand):
 
                 # uncomment this section to enable the async workers for Gunicorn.
                 #   flask-framework/async-dependencies:
-                #     plugin: python
-                #     source: .
+                #       python-packages:
+                #       - gunicorn[gevent]
```
jdkandersson, line 136: Is the spacing correct here?

rockcraft/commands/init.py
```diff
@@ -215,8 +215,9 @@ class InitCommand(AppCommand):
                 #       - libpq-dev
                 # uncomment this section to enable the async workers for Gunicorn.
                 #   django-framework/async-dependencies:
-                #     plugin: python
-                #     source: .
+                #       python-packages:
+                #       - gunicorn[gevent]
```
jdkandersson, line 219: Is the spacing correct here?

rockcraft/commands/init.py
alithethird: It worked but it is not a good exercise I guess .

docs/reference/extensions/django-framework.rst
```diff
@@ -45,13 +45,14 @@ application. In the following example we use it to specify ``libpq-dev``:
 
 .. warning::
   You can only use 1 of the dependencies parts at a time.
+  (eg. `parts-django-framework-async-dependencies`_ or
+  `parts-django-framework-dependencies-stage-packages`_)
```
jdkandersson, line 49: I'm not sure putting it like this makes sense, it could be implying that the stage packages or limited to 1. Perhaps review this documentation with @erinecon?

rockcraft/commands/init.py
```diff
@@ -133,7 +133,7 @@ class InitCommand(AppCommand):
                 # uncomment this section to enable the async workers for Gunicorn.
                 #   flask-framework/async-dependencies:
                 #       python-packages:
-                #       - gunicorn[gevent]
+                #         - gunicorn[gevent]
```
jdkandersson, line 136: Hmm, it seems that from 133 to 134 there are 2 extra leading spaces and from 134 to 135 there are 4 extra leading spaces, should we be consistent?

tests/unit/test_cli.py
```diff
@@ -185,7 +186,7 @@ def test_run_init_flask(mocker, emitter, monkeypatch, new_dir, tmp_path):
             # uncomment this section to enable the async workers for Gunicorn.
             #   flask-framework/async-dependencies:
             #       python-packages:
-            #       - gunicorn[gevent]
+            #         - gunicorn[gevent]
```
jdkandersson, line 189: Same comment here re spacing

jdkandersson (commented):
We still need to discuss re a spread test for async

tests/conftest.py
```diff
@@ -33,6 +33,188 @@ def new_dir(tmpdir):
     os.chdir(cwd)
 
 
+@pytest.fixture()
```
javierdelapuente, line 36: do we need this fixture? it is only used for the `test_cli.py::test_run_init_django`, but the init option does not look like it checks it. Maybe a spread test could be more appropriate that checks everything?

rockcraft/extensions/gunicorn.py
```diff
@@ -273,12 +305,25 @@ def _requirements_txt_error_messages(self) -> list[str]:
 
         return []
 
+    def _dependencies_error_messages(self) -> list[str]:
+        """Ensure only 1 of the dependencies parts is defined."""
+        yaml_parts = self.yaml_data.get("parts", {})
+
+        if yaml_parts.get(
```
javierdelapuente, line 312: Maybe we could check that there is at least one of them?

docs/reference/extensions/django-framework.rst
```diff
@@ -17,6 +17,15 @@ server metrics.
     The Django extension is compatible with the ``bare``, ``ubuntu@22.04``
     and ``ubuntu@24.04`` bases.
 
+Django framework supports both synchronous and asynchronous. If you want
+asynchronous you have to add ``django-framework/async-dependencies``
+read more :ref:`django-framework/dependencies <django-sync-deps>`.
```
erinecon, line 22: I would rephrase this paragraph to say something like
```
The Django extension supports both synchronous and asynchronous
Gunicorn workers. If you want asynchronous workers, you have to add
``django-framework/async-dependencies`` to the ``rockcraft.yaml``.
Read more :ref:`django-framework/dependencies <django-sync-deps>`.
```

docs/reference/extensions/django-framework.rst
```diff
@@ -17,6 +17,15 @@ server metrics.
...
+
+If you define
+``django-framework/async-dependencies`` you can not use
+``django-framework/dependencies``. ``rockcraft pack`` will error if you try to
+use both at the same time.
```
erinecon, line 27: I would rephrase this paragraph to say something like
```
If you define
``django-framework/async-dependencies`` you can not
also include ``django-framework/dependencies``.
Running ``rockcraft pack`` will result in an error if you try to
use both at the same time.
```
Is it possible to include the error here?
(Same for Flask, although I won't add a comment there)

docs/reference/extensions/django-framework.rst
```diff
@@ -43,6 +54,43 @@ application. In the following example we use it to specify ``libpq-dev``:
...
         - libpq-dev
 
+.. _django-async-deps:
+
+``parts`` > ``django-framework/async-dependencies``
+===================================================
+
+In order to be able to use async Gunicorn workers you need to use
+``django-framework/async-dependencies`` part instead of
+``django-framework/dependencies`` part.
```
erinecon, line 64: I would rephrase this paragraph to say something like the following (maybe the last line is too long for the lint checks):
```
In order to use asynchronous Gunicorn workers, you need
to include ``django-framework/async-dependencies`` in the
``rockcraft.yaml`` while removing ``django-framework/dependencies``. 
```

docs/reference/extensions/django-framework.rst
```diff
@@ -43,6 +54,43 @@ application. In the following example we use it to specify ``libpq-dev``:
...
+
+To use this just uncomment the following lines:
```
erinecon, line 66: This line implies to me that the ``rockcraft.yaml`` file automatically contains this snippet as part of the initialization. Is that true?

In any event, just in case the snippet is removed in the future, I would rephrase this line to say something like 
```
In the ``rockcraft.yaml``, add the following lines:
```

docs/reference/extensions/django-framework.rst
```diff
@@ -43,6 +54,43 @@ application. In the following example we use it to specify ``libpq-dev``:
...
+
+.. code-block:: yaml
+
+  parts:
+    django-framework/async-dependencies:
+      python-packages:
+        - gunicorn[gevent]
+
+If your project needs additional debs to run, you can add them to
+``stage-packages`` just like it is done in :ref:`django-framework/dependencies <django-sync-deps>`:
```
erinecon, line 76: I would remove the end of the sentence for conciseness, i.e., I would remove this part:
```
just like it is done in :ref:`django-framework/dependencies <django-sync-deps>`
```

docs/.wordlist.txt
```diff
@@ -30,6 +30,7 @@ Dockerfiles
 DockerHub
 dotnet
 ECR
+eg
```
jdkandersson, line 33: Do you mean e.g.?

jdkandersson (commented):
I think we're just waiting for the spread test now

erinecon (approved):
Thanks for adding these sections and for incorporating the edits!

tests/spread/rockcraft/extension-flask-async/task.yaml
```diff
@@ -0,0 +1,109 @@
...
+  run_rockcraft pack
+
+  test -f "${ROCK_FILE}"
+  test ! -d work
+
+  # Ensure docker does not have this container image
+  docker rmi --force "${IMAGE}"
+  # Install container
+  sudo rockcraft.skopeo --insecure-policy copy "oci-archive:${ROCK_FILE}" "docker-daemon:${IMAGE}"
+  # Ensure container exists
```
jdkandersson, line 51: Ensure image exists?

jdkandersson (approved):

tigarmo (changes requested):
thanks! I left some comments on some improvements I think we should make.

rockcraft/extensions/gunicorn.py
```diff
@@ -273,12 +286,25 @@ def _requirements_txt_error_messages(self) -> list[str]:
...
+            f"{self.framework}-framework/async-dependencies", None
+        ) and yaml_parts.get(f"{self.framework}-framework/dependencies", None):
+            return [
+                f"Cannot have both sync and async dependencies. https://bit.ly/{self.framework}-async-doc"
```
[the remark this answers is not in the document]
alithethird: This makes the output way too long. We also print this out in the Charm Message and when we use the long URL it doesn't print the whole URL. I have talked with our TA and she talked with other TA's and decided to use `bit.ly`

tests/unit/test_cli.py
```diff
@@ -229,3 +234,259 @@ def test_run_init_flask(mocker, emitter, monkeypatch, new_dir, tmp_path):
...
+        """
+    )
+    emitter.assert_message(
+        textwrap.dedent(
+            f"""\
+        Go to {versioned_url}/reference/extensions/django-framework to read more about the 'django-framework' profile."""
+        )
+    )
+    monkeypatch.setenv("ROCKCRAFT_ENABLE_EXPERIMENTAL_EXTENSIONS", "0")
+    project.Project.unmarshal(extensions.apply_extensions(tmp_path, rock_project_yaml))
```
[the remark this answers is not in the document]
alithethird: It applies the extension and makes sure the `rockcraft.yaml` file is proper.

tests/data/flask/rockcraft_async.yaml
```diff
@@ -0,0 +1,76 @@
...
+#     build-packages:
+#       # for example, if you need pkg-config and libxmlsec1-dev to build one
+#       # of your packages:
+#       - pkg-config
+#       - libxmlsec1-dev
+
+# uncomment this section to enable the async workers for Gunicorn.
+  flask-framework/async-dependencies:
+    # python-packages:
+    #   - gunicorn[gevent]
```
jdkandersson, line 55: Look like this should fail

erinecon (changes requested):
Thanks for updating the docs!

docs/reference/extensions/django-framework.rst
```diff
@@ -57,44 +49,32 @@ application. In the following example we use it to specify ``libpq-dev``:
...
-.. warning::
-  You can use either ``django-framework/async-dependencies`` or
-  ``django-framework/dependencies``, but not both at the same time.
-  To read more about synchronous dependencies,
-  see :ref:`django-framework/dependencies <django-sync-deps>`.
+.. note::
+    The Django extension is compatible with the ``bare``, ``ubuntu@22.04`` and
+    ``ubuntu@24.04`` bases.
 
+Example rock
```
jdkandersson, line 77: This seems out of place?

rockcraft/extensions/gunicorn.py
```diff
@@ -125,6 +112,30 @@ def _gen_parts(self) -> dict:
...
+          - gevent [some_extra]
+        """
+        requirements_file = self.project_root / "requirements.txt"
+        with requirements_file.open() as f:
+            lines = f.readlines()
+            for line in lines:
+                for token in line.split():
+                    if re.match("^gevent[|[=|@|\\s|\n<|>]|(^gevent$)", token):
+                        return "gevent"
+        return "sync"
```
jdkandersson, line 137: I would propose that the algorithm here is:
1. Read the `requirements.txt` file using pathlib.Path(...).read_text()
2. Convert each line to a package name using the regex here: https://packaging.python.org/en/latest/specifications/name-normalization/#name-format
3. Normalize the names using https://packaging.python.org/en/latest/specifications/name-normalization/#name-format
4. Check if gevent is in the collection

alithethird: That is a great way! Thank you

alithethird: The regex here doesn't consider when versioning is used with package names. The regex I crafted makes sure user can use versioning or local/vcs with gevent package. https://packaging.python.org/en/latest/specifications/name-normalization/#name-format

I tested the regex with rockcrafts  own `requirements.txt` file but it was not able to find even 1 package since all of them are version pinned. [test code](https://pastebin.canonical.com/p/s4BpVkcXBT/)

jdkandersson: You can remove the `$` at the end of the regex and it should match the package name up to the version part

rockcraft/extensions/gunicorn.py
```diff
@@ -113,27 +114,18 @@ def _gen_parts(self) -> dict:
...
-                    if re.match("^gevent[|[=|@|\\s|\n<|>]|(^gevent$)", token):
-                        return "gevent"
+        requirements_text = requirements_file.read_text()
+        for line in requirements_text.splitlines():
+            if not line.strip():
+                continue
+            clean_line = line.strip()
+            if clean_line.startswith("#"):
+                continue
+            req = Requirement(clean_line)
```
jdkandersson, line 126: What happens if we skip the two checks? Do we get an exception?

alithethird: Nope, it means `gevent` package is not in the `requirements.txt` file so the rock is sync.

jdkandersson: I mean what happens if you delete lines 121-125, what happens if we get an empty line or a line starting with #?

jdkandersson: I'm thinkking we can just try catch pass rather than have these lines

jdkandersson: Also, I would move line 123 above 121 and just check if the clean line is empty

alithethird: If I remove we get an error. Sure, we can do try/catch here.

jdkandersson (approved):

spread.yaml
```diff
@@ -139,6 +139,7 @@ suites:
     summary: tests for rockcraft core functionality, independent of host system
     systems:
     - ubuntu-22.04-64
+    - ubuntu-24.04-64
```
tigarmo, line 142: how come?

tests/unit/test_cli.py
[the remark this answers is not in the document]
tigarmo: please add a comment describing that

tests/unit/extensions/test_gunicorn.py
```diff
@@ -107,7 +107,7 @@ def test_flask_extension_default(tmp_path, flask_input_yaml, packages):
             "flask": {
                 "after": ["statsd-exporter"],
                 "command": "/bin/python3 -m gunicorn -c "
-                "/flask/gunicorn.conf.py app:app",
+                "/flask/gunicorn.conf.py app:app -k [ sync ]",
```
tigarmo, line 110: what about the `gevent` case? I think we should add one more parameter here

tigarmo (changes requested):
thankss

tigarmo:
Readthedocs errors: 

```
<generated text>:3: ERROR: Unexpected indentation.
<generated text>:3: ERROR: Unexpected indentation.
/home/docs/checkouts/readthedocs.org/user_builds/canonical-rockcraft/checkouts/747/docs/reference/extensions/flask-framework.rst:52: WARNING: duplicate label async-gunicorn-workers, other instance in /home/docs/checkouts/readthedocs.org/user_builds/canonical-rockcraft/checkouts/747/docs/reference/extensions/django-framework.rst
```

spread.yaml
alithethird: Removed

jdkandersson (approved):

tests/unit/extensions/test_gunicorn.py
```diff
@@ -127,6 +127,83 @@ def test_flask_extension_default(tmp_path, flask_input_yaml, packages):
     }
 
 
+@pytest.mark.usefixtures("flask_extension")
+def test_flask_extension_async(tmp_path, flask_input_yaml):
```
tigarmo, line 131: this test is almost identical to `test_flask_extension_default`. For the sake of our collective sanity, can't we just do a single test with the proper `parametrize` parameters?

tests/unit/extensions/test_gunicorn.py
```diff
@@ -518,6 +595,78 @@ def test_django_extension_default(tmp_path, django_input_yaml):
     }
 
 
+@pytest.mark.usefixtures("django_extension")
+def test_django_extension_async(tmp_path, django_input_yaml):
```
tigarmo, line 599: same question here

tests/unit/extensions/test_gunicorn.py
tigarmo: some like (I didn't test this)
```python
@pytest.mark.usefixtures("flask_extension")
@pytest.mark.parametrize("packages", ["other\nflask", "flask", "Flask", " flask == 99"])
@pytest.mark.parametrize(("async_package", "expected_worker"), [("gevent", "gevent"),("", "sync")])
def test_flask_extension_default(tmp_path, flask_input_yaml, packages, async_package, expected_worker):
    full_packages = "\n".join([packages, async_package])

**Review by tigarmo — CHANGES_REQUESTED** [review]

File: docs/reference/extensions/django-framework.rst [hunk 29]
```diff
@@ -60,14 +61,10 @@ updates the Pebble plan to use the asynchronous workers. If you have ``gevent``
...
-  :caption: Use sync workers instead of gevent
-   $ docker run \
-       --name django-container \
-       -d -p 8138:8000 \
-       django-image:1.0 \
-       --args django sync
+.. code-block:: bash
 
+   docker run --name django-container -d -p 8138:8000 django-image:1.0 \
+   --args django sync
```
jdkandersson, line 67: What is the reason for picking port 8138 please?

alithethird: It is just a random port. Port 8000 in the host might be unavailable.

jdkandersson: Let's do 8000, if it isn't available the user can change it

jdkandersson (approved):

tests/unit/extensions/test_gunicorn.py
```diff
@@ -51,85 +51,14 @@ def django_input_yaml_fixture():
...
-    }
-
-
-@pytest.mark.usefixtures("flask_extension")
-def test_flask_extension_async(tmp_path, flask_input_yaml):
-    (tmp_path / "requirements.txt").write_text("flask\ngevent")
+@pytest.mark.parametrize(
+    ("async_package", "expected_worker"), [("gevent", "gevent"), ("", "sync")]
+)
+def test_flask_extension_default(
```
tigarmo, line 57: much better, thanks 👍🏾

tigarmo (approved):
thanks! @medubelko can you take a look at the updated docs and give the second review?

rockcraft/extensions/gunicorn.py
```diff
@@ -112,6 +113,19 @@ def _gen_parts(self) -> dict:
             }
         return parts
 
+    def _check_async(self) -> str:
```
toabctl, line 116: shouldn't that return a bool instead of string?

docs/reference/extensions/flask-framework.rst
```diff
@@ -16,6 +16,12 @@ server metrics.
     The Flask extension is compatible with the ``bare``, ``ubuntu@22.04``
     and ``ubuntu@24.04`` bases.
 
+The Flask extension supports both synchronous and asynchronous
+Gunicorn workers. If you want asynchronous workers, you have to add
+``gevent`` package to the ``requirements.txt`` file.
+Read more
+:ref:`Using asynchronous Gunicorn workers <flask-async-gunicorn-workers>`.
```
medubelko, line 23: Since it discusses dependencies, I think the second sentence should be added as an option in the _Project requirements_, section, like so:

```rst
2. The project must...

For the project to make use of asynchronous Gunicorn workers:

- The ``requirements.txt`` file must include ``gevent`` as a dependency.
```

medubelko, line 23:
```suggestion
Gunicorn workers.
```

docs/reference/extensions/flask-framework.rst
```diff
@@ -41,6 +47,23 @@ application. In the following example we use it to specify ``libpq-dev``:
...
+
+Using asynchronous Gunicorn workers
+===================================
+
+If you want to use asynchronous workers, you have to add ``gevent`` package to
+the ``requirements.txt`` file. Rockcraft automatically detects this and updates
+the pebble plan to use the asynchronous workers. If you have ``gevent``
+installed in your rock but decided to use ``sync`` workers instead you can use
+the ``--args`` parameter of ``docker run`` to use ``sync`` workers instead of
+the default ``gevent``:
```
medubelko, line 60: I think we can restyle this section:

- Since this is a reference, we generally want to frame this more as mechanical explantions than how-to language.
- I found a few ways to tighten up the language and clarify meaning.

```suggestion
Gunicorn worker selection
=========================

If the project has gevent as a dependency, Rockcraft automatically updates the
pebble plan to spawn asynchronous Gunicorn workers.

When the project instead needs synchronous workers, you can override the worker
type by adding ``--args flask sync`` to the Docker command that launches the
rock:
```

docs/reference/extensions/django-framework.rst
medubelko, line 1: We can apply my other suggestions to this file, too.

docs/reference/extensions/django-framework.rst
```diff
@@ -43,6 +50,26 @@ application. In the following example we use it to specify ``libpq-dev``:
...
+so by using the ``--args`` parameter of ``docker run``:
+
+.. code-block:: bash
+
+   docker run --name django-container -d -p 8000:8000 django-image:1.0 \
+   --args django sync
+
+.. note::
+    The Django extension is compatible with the ``bare``, ``ubuntu@22.04`` and
+    ``ubuntu@24.04`` bases.
```
medubelko, line 71: This looks like it was misplaced. Can we move this to the introduction, like in the other page?

medubelko (changes requested):
@jdkandersson Apologies for getting to this so late. Let me know if I should clarify any of my suggestions. :)

docs/reference/extensions/flask-framework.rst
alithethird: Hey @medubelko thanks for your comments? Should I move the `Gunicorn workers. If you want asynchronous workers, you have to add
``gevent`` package to the ``requirements.txt`` file.` part to _Project requirements_ section or copy it there?

docs/reference/extensions/django-framework.rst
```diff
@@ -29,6 +32,13 @@ There are 2 requirements to be able to use the ``django-framework`` extension:
    ``./<Rock name with - replaced by _>/<Rock name with - replaced by _>/manage.py``
    relative to the ``rockcraft.yaml`` file.
 
+For the project to make use of asynchronous Gunicorn workers:
+
+- The ``requirements.txt`` file must include ``gevent`` as a dependency.
+
+Read more
+:ref:`Gunicorn worker selection <django-gunicorn-worker-selection>`.
```
medubelko, line 40: On second look, let's cut these lines, too. _Requirements_ isn't the right place to link to another section in the doc. I think the reader can find their way to it on their own.
```suggestion
```

docs/reference/extensions/flask-framework.rst
```diff
@@ -25,7 +28,14 @@ There are 2 requirements to be able to use the ``flask-framework`` extension:
...
    there must be an ``app.py`` file at the root of the project with the name
-   of the Flask object is set to ``app``
+   of the Flask object is set to ``app``.
+
+For the project to make use of asynchronous Gunicorn workers:
+
+- The ``requirements.txt`` file must include ``gevent`` as a dependency.
+
+Read more
+:ref:`Gunicorn worker selection <flask-gunicorn-worker-selection>`.
```
medubelko, line 38:
```suggestion
```

medubelko (changes requested):
@alithethird Apologies for pinging the wrong person before. The fingers were moving too fast. :sweat: 

I have one more small change and then I think we're good to merge.

alithethird:
@medubelko all good :+1: . Thanks a lot for your reviews :)

medubelko (approved):

PR #751, opened 2024-11-15
chore: migrate setup.py to pyproject.toml

lengau:

-----

This removes setup.py and setup.cfg in favour of the newer pyproject.toml file.

It is a necessary portion of our move to starflow workflows, but standalone and per-project.


pyproject.toml
```diff
@@ -1,10 +1,99 @@
...
+[project.scripts]
+rockcraft = "rockcraft.cli:run"
+
+[project.optional-dependencies]
+dev = [
+    "black",
+    "canonical-sphinx[full]",
+    "codespell",
+    "coverage",
+    "mccabe<0.7.0",
```
bepri, line 41: Should this `mccabe` import preserve the comment from the old `setup.cfg`? It was "`# to resolve version conflict`". It seems like it was added 2 years ago, so it may not be relevant anymore, but I'd like to make sure.

bepri (approved):

pyproject.toml
lengau: I'll do ya one better: this should have been removed when we removed flake8, so I've just tossed it now :-)

pyproject.toml
```diff
@@ -1,10 +1,98 @@
...
 
+[tool.setuptools.dynamic]
+readme = {file = "README.rst"}
+
+[tool.setuptools.packages.find]
+include = ["rockcraft"]
+namespaces = false
+
+[tool.setuptools.package-data]
+"share.rockcraft" = ["extensions"]
```
tigarmo, line 94: @lengau the spread tests are failing because they expect to find the files from `<root>/extensions` at `'/snap/rockcraft/x1/share/rockcraft/extensions/flask-framework'`. I guess this worked before because we were 'abusing' the setup.py setup to add files outside the package dir

Am I correct in thinking we should move away from that? We'll need to preserve that path on the snap for backwards-comp but maybe those data files should go in `rockcraft/extensions/data` (or similar)

tigarmo: I updated this by adding `extensions` as a separate package and then moving it from site-packages to share/rockcraft/ when packing the snap

tigarmo (approved):
many thanks, looks great!

PR #754, opened 2024-11-18
lint: suppress linter warning

vpa1977:
pyright complains about assigning str | None type in constructor to a str type.
Exception can not be None, strerror is declared as str. We can do a str() when assigning msg, but this seems a pyright issue, so probably it is best to suppress the type check.

Apparently the declaration does not match the actual returned type: https://github.com/python/typeshed/issues/9864, so using `or "unknown"` as suggested.

-----

### DIFF
--- a/rockcraft/models/project.py
+++ b/rockcraft/models/project.py
@@ -480,10 +480,10 @@ def load_project(filename: Path) -> dict[str, Any]:
         with open(filename, encoding="utf-8") as yaml_file:
             yaml_data = yaml.safe_load(yaml_file)
     except OSError as err:
-        msg = err.strerror
+        msg = err.strerror or "unknown"
         if err.filename:
             msg = f"{msg}: {err.filename!r}."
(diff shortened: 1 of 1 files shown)


tigarmo (approved):
thanks!

dimaqq (commented):
the docs seem to suggest that strerror is always set: https://docs.python.org/3/library/exceptions.html#OSError

yet typeshed is pretty explicit that it’s optional: https://github.com/python/typeshed/blob/6647ae9bc781041b0021de35f278b42e622875ca/stdlib/builtins.pyi#L1916

I suppose a discussion on typeshed is worthwhile, and perhaps until that, what do you think about changing the first assignment instead, to `msg = e.strerror or “unknown”` ?

dimaqq:
See https://github.com/python/typeshed/issues/9864 for a case where strerror may be None.

tigarmo (approved):

PR #756, opened 2024-11-19
feat(snap): add git as craft.git to rockcraft

dariuszd21:

This PR adds stages `git` in the `rockcraft` snap package, to enable `git`-based operations for `craft` libraries.

(CRAFT-3548)

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -71,6 +71,9 @@ parts:
       - "usr/lib/x86_64-linux-gnu/libpsl*"
       - "usr/lib/x86_64-linux-gnu/librtmp*"
       - "usr/lib/x86_64-linux-gnu/libnghttp2*"
+      - "usr/lib/x86_64-linux-gnu/libldap*"
+      - "usr/lib/x86_64-linux-gnu/liblber*"
+      - "usr/lib/x86_64-linux-gnu/libsasl2*"
 
   libgit2: # part came from snapcraft@5fbaab144842f
     source: https://github.com/libgit2/libgit2/archive/refs/tags/v1.7.1.tar.gz
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -65,14 +65,8 @@ parts:
     build-attributes:
       - enable-patchelf
     prime:
-      - "-usr/bin"
-      - "-usr/share/doc"
-      - "-usr/share/man"
-      # perl is part of the core22 / core24
-      - "-usr/share/perl"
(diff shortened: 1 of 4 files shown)


snap/snapcraft.yaml
```diff
@@ -53,6 +53,26 @@ parts:
...
+    plugin: nil
+    stage-packages: [git]
+    build-attributes:
+      - enable-patchelf
+    prime:
+      - "usr/lib/git-core"
+      - "usr/share/git-core"
+      - "usr/lib/x86_64-linux-gnu"
+      # perl is part of the core22 / core24
+      - "-usr/lib/x86_64-linux-gnu/perl"
```
tigarmo, line 72: I suspect these removals aren't needed; if you specify any inclusions when only those paths are primed

tests/spread/rockcraft/craftgit/task.yaml
```diff
@@ -0,0 +1,7 @@
+summary: craft.git tool test
+
+execute: |
+  run_rockcraft pack
+
+  test -f craftgit-test_latest_amd64.rock
+  test ! -d parts -a ! -d stage -a ! -d prime
```
tigarmo, line 7: this line seems unrelated to the test itself

tests/spread/rockcraft/craftgit/rockcraft.yaml
```diff
@@ -0,0 +1,25 @@
+name: craftgit-test
```
tigarmo, line 1: 👍🏾

tigarmo (approved):

snap/snapcraft.yaml
dariuszd21: I'll try to include only libcurl and see if that will work, then I could remove the exclusions entirely:
```
- "usr/lib/x86_64-linux-gnu/libcurl-gnutls*"
```

dariuszd21: In the end I needed to include the following to get rid of exclusion:
```yaml
      - "usr/lib/x86_64-linux-gnu/libcurl-gnutls*"
      - "usr/lib/x86_64-linux-gnu/libpsl*"
      - "usr/lib/x86_64-linux-gnu/librtmp*"
      - "usr/lib/x86_64-linux-gnu/libnghttp2*"
```

tests/spread/rockcraft/craftgit/task.yaml
dariuszd21: Removed that, and added a restore that deletes a created rock instead

mr-cal (approved):

Issue #659, opened 2024-08-01
Kubernetes `runAsNonRoot`  incompatibility with `run-user: _daemon_`


berkayoz:
### Bug Description

Some deployments are setting `runAsNonRoot: true` for the pod security context which prevents a container from running as root. To possibly address this we've set `run-user: _daemon_` however Kubernetes checks uid/gid numerically and complains about the `_daemon_` user.

You can find the Kubernetes implementation [here](https://github.com/kubernetes/kubernetes/blob/f8d5b2074c9b3a1629ef085b9e17a5370f393732/pkg/kubelet/kuberuntime/security_context_others.go#L31-L53)

As a work around we have to set `securityContext.runAsUser=584792` manually on manifests which hurts the drop-in image replacement story. Could we possibly set numeric UID as the OCI user by default? 

Your help is much appreciated, many thanks!

### To Reproduce

Deploy a Pod that uses a rock with the `securityContext` set as so
```
spec:
  securityContext:
    runAsNonRoot: true
```

You can also [check out](https://github.com/canonical/cert-manager-rock/blob/bef62cd152a8f636f9e291a609c1a976c228e982/tests/integration/test_certmanager.py) our `cert-manager` tests to verify it in a real world example.

### Environment

Ubuntu 24.04

### rockcraft.yaml

```shell
name: cert-manager-controller
summary: ROCK for the cert-manager-controller Project.
description: |
  This ROCK is a drop-in replacement for the autoscaling/cert-manager-controller image.
version: "1.12.2"
license: Apache-2.0

base: bare
build-base: ubuntu@22.04
platforms:
  amd64:
  arm64:

run-user: _daemon_
entrypoint-service: cert-manager-controller
services:
  cert-manager-controller:
    override: replace
    summary: "cert-manager-controller service"
    startup: enabled
    command: "/controller-linux [ -h ]"
    on-failure: shutdown

parts:
  cert-manager-controller:
    plugin: nil
    source: https://github.com/cert-manager/cert-manager.git
    source-type: git
    source-tag: v1.12.2
    source-depth: 1
    build-snaps:
      - jq
      - go/1.22/stable
    override-build: |
      # CTR=echo is hacky way of passing docker check not required for build
      make CTR=echo _bin/server/controller-linux-${CRAFT_PLATFORM}
      cp _bin/server/controller-linux-${CRAFT_PLATFORM} ${CRAFT_PART_INSTALL}/controller-linux
    prime:
      - controller-linux
```

[...]

PR #757, opened 2024-11-20
Change .Config.User from username to userid

linostar:

-----

This addresses Issue #659 by setting the OCI image's .Config.User to userid instead of the username.

### DIFF
--- a/rockcraft/oci.py
+++ b/rockcraft/oci.py
@@ -345,16 +345,15 @@ class Image:
         src_path = self.path / f"{name}:{tag}"
         _copy_image(f"oci:{str(src_path)}", f"oci-archive:{filename}:{tag}")
 
-    def set_default_userid(self, userid: int) -> None:
-        """Set the default runtime userid for the OCI image.
+    def set_default_user(self, userid: int, username: str) -> None:
+        """Set the default runtime user for the OCI image.
 
         :param userid: userid of the default user (must already exist)
+        :param username: username of the default user (must already exist)
         """
         image_path = self.path / self.image_name
         params = ["--clear=config.entrypoint", "--config.user", str(userid)]
         _config_image(image_path, params)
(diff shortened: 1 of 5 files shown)


rockcraft/oci.py
```diff
@@ -344,15 +344,15 @@ def to_oci_archive(self, tag: str, filename: str) -> None:
...
 
-        :param user: name of the default user (must already exist)
+        :param userid: userid of the default user (must already exist)
         """
         image_path = self.path / self.image_name
-        params = ["--clear=config.entrypoint", "--config.user", user]
+        params = ["--clear=config.entrypoint", "--config.user", str(userid)]
         _config_image(image_path, params)
-        emit.progress(f"Default user set to {user}")
+        emit.progress(f"Default user set to {userid}")
```
cjdcordeiro, line 355: it would be nice to print the username too, cause people won't know what those large UIDs map to. Something like
`Default user set to {userid} ({username})`

linostar: Done

rockcraft/oci.py
```diff
@@ -344,15 +345,17 @@ def to_oci_archive(self, tag: str, filename: str) -> None:
...
 
-        :param user: name of the default user (must already exist)
+        :param userid: userid of the default user (must already exist)
         """
         image_path = self.path / self.image_name
-        params = ["--clear=config.entrypoint", "--config.user", user]
+        params = ["--clear=config.entrypoint", "--config.user", str(userid)]
         _config_image(image_path, params)
-        emit.progress(f"Default user set to {user}")
+        usernames = [k for k, v in SUPPORTED_GLOBAL_USERNAMES.items() if v == userid]
```
cjdcordeiro, line 356: Wouldn't it be less expensive to simply pass the username as an arg?

linostar: Adjusted.

cjdcordeiro (approved):

tigarmo (approved):

tigarmo:
thanks!

PR #759, opened 2024-11-20
fix: check colon presence in all ubuntu bases

linostar:

-----

This scales up the check we have against colon-using base names (e.g. "ubuntu:22.04"), so that it covers 24.04 and any new ubuntu base we add in the future.

### DIFF
--- a/rockcraft/models/project.py
+++ b/rockcraft/models/project.py
@@ -74,8 +74,6 @@ MESSAGE_INVALID_NAME = (
     "and may not have two hyphens in a row."
 )
 
-DEPRECATED_COLON_BASES = ["ubuntu:20.04", "ubuntu:22.04"]
-
 
 ProjectName = Annotated[
     str,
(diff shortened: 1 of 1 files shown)


rockcraft/models/project.py
```diff
@@ -133,7 +131,7 @@ def _validate_deprecated_build_base(cls, base_value: str | None) -> str | None:
 
     @staticmethod
     def _check_deprecated_base(base_value: str | None, field_name: str) -> str | None:
-        if base_value in DEPRECATED_COLON_BASES:
+        if base_value and base_value.startswith("ubuntu:"):
```
tigarmo, line 134: The reason we special-cased `ubuntu:20.04` and `ubuntu:22.04` was to not break existing rockcraft projects when we moved to `ubuntu@`. For newer bases we want users to use `ubuntu@`, is there a specific reason to allow this now?

tigarmo (changes requested):

rockcraft/models/project.py
linostar: For bases from 24.04 and onwards, it is just a consistency thing, because we don't want users to use `ubuntu:24.04`, right? (users coming from a docker/lxd background are more used to using colons).

PR #764, opened 2024-11-28
feat: update 'rockcraft init' base to 24.04

### LINKED ISSUE #748: set "rockcraft init"'s base to 24.04
### What needs to get done

`base: ubuntu@22.04` -> `base: ubuntu@24.04` for the `rockcraft init`'s template.

### Why it needs to get done

24.04 is the latest LTS

tigarmo:
Fixes #748

-----

### DIFF
--- a/rockcraft/templates/simple/rockcraft.yaml.j2
+++ b/rockcraft/templates/simple/rockcraft.yaml.j2
@@ -1,7 +1,7 @@
 name: {{name}}
 # see {{versioned_url}}/explanation/bases/
 # for more information about bases and using 'bare' bases for chiselled rocks
-base: ubuntu@22.04 # the base environment for this rock
+base: ubuntu@24.04 # the base environment for this rock
 version: '0.1' # just for humans. Semantic versioning is recommended
 summary: Single-line elevator pitch for your amazing rock # 79 char long summary
 description: |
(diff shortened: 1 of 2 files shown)


cjdcordeiro (approved):
LGTM.
Can be done separately outside this PR, but we should also update the docs' examples that provide snippets with 22.04 still

rockcraft/templates/simple/rockcraft.yaml.j2
```diff
@@ -1,7 +1,7 @@
 name: {{name}}
 # see {{versioned_url}}/explanation/bases/
 # for more information about bases and using 'bare' bases for chiselled rocks
-base: ubuntu@22.04 # the base environment for this rock
+base: ubuntu@24.04 # the base environment for this rock
```
upils, line 4: The django and flask templates are still on `22.04`. Is that intentional to not update them to `24.04`?

upils (changes requested):

rockcraft/templates/simple/rockcraft.yaml.j2
tigarmo: Yes, I need to check with the "owners" of the 12f whether we want to bump the default base for the extensions.
Not all extensions support all bases, so this should happen in a separate task

upils (approved):

PR #767, opened 2024-12-03
build(snap): update snap to core24

### LINKED ISSUE #750: Update the rockcraft snap to core24
### What needs to get done

Update the rockcraft snap to core24

### Why it needs to get done

To keep up with the times

tigarmo:
Fixes #750

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,5 +1,5 @@
 name: rockcraft
-base: core22
+base: core24
 summary: A craft like experience to create rocks
 description: |
     Rockcraft aims to take the same primitives used in Charmcraft and Snapcraft
@@ -13,6 +13,14 @@ environment:
   PATH: "$SNAP/libexec/rockcraft:/snap/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
   LD_LIBRARY_PATH: "$SNAP/none"
 
+platforms:
+  amd64:
+  arm64:
+  armhf:
+  riscv64:
+  s390x:
+  ppc64el:
+
 apps:
   rockcraft:
     command: bin/python $SNAP/bin/rockcraft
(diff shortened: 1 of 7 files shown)


snap/snapcraft.yaml
```diff
@@ -65,15 +74,14 @@ parts:
...
-      - "usr/lib/x86_64-linux-gnu/liblber*"
-      - "usr/lib/x86_64-linux-gnu/libsasl2*"
+      - "-usr/bin"
+      - "-usr/share/doc"
+      - "-usr/share/man"
+      # perl is part of the core22 / core24
+      - "-usr/share/perl"
+      - "-usr/lib/x86_64-linux-gnu/perl"
+      - "-usr/lib/x86_64-linux-gnu/libperl*"
+      - "-usr/lib/x86_64-linux-gnu/libgdbm*"
```
dariuszd21, line 84: Previous version has some missing library for `core24` ?

dariuszd21 (approved):
Thank you!

Is it something that should be included in changelog ?

snap/snapcraft.yaml
tigarmo: yes, it was complaining about `libssh.so.0.4`

tigarmo:
> Thank you!
> 
> Is it something that should be included in changelog ?

We're not currently updating the changelog with individual PRs in rockcraft. I want to start doing that next year (maybe) but for now the changelog is a separate task pre-release.

mr-cal (commented):
Do you need to update the github workflows to test on 3.12?

snap/snapcraft.yaml
```diff
@@ -65,15 +74,14 @@ parts:
...
-      - "usr/lib/x86_64-linux-gnu/libpsl*"
-      - "usr/lib/x86_64-linux-gnu/librtmp*"
-      - "usr/lib/x86_64-linux-gnu/libnghttp2*"
-      - "usr/lib/x86_64-linux-gnu/libldap*"
-      - "usr/lib/x86_64-linux-gnu/liblber*"
-      - "usr/lib/x86_64-linux-gnu/libsasl2*"
+      - "-usr/bin"
+      - "-usr/share/doc"
+      - "-usr/share/man"
+      # perl is part of the core22 / core24
```
sergiusens, line 80: are we using perl itself from the core24 base? If so, that could be a problem as it is not `classic` ready

sergiusens (approved):

tests/spread/rockcraft/plugin-python/src/setup.py
```diff
@@ -1,4 +1,4 @@
-from distutils.core import setup
+from setuptools import setup
```
sergiusens, line 1: this worries me a bit because it is a test

PR #772, opened 2024-12-16
build(snap): drop 'riscv64' for now

tigarmo:
Looks like the combination of core24 + riscv64 is not supported by the store currently, which means that 'edge' hasn't been following 'main' since we updated the snap to core24.

-----


sergiusens (approved):
This is strange, but we do not need riscv64 for now anyways

bepri (approved):

PR #775, opened 2024-12-17
docs: apply release-notes template to rockcraft

tigarmo:
The first commit adds the release-notes template from starbase, verbatim. The second commit then "renders" the template specifically for Rockcraft. So really only the second commit has actual changes

A follow-up PR will then do the release notes for the upcoming 1.7.0 release


docs/release-notes/index.rst
tigarmo, line 1: https://canonical-ubuntu-documentation-library--775.com.readthedocs.build/rockcraft/en/775/release-notes/

docs/release-notes/index.rst
```diff
@@ -0,0 +1,257 @@
...
+This page lists past release notes for Rockcraft, summarising new features, bug
+fixes and backwards-incompatible changes in each version. It also contains the
+release and support policies for Rockcraft.
+
+
+Current releases
+----------------
+
+
+Rockcraft 1
```
tigarmo, line 15: @medubelko is this the desired format of the "<latest release>" header?

docs/release-notes/index.rst
```diff
@@ -0,0 +1,257 @@
...
+Rockcraft software bases are derived from Ubuntu LTS releases, and their
+development keeps pace with the OS's new releases and support lifecycle.
+
+.. toctree::
+   :maxdepth: 1
+
+
+.. release note template:
+
+  Starcraft 2.0 release notes
```
tigarmo, line 69: Just to clarify, are these notes for individual releases meant to be added as separate files under `release-notes/`?

medubelko: Correct, yes. So there should be a file for Rockcraft 1.6.0 in this dir. Best to name it `rockcraft-1-6-0`.

docs/release-notes/index.rst
medubelko: If that's the title for this mainline release, yes. I think for Snapcraft we call it "Snapcraft  7".

docs/release-notes/index.rst
```diff
@@ -0,0 +1,257 @@
...
+
+
+Current releases
+----------------
+
+
+Rockcraft 1
+~~~~~~~~~~~
+
+- :ref:`release-1.6.0`.
```
medubelko, line 18: The intention here is to link to the release note pages, not the changelog. I think for the changelog we should instead link it on the release note pages themselves – the changelog is already an indexed page, so it's fairly discoverable as-is.

medubelko (changes requested):
The remaining work is to add `rockcraft-1-6-0.rst` and fill out the template. Let me know if you need support for that.

docs/release-notes/index.rst
tigarmo: There hasn't been a title yet because there's only ever been this one "mainline" release, but "Rockcraft 1" is as good as any

tigarmo: ...although the other option, in this case, would be to omit this header (since there's only one) and re-introduce it once "Rockcraft 2" comes around?

docs/release-notes/index.rst
tigarmo: ah I think I'm missing something here then - I thought these release notes were replacing the existing changelog altogether. What is the difference between the two then?
For example, looking at the changelog for [1.6.0](https://github.com/canonical/rockcraft/blob/main/docs/reference/changelog.rst) - it's a different format but it's already structured in an end-user language. Aren't the two documents redundant?

docs/release-notes/index.rst
medubelko: Oh, I see, I missed your second point about splitting this into two PRs.

For now this is fine, but let's make sure in the next PR that this index no longer links to the changelog.

medubelko: I don't think we need to write notes for 1.6.0. Let's just focus on making a good example using 1.7.0.

docs/release-notes/index.rst
medubelko: The release notes are a new requirement of documentation sets. There's no rule against also having a changelog. When we originally discussed this I saw value in keeping the changelog for developers, but if at a later point we want to rethink the idea, I'm fine with that.

The release notes are intended to be consumer- and user-friendly copy that occupies several roles, among them _announcing_ releases more than just reporting them, highlighting developments and our intentions, and communicating to a broader audience (such as C-level, gov workers).

The two documents can exist in parallel, with the changelog covering every change at any level of minutiae (if you'd like), while with release notes we can be more selective and cover the most significant work and take the time to explain the background, motivation, use-cases, and other bits of _why_ in our progress.

tigarmo: ah I see. I have a draft of release notes for 1.7.0 that I made with the notion that it was replacing the changelog, so we can use that upcoming PR to discuss which entries "deserve" to be in the release notes and which notes can go on the changelog.

docs/release-notes/index.rst
tigarmo: updated (removed the link to the 1.6.0 changelog)

docs/release-notes/index.rst
tigarmo: @medubelko I ended up doing this^. Let me know if you want me to add the "Rockcraft 1" header back.

docs/release-notes/index.rst
medubelko: Sounds good to me! :)

docs/release-notes/index.rst
medubelko: I think that works.

lengau (approved):

PR #776, opened 2024-12-17
docs: release notes for 1.7.0

tigarmo:

-----


docs/release-notes/rockcraft-1-7-0.rst
tigarmo, line 1: https://canonical-ubuntu-documentation-library--776.com.readthedocs.build/rockcraft/en/776/release-notes/rockcraft-1-7-0/

docs/release-notes/index.rst
```diff
@@ -60,6 +60,8 @@ development keeps pace with the OS's new releases and support lifecycle.
 .. toctree::
    :maxdepth: 1
 
+   rockcraft-1-7-0
```
tigarmo, line 63: Without this entry, the build fails because the document isn't added in any table of contents. However, *with* this entry this release-notes doc ends up having two listings for a given version:

image


Should this toctree be `:hidden:`?

docs/release-notes/rockcraft-1-7-0.rst
```diff
@@ -0,0 +1,101 @@
+.. _release-1.7.0:
+
+Rockcraft 1.7.0 release notes
```
tigarmo, line 3: @medubelko as we discussed earlier, I ended up putting everything that I would put on a changelog here. Please take a look and point out the items you feel are out-of-place here.

tigarmo, line 3: I also removed some sections, like the "Requirements" one, which would be empty for this particular release.

docs/release-notes/rockcraft-1-7-0.rst
```diff
@@ -0,0 +1,101 @@
...
+
+- Updated the Maven and Ant plugins to generate the "/usr/bin/java" path to the
+  Java runtime executable, instead of the "/bin/java" symlink which conflicts
+  with the user-merged Ubuntu base.
+
+
+Fixed bugs and issues
+---------------------
+
+The following issues have been resolved in Rockcraft 1.7.1:
```
tigarmo, line 80:
```suggestion
The following issues have been resolved in Rockcraft 1.7.0:
```

docs/release-notes/rockcraft-1-7-0.rst
```diff
@@ -0,0 +1,101 @@
...
+  pebble part.
+
+
+Contributors
+------------
+
+We would like to express a big thank you to all the people who contributed to
+this release.
+
+`@dariuszd21 <https://github.com/dariuszd21>`_,
```
tigarmo, line 93: the spellchecker is complaining about all of these, heh

docs/release-notes/index.rst
medubelko: > Should this toctree be :hidden:?

Absolutely.

docs/release-notes/rockcraft-1-7-0.rst
medubelko: It was decided that the _Requirements_ section should be mandatory for each page, so it will have to stay in. It's in effect boilerplate.

docs/release-notes/rockcraft-1-7-0.rst
medubelko: Annoying. In the short-term, what would be easier?

- Use :literalref: instead
- Add a custom block directive that's a complement for the inline `:spellexception:`

medubelko, line 93: In terms of visual presentation, I'm not sure I like splitting the names across lines. I'd prefer one big paragraph.

docs/release-notes/rockcraft-1-7-0.rst
```diff
@@ -0,0 +1,101 @@
...
+~~~~~~~~~~~~~~
+
+- The Rockcraft snap has been updated from core22 to core24.
+
+
+Go extension
+~~~~~~~~~~~~
+
+- The documentation for the "go-framework" extension has been updated with
+  improved tutorial and reference.
```
medubelko, line 52:
```suggestion
A new :ref:`tutorial <build-a-rock-for-a-go-application>` and :ref:`reference
<go-framework-reference>` are available for the go-framework extension.
```

medubelko (changes requested):
This looks good! I did a deeper review of this, because I want to ensure we're off on the right foot.

I agree with your selection of major and minor changes. We'll have better examples to provide to other projects once we have more releases to document.

docs/release-notes/rockcraft-1-7-0.rst
tigarmo: It's rendered as a single paragraph; they are split over sequential lines here because of the line-length limit, which basically enforces "one per line" in this case
image

docs/release-notes/rockcraft-1-7-0.rst
tigarmo: Alright, but in this case we should review the content - the default template lists "Python 3.x or higher", but in the case of the applications the "official" means of distribution is the snap. In this case, there's no requirement on the user even having Python, but they do need a snap-capable system. Shall I draft a list of requirements here that we can possibly use for the other applications too?

docs/release-notes/rockcraft-1-7-0.rst
```diff
@@ -0,0 +1,101 @@
...
+This feature is experimental and subject to change. Feel free to provide
+feedback on the `Ubuntu Discourse`_ forum.
+
+Minor features
+--------------
+
+Rockcraft 1.7.0 brings the following minor changes.
+
+
+"init" command
```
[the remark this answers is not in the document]
tigarmo: I don't know where I got this from but I thought we preferred quotes over the inline-code markers. Guess I was wrong!

docs/release-notes/rockcraft-1-7-0.rst
```diff
@@ -0,0 +1,101 @@
...
+~~~~~~~~~~~~~~
+
+- The default base of the project file generated by the "rockcraft init" command has
+  been updated from "ubuntu@22.04" to "ubuntu@24.04".
```
tigarmo, line 39: maybe this was my confusion: should these be `"ubuntu@22.04"` or inline-code-formatted?

docs/release-notes/rockcraft-1-7-0.rst
```diff
@@ -0,0 +1,101 @@
...
+- The documentation for the various 12-factor framework extensions has been
+  updated for clarity and to better conform with Canonical's guidelines.
+
+
+Maven and Ant plugins
+~~~~~~~~~~~~~~~~~~~~~
+
+- Updated the Maven and Ant plugins to generate the "/usr/bin/java" path to the
+  Java runtime executable, instead of the "/bin/java" symlink which conflicts
+  with the user-merged Ubuntu base.
```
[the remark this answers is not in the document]
tigarmo: I had something similar to this originally, but the template points to using the past tense. I can see how it could depend on the specific instance, but won't the notes look awkward if some are in the past tense and some use the present one?

docs/release-notes/rockcraft-1-7-0.rst
```diff
@@ -0,0 +1,101 @@
...
+
+
+Fixed bugs and issues
+---------------------
+
+The following issues have been resolved in Rockcraft 1.7.1:
+
+- #659 Kubernetes ``runAsNonRoot`` incompatibility with ``run-user: _daemon_``.
```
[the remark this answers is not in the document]
tigarmo: Sure - I think we need to make it clear-er in the template that these need to be links

docs/release-notes/rockcraft-1-7-0.rst
tigarmo: updated to use `:literalref:`

docs/release-notes/rockcraft-1-7-0.rst
```diff
@@ -0,0 +1,101 @@
...
+
+Support for remote builds
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+We've added support for remote builds, which are powered by Canonical's
+Launchpad platform and allow you to build rocks for multiple architectures with
+a single command.
+
+This feature is experimental and subject to change. Feel free to provide
+feedback on the `Ubuntu Discourse`_ forum.
```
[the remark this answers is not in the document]
tigarmo: Ah I see. I have a task to document the feature on craft-application, but it won't be ready & integrated into Rockcraft by the end of this week. So I added a call to run ``rockcraft remote-build --help``, which has more information and hopefully the user can take it from there.

docs/release-notes/rockcraft-1-7-0.rst
```diff
@@ -0,0 +1,101 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~
+
+- The documentation for the various 12-factor framework extensions has been
+  updated for clarity and to better conform with Canonical's guidelines.
+
+
+Maven and Ant plugins
+~~~~~~~~~~~~~~~~~~~~~
+
+- Updated the Maven and Ant plugins to generate the "/usr/bin/java" path to the
```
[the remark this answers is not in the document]
tigarmo: I updated this paragraph to try to address both concerns (this and the agentless passive voice) - please take a look

docs/release-notes/rockcraft-1-7-0.rst
medubelko: Ah! I reacted instinctively to the code and not the preview, oops.

docs/release-notes/rockcraft-1-7-0.rst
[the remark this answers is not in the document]
medubelko: Not from me. :hear_no_evil:

docs/release-notes/rockcraft-1-7-0.rst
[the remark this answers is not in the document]
medubelko: Perfect!

docs/release-notes/rockcraft-1-7-0.rst
medubelko: One more thought about inline code, while I'm thinking about this:

Sometimes a symbol can become a real word. For example, core22 and commands like grep, which were originally intended as machine expressions, have entered general parlance. Things like that, which have become actual words everyday speech, don't need code fences. Determining whether something is a word can sometimes be difficult, so you can leave it up to me.

docs/release-notes/rockcraft-1-7-0.rst
medubelko: Yes, please! Feel free to crib from Snapcraft's system reqs, too: https://canonical-snapcraft.readthedocs-hosted.com/en/latest/reference/system-requirements/

docs/release-notes/rockcraft-1-7-0.rst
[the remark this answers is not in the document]
medubelko: The new copy makes much more sense, thanks!

docs/release-notes/rockcraft-1-7-0.rst
```diff
@@ -0,0 +1,106 @@
...
+The documentation for the various 12-factor framework extensions has been
+updated for clarity and to better conform with documentation best practices.
+
+
+Maven and Ant plugins
+~~~~~~~~~~~~~~~~~~~~~
+
+Changed the location of the Java runtime executable on rocks built using the
+Maven and Ant plugins from ``/bin/java`` to ``/usr/bin/java``, as the former
+conflicts with the user-merged Ubuntu base.
```
medubelko, line 76: I'm still not a fan of starting a sentence with a past tense verb, so let's insert ourselves:

```suggestion
We changed the location of the Java runtime executable on rocks built using the
the Maven and Ant plugins from ``/bin/java`` to ``/usr/bin/java``, as the
former conflicts with the user-merged Ubuntu base.
```

docs/release-notes/rockcraft-1-7-0.rst
tigarmo: done! Thanks for the cheatsheet

docs/release-notes/rockcraft-1-7-0.rst
tigarmo: I accepted this suggestion "manually", as I worked on the page here

docs/release-notes/rockcraft-1-7-0.rst
```diff
@@ -0,0 +1,147 @@
...
+
+OCI image configuration
+~~~~~~~~~~~~~~~~~~~~~~~
+
+When using ``run-user: _daemon_``, the value of the ``Config.User`` key in the
+generated rock has been updated from the user name (``_daemon_``) to its
+numeric ID.
+
+
+12-factor extensions
```
tigarmo, line 98: This entry is new, to reflect the PR that got merged yesterday

docs/release-notes/rockcraft-1-7-0.rst
```diff
@@ -0,0 +1,147 @@
...
+
+Support for remote builds
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+We added support for remote builds, which leverage off-site compute for
+building rocks and can build for multiple architectures with a single command.
+Remote builds are hosted on Launchpad.
+
+This feature is experimental and subject to change, and more information is
+available by running ``rockcraft remote-build --help``. Feel free to provide
```
tigarmo, line 61: Adding a note to myself here that once #778 is merged I can update these to point to the new docs

docs/release-notes/rockcraft-1-7-0.rst
```diff
@@ -0,0 +1,147 @@
...
+
+When using ``run-user: _daemon_``, the value of the ``Config.User`` key in the
+generated rock has been updated from the user name (``_daemon_``) to its
+numeric ID.
+
+
+12-factor extensions
+~~~~~~~~~~~~~~~~~~~~
+
+We added support for Gunicorn async workers on the Django and Flask 12-factor
```
medubelko, line 101: > 12-factor

Should we call them "12-factor extensions"? "Django and Flask extensions, for supporting 12-factor apps" might be clearer.

docs/release-notes/rockcraft-1-7-0.rst
```diff
@@ -0,0 +1,147 @@
...
+extensions.
+
+
+12-factor documentation
+~~~~~~~~~~~~~~~~~~~~~~~
```
medubelko, line 106:
```suggestion
12-factor app documentation
~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

docs/release-notes/index.rst
```diff
@@ -59,6 +59,9 @@ development keeps pace with the OS's new releases and support lifecycle.
 
 .. toctree::
    :maxdepth: 1
+   :hidden:
+
+   rockcraft-1-7-0
```
medubelko, line 64: Forgot this bit yesterday. This pattern will keep the TOC cleaner.

```suggestion
   Rockcraft 1.7.0 <rockcraft-1-7-0>
```

medubelko (changes requested):
Three small things, then I think we're good to go.

medubelko (approved):
:rock: on. Many thanks for being the test subject.

docs/release-notes/rockcraft-1-7-0.rst
```diff
@@ -0,0 +1,146 @@
...
+What's new
+----------
+
+Rockcraft 1.7.0 brings the following features, integrations, and improvements.
+
+
+Support for remote builds
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+We added support for remote builds, which leverage off-site compute for
```
tigarmo, line 56: @medubelko here's what I finally ended up with after much debating with myself

docs/release-notes/rockcraft-1-7-0.rst
tigarmo: reworded this: kept as Django and Flask, and updated the title to `12-factor app extensions`

medubelko (approved):
Double-approving.

bepri (approved):
These look great - I'm a big fan of this format!

PR #777, opened 2024-12-18
Use helpful messages

yanksyoon:

-----

Use a more helpful message to guide the users on experimental features.
Update documentation link which does not work anymore.


rockcraft/extensions/extension.py
```diff
@@ -87,8 +87,8 @@ def validate(self, extension_name: str) -> None:
             "ROCKCRAFT_ENABLE_EXPERIMENTAL_EXTENSIONS"
         ):
             raise errors.ExtensionError(
-                f"Extension is experimental: {extension_name!r}",
-                docs_url="https://rockcraft.io/docs/supported-extensions",
+                f"Extension is experimental: {extension_name!r}, please run with ROCKCRAFT_ENABLE_EXPERIMENTAL_EXTENSIONS=True",
+                docs_url="https://documentation.ubuntu.com/rockcraft/en/latest/reference/extensions/",
```
sergiusens, line 91: you can now refer to the slug and the correct version is shown depending on the version of rockcraft that prints the error, this is still correct and fixes the issue of referring to the deprecated link

rockcraft/extensions/extension.py
```diff
@@ -87,8 +87,8 @@ def validate(self, extension_name: str) -> None:
             "ROCKCRAFT_ENABLE_EXPERIMENTAL_EXTENSIONS"
         ):
             raise errors.ExtensionError(
-                f"Extension is experimental: {extension_name!r}",
-                docs_url="https://rockcraft.io/docs/supported-extensions",
+                f"Extension is experimental: {extension_name!r}, please run with ROCKCRAFT_ENABLE_EXPERIMENTAL_EXTENSIONS=True",
```
sergiusens, line 90: we do not say please in error messages
https://discourse.ubuntu.com/t/tone-of-voice/18883
this also needs to be in resolution, not the main message

yanksyoon: removed "please", thanks for the guidelines :). May I know what you are referring to by the main message please?

tigarmo: errors have a separate `resolution` field. Together with `doc_slug`, the error could be something like:

```python
raise errors.ExtensionError(
    f"Extension is experimental: {extension_name!r}",
    doc_slug="/reference/extensions/",
    resolution="Run with ROCKCRAFT_ENABLE_EXPERIMENTAL_EXTENSIONS=True to enable experimental extensions"
)

yanksyoon: Updated, thank you :)

tigarmo (approved):
thanks!

bepri (approved):

PR #778, opened 2024-12-18
docs: reference docs for remote builds

tigarmo:
This is loosely inspired on the existing docs for Snapcraft, but with a different set of restrictions and no concerns about 'legacy' versus 'new' builders.

Note that this is complementary to the command reference that is autogenerated from the Python code.

Fixes #597

-----

### DIFF
--- a/docs/reference/remote-build.rst
+++ b/docs/reference/remote-build.rst
@@ -1,0 +1,56 @@
+.. _ref-remote-build:
+
+
+*************
+Remote builds
+*************
+
+Remote build is a feature in Rockcraft that offloads the build process to
+`Launchpad`_'s `build farm`_ to build multiple rocks simultaneously, and for
+different architectures than your running machine.
+
+
+Pre-requisites
+--------------
+
+In order to perform remote builds, the following conditions must be met:
+
+- You must have a `Launchpad account`_, as the remote builds are performed on
+  Launchpad.
+- The Rockcraft project must be version-controlled by Git. This is because
+  Rockcraft uses a Git-based workflow to upload the project to Launchpad.
+- The repository hosting the Rockcraft project must not be a shallow clone,
+  because Git does not support pushing shallow clones.
+
+
+Overview
+--------
+
+Remote builds are launched by running ``rockcraft remote-build``. Rockcraft will
+upload the Git repository on the current working directory to Launchpad on your
+behalf, under your account. Next, it will trigger builds for the Rockcraft
+project present on the root of the repository and continuously monitor the
+status of the new builds.
+
+Once all builds are done (either through a successful build or a failure), the
+rock files will be downloaded to the current directory, together with the build
+logs.
+
+
+Limitations
+-----------
+
+The following is a list of the current limitations of the remote build feature,
+which are planned to be addressed in the future:
+
+- The prospective rock must be open source and public, because the remote builds
+  will be publicly available.
+- All architectures defined on the project's ``platforms`` entry - either
+  explicitly via the ``build-on`` key or implicitly through the platform
+  shorthand - will be built. There is currently no way to restrict the set of
+  platforms to build remotely.
+
+
+.. _`Launchpad`: https://launchpad.net/
+.. _`build farm`: https://launchpad.net/builders
+.. _`Launchpad account`: https://launchpad.net/+login
--- a/docs/reference/remote-build.rst
+++ b/docs/reference/remote-build.rst
@@ -5,13 +5,13 @@
 Remote builds
 *************
 
-Remote build is a feature in Rockcraft that offloads the build process to
-`Launchpad`_'s `build farm`_ to build multiple rocks simultaneously, and for
-different architectures than your running machine.
+Remote builds offload rock builds to the `build farm`_ hosted by `Launchpad`_.
+With remote builds, you can assemble multiple rocks simultaneously and build
+for all supported architectures.
 
 
-Pre-requisites
---------------
+Prerequisites
+-------------
 
 In order to perform remote builds, the following conditions must be met:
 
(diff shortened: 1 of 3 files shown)


docs/reference/remote-build.rst
tigarmo, line 1: https://canonical-ubuntu-documentation-library--778.com.readthedocs.build/rockcraft/en/778/reference/remote-build/

docs/reference/remote-build.rst
```diff
@@ -0,0 +1,56 @@
+.. _ref-remote-build:
+
+
+*************
+Remote builds
```
tigarmo, line 5: @medubelko in the end I opted to document this in Rockcraft now, so that the release has the docs, and then worry about upstreaming this to craft-application in the near future. Once we review and merge this I can point to this new doc on the 1.7.0 release notes

docs/reference/remote-build.rst
```diff
@@ -0,0 +1,56 @@
...
+*************
+
+Remote build is a feature in Rockcraft that offloads the build process to
+`Launchpad`_'s `build farm`_ to build multiple rocks simultaneously, and for
+different architectures than your running machine.
+
+
+Pre-requisites
+--------------
```
medubelko, line 14:
```suggestion
Prerequisites
-------------
```

docs/reference/remote-build.rst
```diff
@@ -0,0 +1,56 @@
+.. _ref-remote-build:
+
+
+*************
+Remote builds
+*************
+
+Remote build is a feature in Rockcraft that offloads the build process to
+`Launchpad`_'s `build farm`_ to build multiple rocks simultaneously, and for
+different architectures than your running machine.
```
medubelko, line 10: It's best not to put two links next to each other.

```suggestion
Remote builds offload rock builds to the `build farm`_ hosted by `Launchpad`_.
With remote builds, you can assemble multiple rocks simultaneously and build
for all supported architectures.
```

docs/reference/remote-build.rst
```diff
@@ -0,0 +1,56 @@
...
+
+- You must have a `Launchpad account`_, as the remote builds are performed on
+  Launchpad.
+- The Rockcraft project must be version-controlled by Git. This is because
+  Rockcraft uses a Git-based workflow to upload the project to Launchpad.
+- The repository hosting the Rockcraft project must not be a shallow clone,
+  because Git does not support pushing shallow clones.
+
+
+Overview
```
medubelko, line 26: How about "How it works" or simply "Remote build process"?

I don't think the content quite fits the definition of overview, and typically an overview would come before prerequisites.

docs/reference/remote-build.rst
```diff
@@ -0,0 +1,56 @@
...
+-----------
+
+The following is a list of the current limitations of the remote build feature,
+which are planned to be addressed in the future:
+
+- The prospective rock must be open source and public, because the remote builds
+  will be publicly available.
+- All architectures defined on the project's ``platforms`` entry - either
+  explicitly via the ``build-on`` key or implicitly through the platform
+  shorthand - will be built. There is currently no way to restrict the set of
```
medubelko, line 50: Sphinx parses double-hyphens as en dashes (–).

```suggestion
- All architectures defined in the project's ``platforms`` entry -- either
  explicitly through the ``build-on`` key or implicitly through the platform
  shorthand -- are built. There's currently no way to restrict the set of
```

medubelko, line 50: "platform shorthand" might be difficult for newer users to intuit.

docs/reference/remote-build.rst
```diff
@@ -0,0 +1,56 @@
...
+
+
+Limitations
+-----------
+
+The following is a list of the current limitations of the remote build feature,
+which are planned to be addressed in the future:
+
+- The prospective rock must be open source and public, because the remote builds
+  will be publicly available.
```
medubelko, line 47:
```suggestion
- The prospective rock must be open source and public, because remote builds
  are publicly available.
```

medubelko (changes requested):
If any of these delay the merge by too much, let me know.

docs/reference/remote-build.rst
tigarmo: there's a subtle difference here though: it's not that "all" remote builds are public, just the ones that Rockcraft currently does.

Contrast this with [Snapcraft](https://canonical-snapcraft.readthedocs-hosted.com/en/latest/explanation/remote-build/#open-vs-closed-source): there users can do a bit more work to do remote builds for private projects.

How about something longer like "...because the remote builds triggered by Rockcraft are publicly available"?

docs/reference/remote-build.rst
tigarmo: I see your point. I think a section like "How it works" and "Remote build process" feels a bit... generic and redundant with the page itself, so I removed this section and put this paragraph in the "introduction", before the Prerequisites section. Looks like it might flow better like this.

Let me know what you think!

docs/reference/remote-build.rst
tigarmo: You're right, and it's a bit too much detail for this point anyway. Reworded this entry.

docs/reference/remote-build.rst
tigarmo: ^did this

tigarmo:
@lengau I added you to double-check if my docs make sense (from a technical pov) and also so we can start looking into how cleanly these docs also apply to Charmcraft

lengau (approved):
LGTM!

I wonder if we can get a template that lets us replace something with the app's name/command?

docs/reference/remote-build.rst
medubelko: Looks good. :)

docs/reference/remote-build.rst
medubelko: :+1:

docs/reference/remote-builds.rst
```diff
@@ -0,0 +1,50 @@
...
+
+Limitations
+-----------
+
+The following is a list of the current limitations of the remote build feature,
+which are planned to be addressed in the future:
+
+- The prospective rock must be open source and public, because the remote builds
+  triggered by Rockcraft are publicly available.
+- All architectures defined in the project's ``platforms`` entry are built --
```
medubelko, line 44:
```suggestion
- All architectures defined in the project's ``platforms`` key are built --
```

medubelko (approved):
LGTM. One more non-blocking suggestion.

tigarmo:
> LGTM!
> 
> I wonder if we can get a template that lets us replace something with the app's name/command?

maybe substitutions? https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#substitution-definitions

PR #779, opened 2024-12-19
Update link to Pebble layer specification

benhoyt:
As of a few months ago we have a proper docs site and the README is minimal, so update these links to https://canonical-pebble.readthedocs-hosted.com/en/latest/reference/layer-specification/


rockcraft/pebble.py
```diff
@@ -59,7 +59,7 @@ class Check(CraftBaseModel):
     """Lightweight schema validation for a Pebble checks.
 
     Based on
-    https://github.com/canonical/pebble#layer-specification
+    https://canonical-pebble.readthedocs-hosted.com/en/latest/reference/layer-specification/
```
tigarmo, line 62: Thanks for the PR Ben. I would like to point to a `stable` docs page (as opposed to this `latest`, which I assume is built off of `main`) to protect us a bit against dev changes in Pebble. Do you know who's in charge of the Pebble RTD page? The `stable` version would need to be enabled/activated.

benhoyt: > Do you know who's in charge of the Pebble RTD page?

That would be me. :-)

Yeah, we don't version the Pebble docs right now, but we consider the "latest" to be stable. We don't make breaking changes to Pebble in minor releases (which is all we do). So I think this is a good question, but that you're fine using the "latest" links here -- that's what we do in the Juju docs as well.

tigarmo: Ah, thanks for the update. My concern was less about breaking changes, and more about new compatible features. For example - let's say that Pebble gets a new keyword on the layer specification format. That new keyword will be available on the `latest` docs, but not on the version of Pebble used by Rockcraft.

But if you say we're fine using "latest" then that's good enough for me. This change is strictly better than what we were doing (pointing straight to main) anyway.

tigarmo (approved):
Thanks, and happy holidays!

rockcraft/pebble.py
cjdcordeiro: Indeed I agree with @tigarmo . 

However, that was the case already with the previous link anyway, so this change isn't introducing it. So I think it's ok for now.

cjdcordeiro (approved):

PR #781, opened 2024-12-19
[WIP] feat: express JS

yanksyoon:

-----


rockcraft/extensions/express.py
```diff
@@ -0,0 +1,121 @@
...
+    @override
+    def get_supported_bases() -> Tuple[str, ...]:
+        """Return supported bases."""
+        return "bare", "ubuntu@22.04", "ubuntu:22.04", "ubuntu@24.04", "ubuntu:24.04"
+
+    @staticmethod
+    @override
+    def is_experimental(base: str | None) -> bool:
+        """Check if the extension is in an experimental state."""
+        return False
```
jdkandersson, line 38: Should be `True` until we have a production deployment please

rockcraft/extensions/express.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+class ExpressJSFramework(Extension):
+    """An extension for constructing Javascript applications based on the ExpressJS framework."""
+
+    @staticmethod
+    @override
+    def get_supported_bases() -> Tuple[str, ...]:
+        """Return supported bases."""
+        return "bare", "ubuntu@22.04", "ubuntu:22.04", "ubuntu@24.04", "ubuntu:24.04"
```
jdkandersson, line 32: I think we can just have the `@` notation. Could you please check which version of node comes with 22.04 and how commonly it is used? Then we can decide what we support

rockcraft/extensions/express.py
```diff
@@ -0,0 +1,121 @@
...
+    @override
+    def is_experimental(base: str | None) -> bool:
+        """Check if the extension is in an experimental state."""
+        return False
+
+    @property
+    @override
+    def framework(self) -> str:
+        """Return the framework name, i.e. expressjs."""
+        return "expressjs"
```
jdkandersson, line 44: Check if you need this and this should be `expressjs-framework` please

rockcraft/extensions/express.py
jdkandersson, line 1: Thanks, in general perhaps a better reference is https://github.com/canonical/rockcraft/blob/main/rockcraft/extensions/fastapi.py as it is stand-alone. Flask is unique because it shares a common parent class for Django

jdkandersson: We'll need:
* Docs (tutorial, reference)
* Spread test for the tutorial
* Unit tests

Also check anywhere that mentions `fast-api` framework and check if we need to do something similar for `expressjs-framework`

yanksyoon:
Closing (Opened https://github.com/canonical/rockcraft/pull/784)

PR #784, opened 2025-01-02
feat: expressjs framework

yanksyoon:

-----

Adds support for [ExpressJS](https://expressjs.com/) framework.

### Key Decisions

- Support for `bare` and `24.04` only. Drop support for 22.04 (jammy)
  - Jammy comes with an old version of Node which does not support `??` operator commonly used in JS projects. This means that the projects initialized with `express-generator` tooling (assumed to be the case) would not work.
- Include runtime slices in the `expressjs-framework/install-app > stage-packages` part.
  - A separate `expressjs-framework/runtime` exists to cater to apt packages, since slices and apt packages cannot be used together by design. See https://github.com/canonical/rockcraft/issues/785.
- Do not include user customizable part for runtime slices/packages in the template
  - There is an issue with the NPM plugin which conflicts with the `lib` directory of apt/slice installs. Remove the offering until the issue is resolved. See https://github.com/canonical/rockcraft/issues/790.
- Provide Ubuntu packaged Node/NPM by default, allow user to specify Node version if desired.
  - To follow the principles of Ubuntu LTS, by default use the NodeJS and NPM provided by apt/slices but allow users to customize the behavior through overriding the NPM plugin's `npm-include-node` and `npm-node-version` fields.
- Include all paths under the `app/` directory.
  - Users should not include/add paths which are not desired in the `app` path (e.g. .env files)


rockcraft/pebble.py
```diff
@@ -137,9 +137,7 @@ class Pebble:
         "stage-snaps": ["pebble/latest/stable"],
         # We need this because "services" is Optional, but the directory must exist
         "override-prime": str(
-            "craftctl default\n"
-            f"mkdir -p {PEBBLE_LAYERS_PATH}\n"
-            f"chmod 777 {PEBBLE_PATH}"
+            f"craftctl default\nmkdir -p {PEBBLE_LAYERS_PATH}\nchmod 777 {PEBBLE_PATH}"
```
yanksyoon, line 140: Linting change only

rockcraft/templates/expressjs-framework/rockcraft.yaml.j2
```diff
@@ -0,0 +1,53 @@
...
+# script  defined in the package.json's scripts section under the "run" key.
+extensions:
+    - expressjs-framework
+
+# uncomment the sections you need and adjust according to your requirements.
+# parts:  # you need to uncomment this line to add or update any part.
+#   expressjs-framework/install-app:
+#     prime:
+#       # by default, only the files in bin/, public/, routes/, views/, app.js,
+# 	    # package.json package-lock.json are copied into the image.
```
yanksyoon, line 33: Indentation

rockcraft/templates/expressjs-framework/rockcraft.yaml.j2
```diff
@@ -0,0 +1,53 @@
...
+# see https://documentation.ubuntu.com/rockcraft/en/latest/explanation/chisel/
+
+# add this part if you want to add packages slices to your image.
+# you can find a list of packages slices at https://github.com/canonical/chisel-releases
+#   expressjs-framework/runtime:
+#     plugin: nil
+#     stage-packages:
+#       # list the required package slices for your expressjs application below.
+#       # for example, for the slice libs of libpq5:
+#       - libpq5_libs
```
yanksyoon, line 53: Wait until issue mentioned above (matrix) is resolved.

tests/spread/rockcraft/extension-expressjs/app/app.js
yanksyoon, line 1: Generated by `express-generator`

tests/spread/rockcraft/extension-expressjs/app/bin/www
yanksyoon, line 1: Generated by `express-generator`

tests/spread/rockcraft/extension-expressjs/app/package.json
yanksyoon, line 1: Generated by `express-generator`

tests/spread/rockcraft/extension-expressjs/app/public/stylesheets/style.css
yanksyoon, line 1: Generated by `express-generator`

tests/spread/rockcraft/extension-expressjs/app/routes/index.js
yanksyoon, line 1: Generated by `express-generator`

tests/spread/rockcraft/extension-expressjs/app/routes/users.js
yanksyoon, line 1: Generated by `express-generator`

tests/spread/rockcraft/extension-expressjs/app/views/error.jade
yanksyoon, line 1: Generated by `express-generator`

tests/spread/rockcraft/extension-expressjs/app/views/index.jade
yanksyoon, line 1: Generated by `express-generator`

tests/spread/rockcraft/extension-expressjs/app/views/layout.jade
yanksyoon, line 1: Generated by `express-generator`

tests/unit/extensions/test_expressjs.py
```diff
@@ -0,0 +1,224 @@
...
+            [],
+            {
+                "lib/node_modules/test-expressjs-project/.npmrc": "app/.npmrc",
+                "lib/node_modules/test-expressjs-project/app.js": "app/app.js",
+                "lib/node_modules/test-expressjs-project/package.json": "app/package.json",
+                "lib/node_modules/test-expressjs-project/node_modules": "app/node_modules",
+            },
+            id="single file defined",
+        ),
+    ],
```
yanksyoon, line 178: Single file, no files, multiple files

rockcraft/templates/expressjs-framework/rockcraft.yaml.j2
yanksyoon, line 1: Add link to documentation on libs that are required on charm integrations (e.g. pg)

yanksyoon (commented):
Initial self review

rockcraft/extensions/expressjs.py
```diff
@@ -0,0 +1,230 @@
...
+    ]
+    BUILD_GENERATED_DIRS = ["node_modules", ".npmrc"]
+    RUNTIME_SLICES = ["ca-certificates_data"]
+    BARE_RUNTIME_SLICES = [
+        "bash_bins",
+        "coreutils_bins",
+        "libc6_libs",
+        "libnode109_libs",
+    ]
+
```
jdkandersson, line 52: Check whether these are needed or can be inlined

rockcraft/extensions/expressjs.py
```diff
@@ -0,0 +1,230 @@
...
+          - platform: amd64
+          - services: a service to run the ExpressJS server
+          - parts: see ExpressJSFramework._gen_parts
+        """
+        self._check_project()
+
+        snippet: dict[str, Any] = {
+            "run-user": "_daemon_",
+            "services": {
+                "app": {
```
jdkandersson, line 81: Let's change this to the name of the extension

rockcraft/extensions/expressjs.py
```diff
@@ -0,0 +1,230 @@
...
+                    "override": "replace",
+                    "command": "npm start",
+                    "startup": "enabled",
+                    "on-success": "shutdown",
+                    "on-failure": "shutdown",
+                    "working-dir": "/app",
```
jdkandersson, line 87: Check the other frameworks and align please

jdkandersson (commented):
Let's review this extension against fastAPI and go and make sure they are aligned in approach please :)

docs/tutorial/expressjs.rst
```diff
@@ -0,0 +1,418 @@
...
+
+.. _update-expressjs-application:
+
+Update the ExpressJS application
+================================
+
+As a final step, let's update our application. For example,
+we want to add a new ``/time`` endpoint which returns the current time.
+
+Start by opening the ``time.js`` file in a text editor and update the code to
```
erinecon, line 307: For my understanding, did the `time.js` file exist prior to this step, or does this step create the `time.js` file? 

I ask because, if the user is creating the file in this step, then I think this sentence should start with "Start by creating the ``time.js`` file using a text editor..."

docs/tutorial/expressjs.rst
```diff
@@ -0,0 +1,418 @@
...
+
+Place ``time.js`` file into the appropriate ``routes/`` directory. Import the
+time route from the the main ``app.js`` file and update the code to look like
+the following:
+
+.. literalinclude:: code/expressjs/app.js
+    :language: javascript
+
+Notice the addition of timerouter import and the registration of the ``/time``
+endpoint.
```
erinecon, line 321: Along with this notice, you can also use `emphasize-lines` to highlight the specific lines that the user should pay attention to. I may not have the exact lines that you would like to highlight, but perhaps the literalinclude would look something like

```
.. literalinclude:: code/expressjs/app.js
    :language: javascript
    :emphasize-lines: 9,25
```

erinecon (commented):
Hi, I took a brief look at what's in the tutorial so far, and I left two comments. 

I have some other comments that I excluded because they will affect all of the tutorials. I'd rather keep the set of tutorials consistent for now and update them all in one go.

docs/tutorial/expressjs.rst
yanksyoon: that is pretty cool, thanks!

docs/explanation/extensions/expressjs-framework.rst
```diff
@@ -0,0 +1,100 @@
+.. _expressjs-framework-explanation:
+
+expressjs-framework
+===================
```
erinecon, line 4: I feel like the title could be more specific, maybe "Ubuntu base and Node combinations for the expressjs-framework extension"
Unless we think there will be more sections in this document? It's exciting because we haven't had any Explanation material for the extensions yet :)

docs/explanation/extensions/expressjs-framework.rst
```diff
@@ -0,0 +1,100 @@
...
+
+When using the expressjs-framework extension, there are four different cases for
+customizing the Ubuntu base and the Node version to be included.
```
erinecon, line 7:
```suggestion
customising the Ubuntu base and the Node version to be included.
```
For now the spelling checks seem to still be for British English, but perhaps @medubelko can advise further

docs/reference/extensions/expressjs-framework.rst
```diff
@@ -0,0 +1,71 @@
...
+
+.. code-block:: yaml
+
+  parts:
+    expressjs-framework/install-app:
+      npm-include-node: true
+      npm-node-version: 20.12.2
+
+For more examples of npm-node-version options, see:
+https://documentation.ubuntu.com/rockcraft/en/1.5.3/common/craft-parts/reference/plugins/npm_plugin/#examples
```
erinecon, line 42:
```suggestion
For more examples of npm-node-version options, see the [Rockcraft documentation on the NPM Plugin](https://documentation.ubuntu.com/rockcraft/en/1.5.3/common/craft-parts/reference/plugins/npm_plugin/#examples).
```

docs/reference/extensions/expressjs-framework.rst
```diff
@@ -0,0 +1,71 @@
...
+      stage-packages:
+        # when using the default provided node version, the npm package is
+        # required.
+        # - npm
+        # list required packages or slices for your ExpressJS application below.
+        - libpq-dev
+
+When using the NPM plugin installed Node and NPM, additional runtime packages
+is currently unsupported due to an issue with ``lib`` path permission error. See
+https://github.com/canonical/rockcraft/issues/790 for more information.
```
erinecon, line 66:
```suggestion
is currently unsupported due to an issue with ``lib`` path permission error. See
[Issue #790](https://github.com/canonical/rockcraft/issues/790) for more information.
```

erinecon (commented):
Thanks for adding more documentation!

docs/explanation/extensions/expressjs-framework.rst
yanksyoon: I think this should be about the explanations of decisions/architectures behind the expressjs-framework. It can include more things in the future (perhaps things like why extra libs are unable to be installed due to issues with NPM plugin).
The reason why I think it should still stay is because I think we should have one documentation per framework - and if the title is specific to the topic, we might over-populate this part of the documentation section.
What do you think?

docs/explanation/extensions/expressjs-framework.rst
yanksyoon: A little confused here because directives like `organize` are spelt in American English, but thanks for pointing this out!

docs/explanation/extensions/expressjs-framework.rst
erinecon: Thanks for the additional context. If there are potentially more sections on the way, then I'm fine with the title remaining so generic.

docs/tutorial/code/expressjs/task.yaml
```diff
@@ -0,0 +1,136 @@
...
+  # [docs:curl-expressjs-bare-rock-end]
+
+  # [docs:stop-docker-chisel]
+  sudo docker stop expressjs-hello-world
+  sudo docker rmi expressjs-hello-world:0.1-chiselled
+  # [docs:stop-docker-chisel-end]
+
+  ROCKCRAFT_ENABLE_EXPERIMENTAL_EXTENSIONS=True rockcraft clean
+  mv time.js app/routes/
+  mv app.js app/
```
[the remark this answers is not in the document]
yanksyoon: I've kept the file names `time.js` to match the route filenames of other frameworks and called the app file `time_app.js`. Would this approach sound reasonable?

docs/tutorial/code/expressjs/task.yaml
```diff
@@ -0,0 +1,136 @@
...
+
+  npm start &
+  cd ..
+  retry -n 5 --wait 2 curl --fail localhost:3000
+
+  # [docs:curl-expressjs]
+  curl localhost:3000
+  # [docs:curl-expressjs-end]
+
+  kill $(lsof -t -i:3000)
```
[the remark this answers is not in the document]
yanksyoon: This is because the `kill $!` only kills the npm start command which launches additional children processes via sh command which then executes node command.
To effectively kill all commands, i've updated the code to:
```
kill $! # kill most recent backgrounded process
kill $(lsof -t -i:3000) # kill process taking up port 3000 (node)
```

docs/tutorial/code/expressjs/time_app.js
jdkandersson, line 1:
```js
// [docs:append-lines]
var timeRouter = require('./routes/time');
app.use('/time', timeRouter);
// [docs:append-lines-end]
export app;
```

jdkandersson: And then use the start-after, end-before directive

docs/explanation/extensions/expressjs-framework.rst
```diff
@@ -0,0 +1,100 @@
...
+The following example uses the Ubuntu 24.04 base and Node from Ubuntu packages.
+
+.. code-block:: yaml
+
+    base: ubuntu@24.04
+    extensions:
+    - expressjs-framework
+    parts:
+        expressjs-framework/install-app:
+            npm-include-node: false
```
jdkandersson, line 48: Perhaps remove this since it is the default

jdkandersson: And perhaps move this as the first one

docs/explanation/extensions/expressjs-framework.rst
```diff
@@ -0,0 +1,100 @@
...
+
+The following example uses the bare base and Node from Ubuntu packages.
+
+.. code-block:: yaml
+
+    base: bare
+    build-base: ubuntu@24.04
+    parts:
+        expressjs-framework/install-app:
+            npm-include-node: false
```
jdkandersson, line 88: We can remove this and have it before the other bare one

docs/tutorial/code/django/task.yaml
```diff
@@ -131,7 +131,7 @@ execute: |
   retry -n 5 --wait 2 curl --fail localhost:8000/time/
 
   # [docs:curl-time]
-  curl localhost:8000/time/
+  curl --fail localhost:8000/time/
```
jdkandersson, line 134: Could we do this for all extensions and all curl commands please

docs/tutorial/code/expressjs/package.json
jdkandersson, line 1: Can we remove this? It should get auto generated

docs/tutorial/code/fastapi/task.yaml
```diff
@@ -117,7 +117,7 @@ execute: |
   retry -n 5 --wait 2 curl --fail localhost:8000/time
 
   # [docs:curl-time]
-  curl localhost:8000/time
+  curl --fail localhost:8000/time
```
jdkandersson, line 120: Also for the other curls for this file

docs/tutorial/code/flask/task.yaml
```diff
@@ -118,7 +118,7 @@ execute: |
   retry -n 5 --wait 2 curl --fail localhost:8000/time
 
   # [docs:curl-time]
-  curl localhost:8000/time
+  curl --fail localhost:8000/time
```
jdkandersson, line 121: Also here for the other curls

tests/spread/rockcraft/extension-expressjs/task.yaml
```diff
@@ -0,0 +1,46 @@
...
+  docker images "${IMAGE}" | MATCH "${NAME}"
+
+  # ensure container doesn't exist
+  docker rm -f "${NAME}-container"
+
+  # test the default expressjs service
+  docker run --name "${NAME}-container" -d -p 8137:3000 "${IMAGE}"
+  retry -n 5 --wait 2 curl localhost:8137
+  http_status=$(curl -s -o /dev/null -w "%{http_code}" localhost:8137)
+  [ "${http_status}" -eq 200 ]
```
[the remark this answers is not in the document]
jdkandersson: let's check we get the expected version of node

tests/unit/extensions/test_expressjs.py
```diff
@@ -0,0 +1,217 @@
...
+    with pytest.raises(ExtensionError) as exc:
+        extensions.apply_extensions(tmp_path, expressjs_input_yaml)
+    assert str(exc.value) == "missing package.json file"
+    assert str(exc.value.doc_slug) == "/reference/extensions/expressjs-framework"
+
+
+@pytest.mark.parametrize(
+    "package_json_contents, error_message",
+    [
+        ("{}", "missing start script"),
```
jdkandersson, line 117: Perhaps include not a dict and a name not a string

tests/unit/extensions/test_expressjs.py
```diff
@@ -0,0 +1,217 @@
...
+    "name": "{expressjs_project_name}",
+    "scripts": {{
+        "start": "node ./bin/www"
+    }}
+}}"""
+    )
+
+
+@pytest.mark.usefixtures("expressjs_extension", "package_json_file")
+def test_expressjs_extension_default(
```
jdkandersson, line 63: let's put the 4 scenarios here:

* 24.04 base and node not customised
* 24.04 base and node customised
* bare base and node not customised
* bare base and node version customised

Then we don't need the sub functions. It would probably be best to do this using parametrize

jdkandersson (commented):
Thanks, this is close now!

jdkandersson (approved):

docs/tutorial/expressjs.rst
```diff
@@ -0,0 +1,420 @@
...
+
+and then using the same ``curl`` request:
+
+.. literalinclude:: code/expressjs/task.yaml
+    :language: text
+    :start-after: [docs:curl-expressjs-bare-rock]
+    :end-before: [docs:curl-expressjs-bare-rock-end]
+    :dedent: 2
+
+Unsurprisingly, the ExpressJS application should still respond with
```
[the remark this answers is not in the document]
erinecon: Apologies for my missing of this message until now! 

Personally I think we can remove "Unsurprisingly" -- I feel like it's a bit unnecessary. I do think it's a fair point that some users may be surprised that everything still works the same, so I don't have a strong opinion about whether we include "Unsurprisingly" here or not. It's a personal preference. 

@medubelko may have a different opinion -- feel free to weigh in if you have any thoughts!

docs/reference/extensions/expressjs-framework.rst
```diff
@@ -0,0 +1,57 @@
...
+
+Project requirements
+====================
+
+There are 3 requirements to be able to use the ``expressjs-framework``
+extension:
+
+1. The application should reside in the ``app`` directory.
+2. The application should have a ``package.json`` file.
+3. The ``package.json`` file should defined the ``start`` script.
```
erinecon, line 25:
```suggestion
3. The ``package.json`` file should define the ``start`` script.
```

docs/reference/extensions/expressjs-framework.rst
```diff
@@ -0,0 +1,57 @@
...
+.. code-block:: yaml
+
+  parts:
+    expressjs-framework/install-app:
+      npm-include-node: true
+      npm-node-version: 20.12.2
+
+For more examples of npm-node-version options, see: https://documentation.\
+ubuntu.com/rockcraft/en/1.5.3/common/craft-parts/reference/plugins/npm_plugin/\
+#examples
```
erinecon, line 43: The link doesn't seem to render correctly in the RTD build.
I suggest the following alternative:
```
`Rockcraft | NPM Plugin | Examples <https://documentation.ubuntu.com/rockcraft/en/1.5.3/common/craft-parts/reference/plugins/npm_plugin/#examples>`_
```

docs/tutorial/expressjs.rst
```diff
@@ -0,0 +1,418 @@
...
+    :dedent: 2
+
+The output should list the ExpressJS container image, along with its tag, ID and
+size:
+
+..  code-block:: text
+    :class: log-snippets
+
+    REPOSITORY              TAG       IMAGE ID       CREATED       SIZE
+    expressjs-hello-world   0.1       30c7e5aed202   2 weeks ago   304MB
```
erinecon, line 151: Nitpick: Another option for this code block would be to use the `.. terminal::` directive.

docs/tutorial/expressjs.rst
```diff
@@ -0,0 +1,418 @@
...
+As a result, :ref:`pebble_explanation_page` will give us the logs for the
+``expressjs`` service running inside the container.
+We should expect to see something similar to this:
+
+..  code-block:: text
+    :class: log-snippets
+
+    app@0.0.0 start
+    node ./bin/www
+    GET / 200 62.934 ms - 170
```
erinecon, line 199: Another place where `.. terminal::` might be good to use

erinecon (commented):
Thanks for adding the ExpressJS docs!

docs/tutorial/expressjs.rst
```diff
@@ -0,0 +1,418 @@
...
+    :dedent: 2
+
+The ExpressJS application should again respond with the ``Welcome to Express``
+HTML.
+
+View the application logs
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+When deploying the ExpressJS rock, we can always get the application logs via
+``pebble``:
```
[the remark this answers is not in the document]
yanksyoon: Updated along with all other tutorials that use the same sentence, thank you!

docs/tutorial/expressjs.rst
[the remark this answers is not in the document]
yanksyoon: Removed the vocab "Unsurprisingly" for now - it should now be surprising! (just kidding, it sounds very neutral without it now :) )

docs/tutorial/expressjs.rst
erinecon: Hey! You can find more information here: https://canonical-documentation-with-sphinx-and-readthedocscom.readthedocs-hosted.com/style-guide/#terminal-output

For the above snippet, the change would look like

```
.. terminal::

    REPOSITORY              TAG       IMAGE ID       CREATED       SIZE
    expressjs-hello-world   0.1       30c7e5aed202   2 weeks ago   304MB
```

docs/reference/extensions/expressjs-framework.rst
```diff
@@ -0,0 +1,57 @@
...
+.. code-block:: yaml
+
+  parts:
+    expressjs-framework/install-app:
+      npm-include-node: true
+      npm-node-version: 20.12.2
+
+For more examples of npm-node-version options, see: `Rockcraft | NPM Plugin |
+Examples <https://documentation.ubuntu.com/rockcraft/en/1.5.3/common/\
+craft-parts/reference/plugins/npm_plugin/#examples>`_.
```
erinecon, line 43: This link seems to be broken in the RTD rendering. I think the issue is the inclusion of the `\` -- I think the problem will go away if you remove it from the link.
```suggestion
For more examples of npm-node-version options, see: `Rockcraft | NPM Plugin |
Examples <https://documentation.ubuntu.com/rockcraft/en/1.5.3/common/
craft-parts/reference/plugins/npm_plugin/#examples>`_.
```

docs/tutorial/expressjs.rst
```diff
@@ -0,0 +1,422 @@
...
+.. literalinclude:: code/expressjs/task.yaml
+    :language: bash
+    :start-after: [docs:init-app]
+    :end-before: [docs:init-app-end]
+    :dedent: 2
+
+Run the ExpressJS application from within the ``app/`` directory using
+.. code-block:: bash
+
+    npm start
```
erinecon, line 45:
```suggestion
Run the ExpressJS application from within the ``app/`` directory using

.. code-block:: bash

    npm start
```

jdkandersson (approved):

jdkandersson (approved):

erinecon (approved):
Thanks for adding these docs, Yanks! Besides the link change and the nitpicks about the `terminal` directive, I think everything looks great :rocket:

docs/tutorial/expressjs.rst
yanksyoon: Thank you! Updated all other documentation to match this styling :)

rockcraft/extensions/expressjs.py
```diff
@@ -0,0 +1,230 @@
...
+                    "environment": {
+                        "NODE_ENV": "production",
+                    },
+                }
+            },
+        }
+        if not self.yaml_data.get("services", {}).get("expressjs", {}).get("command"):
+            snippet["services"]["expressjs"]["command"] = "npm start"
+
+        snippet["parts"] = {
```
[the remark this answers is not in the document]
yanksyoon: This style was used to match other styles of other extensions (fastapi, go, gunicorn).
Would it be preferred that `get_parts_snippet()` is used?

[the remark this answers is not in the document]
tigarmo: I suppose it's better to keep it consistent with the other extensions

tests/spread/rockcraft/extension-expressjs/task.yaml
```diff
@@ -0,0 +1,72 @@
...
+  ROCK_FILE="${NAME}_0.1_amd64.rock"
+  IMAGE="${NAME}:0.1"
+
+  sudo apt-get update -y && sudo apt-get install npm -y
+  npm install -g express-generator@4
+  express app
+  cd app && npm install && cd ..
+
+  run_rockcraft init --name "${NAME}" --profile expressjs-framework
+  sed -i "s/^name: .*/name: ${NAME}/g" rockcraft.yaml
```
tigarmo, line 18: weird, shouldn't the name be right already (because of `--name`)?
Can you confirm whether this is a bug in rockcraft?

tests/spread/rockcraft/extension-expressjs/task.yaml
```diff
@@ -0,0 +1,72 @@
...
+  sed -i "s/^base: .*/base: ${SCENARIO//-/@}/g" rockcraft.yaml
+  if [ "${SCENARIO}" != "bare" ]; then
+    sed -i "s/^build-base: .*/build-base: ${SCENARIO//-/@}/g" rockcraft.yaml
```
tigarmo, line 21: this is a bit weird - where does the original `build-base`, that you're replacing here, come from?

jdkandersson (approved):

tests/spread/rockcraft/extension-expressjs/task.yaml
yanksyoon: Updated the template to be by default: bare base
If the scenario is not bare, it will change the build-base to match the scenario's build base from templates embedded 24.04 base.
Thank you!

tests/spread/rockcraft/extension-expressjs/task.yaml
yanksyoon: You're right, removed!

tests/spread/rockcraft/extension-expressjs/task.yaml
yanksyoon: Actually, I'm wondering what the default build base should be in this case - bare or 24.04.

The reason for using the bare as default is to be **production ready** (with a small image with less surface area)
The reason for using the 24.04 as default is **for the tutorial** where the initialization tool defaults to 24.04 and we guide them to using the slimmed version of the image (by switching to bare).

What do you think should be the default? I think we could use the bare base and start the tutorial by changing the base to 24.04, although that might seem a little weird until the user finishes all the tutorial steps.

Perhaps @erinecon and @jdkandersson would also be interested in this discussion point :)

yanksyoon:
Hello @tigarmo! I'm having a bit of issues trying to get the tests passing on the CI.
Here are the steps to test the failing CI test locally (`tests/spread/rockcraft/extension-expressjs:bare`)

1. Modify the `spread.yaml` to use multipass instaed of Google backend (copy paste-able pastebin: https://pastebin.canonical.com/p/y3RHnNWBDc/)
2. Checkout the current branch (`git remote add yanks https://github.com/yanksyoon/rockcraft.git && git checkout feat/expressjs-framework`)
3. Build snapcraft (`snapcraft --use-lxd`)
  * to test rock manually `sudo snap install *.snap --dangerous`
5. Move the built snap to tests (`mv *.snap tests/`)
6. Run test `spread tests/spread/rockcraft/extension-expressjs:bare`

The test fails with `<kill-timeout reached>`.
Here's the full log: https://pastebin.canonical.com/p/yrfh5YrbP5/

Would you know if there would be a way for me to increase the timeout please?

yanksyoon:
I figured out how to increase the timeout and investigated further locally - 
```
E: Failed to fetch http://archive.ubuntu.com/ubuntu/pool/universe/n/npm/npm_9.2.0%7eds1-2_all.deb  Unable to connect to archive.ubuntu.com:http: [IP: 91.189.91.82 80]
```
I'm quite surprised I am getting these for all runs.

docs/tutorial/expressjs.rst
```diff
@@ -0,0 +1,430 @@
...
+    :end-before: [docs:stop-docker-end]
+    :dedent: 2
+
+Chisel the rock
+===============
+
+This is an optional but recommended step, especially if we're looking to
+deploy the rock into a production environment. With :ref:`chisel_explanation`
+we can produce lean and production-ready rocks by getting rid of all the
+contents that are not needed for the ExpressJS application to run. This results
```
medubelko, line 236:
```suggestion
we can produce lean and production-ready rocks by chipping away all the
contents that the application doesn't need to run. This will result
```

docs/tutorial/expressjs.rst
```diff
@@ -0,0 +1,430 @@
...
+in a much smaller rock with a reduced attack surface.
+
+.. note::
+    It is recommended to run chiselled images in production. For development,
+    we may prefer non-chiselled images as they will include additional
+    development tooling (such as for debugging).
```
medubelko, line 242:
```suggestion
```

docs/tutorial/expressjs.rst
```diff
@@ -0,0 +1,430 @@
...
+.. literalinclude:: code/expressjs/task.yaml
+    :language: bash
+    :start-after: [docs:change-base]
+    :end-before: [docs:change-base-end]
+    :dedent: 2
+
+.. note::
+    The ``sed`` command replaces the current ``base`` in ``rockcraft.yaml`` with
+    the ``bare`` base. Note that ``build-base`` is also required when using the
+    ``bare`` base.
```
medubelko, line 258:
```suggestion
This sed command sets the current ``base`` key in ``rockcraft.yaml`` with
``bare``. When using the bare base, we also need the ``build-base`` key.
```

docs/tutorial/expressjs.rst
```diff
@@ -0,0 +1,430 @@
...
+    on an ``amd64`` platform, the name of the ``.rock`` file will be
+    different.
+
+    The size of the rock may vary depending on factors like the architecture
+    we are building on and the packages installed at the time of packing.
+
+Run the ExpressJS rock with Docker
+==================================
+
+We already have the rock as an `OCI <OCI_image_spec_>`_ archive. Now we
```
medubelko, line 141: We don't need to link to it twice:

```suggestion
We already have the rock as an OCI archive. Now we
```

docs/tutorial/expressjs.rst
```diff
@@ -0,0 +1,430 @@
...
+    :dedent: 2
+
+.. note::
+    The ``sed`` command replaces the current ``base`` in ``rockcraft.yaml`` with
+    the ``bare`` base. Note that ``build-base`` is also required when using the
+    ``bare`` base.
+
+So that we can compare the size after chiselling, open the ``rockcraft.yaml``
+file and change the ``version`` (e.g. to ``0.1-chiselled``). Pack the rock with
+the new ``bare`` :ref:`base <bases_explanation>`:
```
medubelko, line 262:
```suggestion
the new bare base:
```

docs/tutorial/expressjs.rst
```diff
@@ -0,0 +1,430 @@
...
+
+.. literalinclude:: code/expressjs/task.yaml
+    :language: bash
+    :start-after: [docs:change-base]
+    :end-before: [docs:change-base-end]
+    :dedent: 2
+
+.. note::
+    The ``sed`` command replaces the current ``base`` in ``rockcraft.yaml`` with
+    the ``bare`` base. Note that ``build-base`` is also required when using the
```
medubelko, line 257: I don't see a step where we instruct the user to add `build-base`. Should we include one?

docs/tutorial/expressjs.rst
```diff
@@ -0,0 +1,430 @@
...
+page.
+
+The application looks good, so let's stop it for now by pressing :kbd:`Ctrl` +
+:kbd:`C`.
+
+Pack the ExpressJS application into a rock
+==========================================
+
+First, we'll need a ``rockcraft.yaml`` file. Rockcraft will automate its
+creation and tailoring for a ExpressJS application by using the
```
medubelko, line 70:
```suggestion
First, we'll need a ``rockcraft.yaml`` project file. Rockcraft will automate
its creation and tailoring for a ExpressJS application by using the
```

docs/tutorial/expressjs.rst
```diff
@@ -0,0 +1,430 @@
...
+``expressjs-framework`` profile:
+
+.. literalinclude:: code/expressjs/task.yaml
+    :language: bash
+    :start-after: [docs:create-rockcraft-yaml]
+    :end-before: [docs:create-rockcraft-yaml-end]
+    :dedent: 2
+
+The ``rockcraft.yaml`` file will automatically be created in the project's
```
medubelko, line 79:
```suggestion
The project file will automatically be created in the
```

docs/tutorial/expressjs.rst
```diff
@@ -0,0 +1,430 @@
...
+Place the code snippet below in ``app.js`` under routes registration section.
+It will register the new ``/time`` endpoint:
+
+.. literalinclude:: code/expressjs/time_app.js
+    :language: javascript
+    :start-after: [docs:append-lines]
+    :end-before: [docs:append-lines-end]
+
+Since we are creating a new version of the application, open the
+``rockcraft.yaml`` file and change the ``version`` (e.g. to ``0.2``).
```
medubelko, line 336:
```suggestion
Since we are creating a new version of the application, set
``version: '0.2'`` in the project file.
```

docs/tutorial/expressjs.rst
```diff
@@ -0,0 +1,430 @@
...
+As before, verify that the new rock was created:
+
+.. literalinclude:: code/expressjs/task.yaml
+    :language: bash
+    :start-after: [docs:ls-bare-rock]
+    :end-before: [docs:ls-bare-rock-end]
+    :dedent: 2
+
+We'll verify that the new ExpressJS rock is now approximately **15% smaller**
+in size! And that's just because of the simple change of ``base``.
```
medubelko, line 279: We must be cautious with superlative statements.

```suggestion
The updated ExpressJS rock is a respectable 15% smaller
in size. We achieved that from a simple change of ``base``.
```

docs/tutorial/expressjs.rst
```diff
@@ -0,0 +1,430 @@
...
+
+And the functionality is still the same. As before, we can confirm this by
```
medubelko, line 281:
```suggestion
The functionality of the new rock is the same. Let's confirm this by
```

docs/tutorial/expressjs.rst
```diff
@@ -0,0 +1,430 @@
...
+running the rock with Docker
+
+.. literalinclude:: code/expressjs/task.yaml
+    :language: text
+    :start-after: [docs:docker-run-chisel]
+    :end-before: [docs:docker-run-chisel-end]
+    :dedent: 2
+
+and then using the same ``curl`` request:
```
medubelko, line 290:
```suggestion
running the rock with Docker:

.. literalinclude:: code/expressjs/task.yaml
    :language: text
    :start-after: [docs:docker-run-chisel]
    :end-before: [docs:docker-run-chisel-end]
    :dedent: 2

And using the same curl request:
```

medubelko: It would likely be simpler to combine these steps.

docs/tutorial/expressjs.rst
```diff
@@ -0,0 +1,430 @@
...
+    :end-before: [docs:curl-expressjs-bare-rock-end]
+    :dedent: 2
+
+The ExpressJS application should still respond with
+``Welcome to Express`` HTML.
+
+Cleanup
+~~~~~~~
+
+And that's it. We can now stop the container and remove the corresponding
```
medubelko, line 304:
```suggestion
We can now stop the container and remove the corresponding
```

docs/tutorial/expressjs.rst
```diff
@@ -0,0 +1,430 @@
...
+    :end-before: [docs:stop-docker-chisel-end]
+    :dedent: 2
+
+.. _update-expressjs-application:
+
+Update the ExpressJS application
+================================
+
+As a final step, let's update our application. For example,
+we want to add a new ``/time`` endpoint which returns the current time.
```
medubelko, line 319:
```suggestion
For our final task, let's update our application. As an example,
let's add a new ``/time`` endpoint that returns the current time.
```

docs/tutorial/expressjs.rst
```diff
@@ -0,0 +1,430 @@
...
+
+Start by creating the ``app/routes/time.js`` file in a text editor and paste the
+code from the snippet below:
+
+.. literalinclude:: code/expressjs/time.js
+    :language: javascript
```
medubelko, line 325:
```suggestion
    :caption: time.js
    :language: javascript
```

docs/tutorial/expressjs.rst
```diff
@@ -0,0 +1,430 @@
...
+
+Place the code snippet below in ``app.js`` under routes registration section.
+It will register the new ``/time`` endpoint:
+
+.. literalinclude:: code/expressjs/time_app.js
+    :language: javascript
```
medubelko, line 331:
```suggestion
    :caption: app.js
    :language: javascript
```

docs/tutorial/expressjs.rst
```diff
@@ -0,0 +1,430 @@
...
+.. literalinclude:: code/expressjs/task.yaml
+    :language: text
+    :start-after: [docs:docker-run-update]
+    :end-before: [docs:docker-run-update-end]
+    :dedent: 2
+
+.. note::
+
+    Note that the resulting ``.rock`` file will now be named differently, as
+    its new version will be part of the filename.
```
medubelko, line 355:
```suggestion
The resulting ``.rock`` file will be named differently, as
its new version will be part of the filename.
```

docs/tutorial/expressjs.rst
```diff
@@ -0,0 +1,430 @@
...
+
+Run the ExpressJS application from within the ``app/`` directory using
+
+.. code-block:: bash
+
+    npm start
+
+to verify that it works.
+
+Test the ExpressJS application by using ``curl`` to send a request to the root
```
medubelko, line 50:
```suggestion
Test the ExpressJS application by using curl to send a request to the root
```

docs/tutorial/expressjs.rst
```diff
@@ -0,0 +1,430 @@
...
+Now we're finally ready to run the rock and test the containerised ExpressJS
+application:
+
+.. literalinclude:: code/expressjs/task.yaml
+    :language: text
+    :start-after: [docs:docker-run]
+    :end-before: [docs:docker-run-end]
+    :dedent: 2
+
+Use the same ``curl`` command as before to send a request to the ExpressJS
```
medubelko, line 179:
```suggestion
Use the same curl command as before to send a request to the ExpressJS
```

docs/tutorial/expressjs.rst
```diff
@@ -0,0 +1,430 @@
...
+    :start-after: [docs:docker-run-update]
+    :end-before: [docs:docker-run-update-end]
+    :dedent: 2
+
+.. note::
+
+    Note that the resulting ``.rock`` file will now be named differently, as
+    its new version will be part of the filename.
+
+Finally, use ``curl`` to send a request to the ``/time`` endpoint:
```
medubelko, line 357:
```suggestion
Finally, use curl to send a request to the ``/time`` endpoint:
```

docs/tutorial/expressjs.rst
```diff
@@ -0,0 +1,430 @@
...
+.. literalinclude:: code/expressjs/task.yaml
+    :language: text
+    :start-after: [docs:curl-time]
+    :end-before: [docs:curl-time-end]
+    :dedent: 2
+
+The updated application should respond with the current date and time (e.g.
+``Fri Jan 10 2025 03:11:44 GMT+0000 (Coordinated Universal Time)``).
+
+.. note::
```
medubelko, line 368:
```suggestion
.. tip::
```

docs/tutorial/expressjs.rst
```diff
@@ -0,0 +1,430 @@
...
+
+    If you are getting a ``404`` for the ``/time`` endpoint, check the
+    :ref:`troubleshooting-expressjs` steps below.
+
+Cleanup
+~~~~~~~
```
medubelko, line 374:
```suggestion
Final cleanup
~~~~~~~~~~~~~
```

medubelko (changes requested):
@yanksyoon This is quite a hefty branch, so I apologise for the number of suggestions. Overall it fits the pattern of the other tutorials, but I think the new pages need some refinement.

docs/explanation/extensions/expressjs-framework.rst
```diff
@@ -0,0 +1,95 @@
...
+
+expressjs-framework
+===================
+
+When using the ``expressjs-framework`` extension, there are four different cases
+for customising the Ubuntu base and the Node version to be included.
+The main differences are:
+
+* whether the bare base is used or the Ubuntu 24.04 base is used.
+* whether Node is installed from Ubuntu packages or the NPM plugin.
```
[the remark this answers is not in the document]
yanksyoon: Generally the convention is to use `*` over `-` . fixed the other capitalisation/grammar issues, thank you!

docs/explanation/extensions/expressjs-framework.rst
```diff
@@ -0,0 +1,95 @@
...
+Ubuntu base and Node combinations
+---------------------------------
+
+In this section of the document, we will discuss the possible combinations of
+Ubuntu bases and possible usages of NPM plugin.
+
+24.04 base and Node from Ubuntu packages
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The following example uses the Ubuntu 24.04 base and Node from Ubuntu packages.
```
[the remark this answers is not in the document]
yanksyoon: This sentence has been added as a recommendation to put a single line of descriptor of the section before diving into the example code block. Would you think this would not be necessary?

docs/explanation/extensions/expressjs-framework.rst
```diff
@@ -0,0 +1,95 @@
...
+    extensions:
+    - expressjs-framework
+
+In this case, the application is installed and run via Node and NPM is installed
+by the Ubuntu packages. The NPM and Node versions are determined by the versions
+of NPM and NodeJS shipped with the Ubuntu base. See the NodeJS version shipped
+with the corressponding Ubuntu base from the chilsel-slices repository. This
+`link to the slices repository <https://github.com/canonical/chisel-releases/\
+blob/ubuntu-24.04/slices/nodejs.yaml>`_ is an example of the NodeJS version
+shipped with the Ubuntu 24.04 base.
```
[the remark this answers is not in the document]
yanksyoon: Node JS does not install the app - NPM installs the dependencies. The NPM is provided by the Ubuntu packages - Ubuntu base refers to the Ubuntu OS bases - not the package repository.
Also may I know if you intended to mean `sed slice`? The link refers to the NodeJS slice :)

Also I'm not quite sure what you mean by `sed slice uses this combination` - the slice is included as the staged packages but not quite as a combination of what has been described.

Thank you!

docs/explanation/extensions/expressjs-framework.rst
```diff
@@ -0,0 +1,95 @@
...
+    base: ubuntu@24.04
+    extensions:
+    - expressjs-framework
+    parts:
+        expressjs-framework/install-app:
+            npm-include-node: true
+            npm-node-version: 20.12
+
+In this case, the application is installed and run via Node, and NPM is
+installed by the NPM plugin.
```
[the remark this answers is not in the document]
yanksyoon: Same comment as [previous](https://github.com/canonical/rockcraft/pull/784#discussion_r1970905024) :D

docs/explanation/extensions/expressjs-framework.rst
```diff
@@ -0,0 +1,95 @@
...
+
+Bare base and Node from Ubuntu packages
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The following example uses the bare base and Node from Ubuntu packages.
```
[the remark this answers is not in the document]
yanksyoon: Updated to `Bare base, and Node.js from Ubuntu packages` :)

docs/explanation/extensions/expressjs-framework.rst
```diff
@@ -0,0 +1,95 @@
...
+    base: ubuntu@24.04
+    extensions:
+    - expressjs-framework
+
+In this case, the application is installed and run via Node and NPM is installed
+by the Ubuntu packages. The NPM and Node versions are determined by the versions
+of NPM and NodeJS shipped with the Ubuntu base. See the NodeJS version shipped
+with the corressponding Ubuntu base from the chilsel-slices repository. This
+`link to the slices repository <https://github.com/canonical/chisel-releases/\
+blob/ubuntu-24.04/slices/nodejs.yaml>`_ is an example of the NodeJS version
```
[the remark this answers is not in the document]
yanksyoon: The branch this slice refers to is in the `ubuntu-24.04` branch which contains `libnode109_libs`. If we try another base - let's say 22.04 base (https://github.com/canonical/chisel-releases/blob/ubuntu-22.04/slices/nodejs.yaml), such package does not exist.
Such dependency should be resolved by the user if they desire to override it or wish to find more about it.
This section explains that.

docs/explanation/extensions/expressjs-framework.rst
```diff
@@ -0,0 +1,95 @@
...
+    build-base: ubuntu@24.04
+    parts:
+        expressjs-framework/install-app:
+            npm-include-node: true
+            npm-node-version: 20.12
+
+In this case, the application is installed and run via Node and NPM is installed
+by the NPM plugin. For different possible inputs for npm-node-version, refer to
+the `NPM plugin documentation <https://documentation.ubuntu.com/rockcraft/en/\
+latest/common/craft-parts/reference/plugins/npm_plugin>`_.
```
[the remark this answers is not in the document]
yanksyoon: Node.JS is a runtime for running JavaScript code - the installation is done by the NPM plugin during build time.

docs/reference/extensions/expressjs-framework.rst
```diff
@@ -0,0 +1,57 @@
...
+
+Project requirements
+====================
+
+There are 3 requirements to be able to use the ``expressjs-framework``
+extension:
+
+1. The application should reside in the ``app`` directory.
+2. The application should have a ``package.json`` file.
+3. The ``package.json`` file should define the ``start`` script.
```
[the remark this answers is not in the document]
yanksyoon: The `package.json` file with the `start` script comes with the ExpressJS' official `express-generator` tool and is a common script in the NodeJS community to start the server.

docs/reference/extensions/expressjs-framework.rst
```diff
@@ -0,0 +1,57 @@
...
+package repository.
+
+``parts`` > ``expressjs-framework/runtime:`` > ``stage-packages``
+=================================================================
+
+Installing additional runtime packages is currently unsupported due to some
+limitations of the Rockcraft tool.
+
+Useful links
+============
```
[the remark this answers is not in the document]
yanksyoon: We would be diverging from the patterns used in other frameworks with this change - would this be intended?

docs/tutorial/expressjs.rst
```diff
@@ -0,0 +1,430 @@
...
+``multipass shell rock-dev`` to get another terminal:
+
+.. literalinclude:: code/expressjs/task.yaml
+    :language: bash
+    :start-after: [docs:curl-expressjs]
+    :end-before: [docs:curl-expressjs-end]
+    :dedent: 2
+
+The ExpressJS application should respond with a ``Welcome to Express`` HTML web
+page.
```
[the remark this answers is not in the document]
yanksyoon: We would be diverging from the patterns used in other frameworks with this change as well (e.g. django.rst, fastapi.rst,...etc). Would this be intended?

docs/tutorial/expressjs.rst
```diff
@@ -0,0 +1,430 @@
...
+application's root endpoint which is running inside the container:
+
+.. literalinclude:: code/expressjs/task.yaml
+    :language: text
+    :start-after: [docs:curl-expressjs-rock]
+    :end-before: [docs:curl-expressjs-rock-end]
+    :dedent: 2
+
+The ExpressJS application should again respond with the ``Welcome to Express``
+HTML.
```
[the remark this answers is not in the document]
yanksyoon: We're interacting with ExpressJS service via `curl` CLI interface - we wouldn't necessarily be seeing the page but rather the HTML response. Would this be still viable?

docs/tutorial/expressjs.rst
```diff
@@ -0,0 +1,430 @@
...
+and then using the same ``curl`` request:
+
+.. literalinclude:: code/expressjs/task.yaml
+    :language: text
+    :start-after: [docs:curl-expressjs-bare-rock]
+    :end-before: [docs:curl-expressjs-bare-rock-end]
+    :dedent: 2
+
+The ExpressJS application should still respond with
+``Welcome to Express`` HTML.
```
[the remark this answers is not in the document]
yanksyoon: Same comment as before, we're interacting via CLI and we would be seeing the HTML rather than the page.

docs/tutorial/expressjs.rst
yanksyoon: May I know the justification for removing this part please?

docs/tutorial/expressjs.rst
yanksyoon: This is in the `code/expressjs/task.yaml` code snippet referred to by the 
```
:start-after: [docs:change-base]
:end-before: [docs:change-base-end]
```
right above.

docs/tutorial/expressjs.rst
yanksyoon: Just out of curiosity, may I know when to use wrap things inside `` `` and when not to please? :D

docs/tutorial/expressjs.rst
yanksyoon: This also isn't aligned with other documentation for the extensions, shall I change all the others to match the styles?

docs/tutorial/expressjs.rst
yanksyoon: This also isn't aligned with other documentation for the extensions, shall I change all the others to match the styles?

erinecon: We can make the change to the other tutorials in a follow-up PR :)

docs/tutorial/expressjs.rst
```diff
@@ -0,0 +1,430 @@
...
+-----------------------------------------
+
+In this tutorial, we'll create a simple ExpressJS application and learn how to
+containerise it in a rock with Rockcraft's
+:ref:`expressjs-framework <expressjs-framework-reference>` extension.
+
+Setup
+=====
+
+.. include:: /reuse/tutorial/setup.rst
```
jdkandersson, line 13: For the Docker installation, after the newgroup command a new terminal is spawned. So if someone copies the entire block and executes it at once, the following commands won't run

docs/tutorial/expressjs.rst
```diff
@@ -0,0 +1,430 @@
...
+the first part of this tutorial, so we'll stop the container and remove the
+respective image for now:
+
+.. literalinclude:: code/expressjs/task.yaml
+    :language: bash
+    :start-after: [docs:stop-docker]
+    :end-before: [docs:stop-docker-end]
+    :dedent: 2
+
+Chisel the rock
```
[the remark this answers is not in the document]
yanksyoon: Then I will redo the entire tutorial to match the style of bare base extensions, i.e. go, which will take extended time to finish. Perhaps I revert this PR as draft again until that is done. I can reopen this next week.

docs/tutorial/expressjs.rst
medubelko: Great question! The overall goal is to reduce the amount of extra decoration, like inline code, in text.

We're planning on adding this convention to the style guide:

- When guiding a specific _usage_ of a command, or discussing the command itself, use inline code.
- When discussing or simply _identifying_ a product/project by name, use its product name.

It's about product recognition and identity. curl is a project/product, the same as Charmcraft, Ubuntu, Node.js. It might have originally started as a command, but it's since become a standalone app that users consume, update, and think about on its own merits.

Consider these paragraphs:

- "I'm having a hard time understanding the build process in Charmcraft. Can you help?"
- "When I run `charmcraft pack`, it throws an error. Can you help?"

docs/explanation/extensions/expressjs-framework.rst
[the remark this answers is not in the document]
medubelko: > Generally the convention is to use * over - .

Our convention is to use hyphens: https://canonical-documentation-with-sphinx-and-readthedocscom.readthedocs-hosted.com/style-guide/#lists

It's a minor point, but they are more readable.

docs/explanation/extensions/expressjs-framework.rst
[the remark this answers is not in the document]
medubelko: The original text says Node.js performs the installation!

> the application is installed and run via Node

I believe this is an ambiguous scope problem in the grammar.

docs/reference/extensions/expressjs-framework.rst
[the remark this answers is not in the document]
medubelko: That's fine – those tutorials are undergoing continuous updates.

docs/tutorial/expressjs.rst
medubelko: No need – the rest of the team can update them later.

docs/tutorial/expressjs.rst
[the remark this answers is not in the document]
medubelko: I see what you're saying.

curl returns any web resource in the terminal as text. A URI that serves a `content="text/html"` resource is synonymous with the idea of "web page". With that understanding, I think it's fine to still call it a page in this context.

To me, saying it returns "\<page title\> HTML" feels like we'll lead the user to do something HTML-related.

As a real-world corollary, one would say "I read _A Brief History of Time_ as an e-book", not "I read _A Brief History of Time_ as an EPUB file".

docs/tutorial/expressjs.rst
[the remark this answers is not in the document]
medubelko: Correct. We can clean up the others later. Apologies if that feels like an out-of-scope change, but I've recently had more time to dedicate to these tutorials than before.

docs/tutorial/expressjs.rst
medubelko: You're right – the order of the paragraphs mixed me up.

docs/tutorial/expressjs.rst
medubelko: Tutorials, according to Diataxis, walk the user down a single path. It's very strict about this, directing us to avoid any unnecessary explanation and forbidding us from branching or offering alternative paths.

My perspective on notes is that they're optional information. And since this discusses a completely different concern, debugging, I found it was a diversion.

docs/explanation/extensions/expressjs-framework.rst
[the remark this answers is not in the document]
medubelko: Yanks and I discussed this. The confusion comes from the phrase "Ubuntu packages" to mean https://packages.ubuntu.com. I'm checking to see if there's a less ambiguous name we can use.

docs/explanation/extensions/expressjs-framework.rst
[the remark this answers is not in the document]
yanksyoon: Moved to `how-to/extensions/configure-the-build-base-for-an-expressjs-app.rst` :)

docs/explanation/extensions/expressjs-framework.rst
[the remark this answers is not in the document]
yanksyoon: Removed all initial sentences :)

docs/explanation/extensions/expressjs-framework.rst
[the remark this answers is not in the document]
yanksyoon: Updated "ubuntu packages" to "Ubuntu package archive" :) thank you

docs/explanation/extensions/expressjs-framework.rst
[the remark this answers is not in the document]
yanksyoon: Updated to
```
In this case, the application is installed and run via NodeJS provided by the
NPM plugin. The NPM is also installed by the NPM plugin.
```
to clarify where the sources of installations are from :)

docs/explanation/extensions/expressjs-framework.rst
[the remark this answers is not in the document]
yanksyoon: Updated the previous to:
```
In this case, the application is installed and run via NodeJS provided by the
NPM plugin. The NPM is also installed by the NPM plugin.
```

Updated the current example to:
```
In this case, the application is installed and run via NodeJS and NPM from the
NPM plugin.
```
Hope this clears the ambiguity regarding the sources of installation!

docs/tutorial/expressjs.rst
yanksyoon: Bare base tutorial has been removed in favor of matching the style in `go` tutorial.

yanksyoon: (the default starts with bare base)

docs/tutorial/expressjs.rst
yanksyoon: Thank you for the explanation!

docs/tutorial/expressjs.rst
```diff
@@ -0,0 +1,430 @@
...
+
+Run the ExpressJS application from within the ``app/`` directory using
+
+.. code-block:: bash
+
+    npm start
+
+to verify that it works.
+
+Test the ExpressJS application by using curl to send a request to the root
```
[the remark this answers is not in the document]
yanksyoon: Added the note to render in browser, thank you :)

docs/tutorial/expressjs.rst
```diff
@@ -0,0 +1,430 @@
...
+
+.. note::
+    For this tutorial, we'll use the ``name`` ``expressjs-hello-world`` and
+    assume we're running on the  ``amd64`` platform. Check the architecture of
+    the system using ``dpkg --print-architecture``.
+
+    The ``name``, ``version`` and ``platform`` all influence the name of the
+    generated ``.rock`` file.
+
+Pack the rock:
```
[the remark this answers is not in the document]
yanksyoon: This is an issue outside of the PR, will raise an issue against the repo :)

docs/tutorial/expressjs.rst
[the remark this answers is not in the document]
yanksyoon: The tutorial is updated to use bare base as default and hence excludes the chiselling step, following the Go tutorial.

docs/tutorial/expressjs.rst
```diff
@@ -0,0 +1,430 @@
...
+let's add a new ``/time`` endpoint that returns the current time.
+
+Start by creating the ``app/routes/time.js`` file in a text editor and paste the
+code from the snippet below:
+
+.. literalinclude:: code/expressjs/time.js
+    :caption: time.js
+    :language: javascript
+
+Place the code snippet below in ``app.js`` under routes registration section.
```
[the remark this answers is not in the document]
yanksyoon: Added `app/` prefix (`app/app.js`)

docs/tutorial/expressjs.rst
```diff
@@ -0,0 +1,430 @@
...
+It will register the new ``/time`` endpoint:
```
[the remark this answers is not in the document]
yanksyoon: Updated, thank you :)

docs/tutorial/expressjs.rst
yanksyoon: That seems like it should be fixed as a separate PR, would you like to create an issue for it against this repository please?

docs/tutorial/expressjs.rst
```diff
@@ -0,0 +1,430 @@
...
+    :end-before: [docs:curl-expressjs-end]
+    :dedent: 2
+
+The ExpressJS application should respond with a ``Welcome to Express`` HTML web
+page.
+
+The application looks good, so let's stop it for now by pressing :kbd:`Ctrl` +
+:kbd:`C`.
+
+Pack the ExpressJS application into a rock
```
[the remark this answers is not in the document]
yanksyoon: Good point, i'll remove any `node_modules` copied over before installing the application.

docs/tutorial/expressjs.rst
yanksyoon: Updated this section of the tutorial, thank you :)

docs/tutorial/expressjs.rst
yanksyoon: Chiselling step has been removed in favor of the `bare` base default :)

docs/tutorial/expressjs.rst
yanksyoon: Updated, thank you :)

docs/tutorial/code/expressjs/task.yaml
```diff
@@ -137,10 +108,10 @@ execute: |
 
   # [docs:cleanup]
   # delete all the files created during the tutorial
+  npm uninstall -g express-generator@4
```
jdkandersson, line 111: maybe need sudo

jdkandersson (approved):

docs/reference/extensions/expressjs-framework.rst
```diff
@@ -0,0 +1,55 @@
...
+.. code-block:: yaml
+
+  parts:
+    expressjs-framework/install-app:
+      npm-include-node: true
+      npm-node-version: 20.12.2
+
+For more examples of the ``npm-node-version`` key, see the
+`NPM plugin <httpshttps://documentation.ubuntu.com/rockcraft/en/latest/common/
+craft-parts/reference/plugins/npm_plugin/#examples>`_ documentation.
```
medubelko, line 42:
```suggestion
For more examples of the ``npm-node-version`` key, see :ref:`craft_parts_npm_plugin`.
```

docs/explanation/extensions/expressjs-framework.rst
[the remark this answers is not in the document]
medubelko: I'm still confused, sorry. The updated text reads:

> the application is installed and run via NodeJS

Which is what I put in my suggestion:

> Node.js installs and runs the app

Those two sentences are equivalent.

If the npm plugin performs the installation, the text should read:

```
... the npm plugin installs the app and Node.js, while Node.js runs the app.
```

docs/how-to/extensions/configure-the-build-base-for-expressjs-app.rst
```diff
@@ -0,0 +1,86 @@
...
+
+expressjs-framework
+===================
+
+When using the ``expressjs-framework`` extension, there are four different cases
+for customising the Ubuntu base and the Node version to be included.
+The main differences between the cases are:
+
+- Whether to use the bare base or Ubuntu 24.04 base.
+- Whether Node is installed from Ubuntu package archive or the NPM plugin.
```
medubelko, line 11: "Node.js" for all instances on this page.

docs/tutorial/expressjs.rst
```diff
@@ -0,0 +1,334 @@
...
+Now we're finally ready to run the rock and test the containerised ExpressJS
+application:
+
+.. literalinclude:: code/expressjs/task.yaml
+    :language: text
+    :start-after: [docs:docker-run]
+    :end-before: [docs:docker-run-end]
+    :dedent: 2
+
+Use the same ``curl`` command as before to send a request to the ExpressJS
```
medubelko, line 164:
```suggestion
Use the same curl command as before to send a request to the ExpressJS
```

docs/tutorial/expressjs.rst
```diff
@@ -0,0 +1,334 @@
...
+    :language: bash
+    :start-after: [docs:ls-rock]
+    :end-before: [docs:ls-rock-end]
+    :dedent: 2
+
+
+Run the ExpressJS rock with Docker
+==================================
+
+We already have the rock as an OCI archive. Load the image into Docker:
```
medubelko, line 130: > OCI archive

We used "image" in the previous section. Should they both be "image"?

docs/tutorial/expressjs.rst
```diff
@@ -0,0 +1,334 @@
...
+
+Check that the image was successfully loaded into Docker:
+
+.. literalinclude:: code/expressjs/task.yaml
+    :language: bash
+    :start-after: [docs:docker-images]
+    :end-before: [docs:docker-images-end]
+    :dedent: 2
+
+The output should list the Go container image, along with its tag, ID and
```
medubelko, line 147:
```suggestion
The output should list the Go container, along with its tag, ID and
```

As far as I know, images become containers when run.

docs/tutorial/expressjs.rst
```diff
@@ -0,0 +1,334 @@
...
+    :start-after: [docs:append-lines]
+    :end-before: [docs:append-lines-end]
+
+Since we are creating a new version of the application, set
+``version: '0.2'`` in the project file.
+
+Pack and run the rock using similar commands as before:
+
+.. literalinclude:: code/expressjs/task.yaml
+    :language: text
```
medubelko, line 245:
```suggestion
    :language: bash
```

docs/tutorial/expressjs.rst
```diff
@@ -0,0 +1,334 @@
...
+    :end-before: [docs:docker-run-update-end]
+    :dedent: 2
+
+The resulting ``.rock`` file will be named differently, as
+its new version will be part of the filename.
+
+Finally, use curl to send a request to the ``/time`` endpoint:
+
+.. literalinclude:: code/expressjs/task.yaml
+    :language: text
```
medubelko, line 256:
```suggestion
    :language: bash
```

yanksyoon:
@medubelko 

```
I'm still confused, sorry. The updated text reads:

the application is installed and run via NodeJS

Which is what I put in my suggestion:

Node.js installs and runs the app

Those two sentences are equivalent.

If the npm plugin performs the installation, the text should read:

... the npm plugin installs the app and Node.js, while Node.js runs the app.
```

Thanks for the comment! To clarify, the application "installation" is done by the NPM plugin - which involves two things:
1. Copying over the source files (done by rockcraft NPM plugin)
2. Installing the libraries in the package.json (done by npm tool)

Then, NodeJS provides a runtime in which the application is loaded and served. Would this make a difference in how we phrase the above?

docs/how-to/extensions/configure-the-build-base-for-expressjs-app.rst
```diff
@@ -0,0 +1,86 @@
...
+        - expressjs-framework
+
+In this case, the application is installed and run via Node and NPM is installed
+from the Ubuntu package archive. The NPM and Node versions are determined by the
+versions of NPM and NodeJS shipped with the Ubuntu base. See the NodeJS version
+shipped with the corressponding Ubuntu base from the chilsel-slices repository.
+This `link to the slices repository
+<https://github.com/canonical/chisel-releases/blob/ubuntu-24.04/slices/
+nodejs.yaml>`_ is an example of the NodeJS version shipped with the Ubuntu 24.04
+base.
```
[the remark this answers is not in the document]
yanksyoon: I agree with your concerns, would removing the link be better in this case? I'm wondering where we can provide the users information about the chiseled slices that are being used here - or if it is necessary to guide them to slices at all.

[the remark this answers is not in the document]
medubelko: What about linking to the Ubuntu archive? https://packages.ubuntu.com/noble/npm

medubelko:
@yanksyoon 

> Would this make a difference in how we phrase the above?

Yes. :)

Let me try again:

> In this case, the NPM plugin copies the app's files and installs its dependencies. Node.js then provides a runtime and launches the app.
> 
> For different possible values for the `npm-node-version` key...

jdkandersson (approved):

medubelko (approved):
Thanks @yanksyoon!

docs/how-to/index.rst
```diff
@@ -39,6 +39,16 @@ compiled a couple of guides for you to facilitate your contribution.
...
 
+Extensions
+----------
+
+Learn how to use framework and platform extensions in your project.
+
+.. toctree::
+   :maxdepth: 1
+
+   Configure the build base for ExpressJS App <configure-the-build-base-for-expressjs-app>
```
tigarmo, line 50: This line is too long and has a typo ("an" ExpressJS App). Can we just use the target page's title?

tigarmo (approved):
Thanks! Approved pending the linting fix

docs/how-to/index.rst
```diff
@@ -40,6 +40,16 @@ compiled a couple of guides for you to facilitate your contribution.
...
```
tigarmo, line 51:
```suggestion
   configure-the-build-base-for-expressjs-app
```

tigarmo:
I think the breakage came in from the update in `main`:
```
/home/docs/checkouts/readthedocs.org/user_builds/canonical-rockcraft/checkouts/784/docs/tutorial/expressjs.rst:24: CRITICAL: Problems with "include" directive path:
InputError: [Errno 2] No such file or directory: 'reuse/tutorial/setup.rst'.
```

jdkandersson (approved):

PR #792, opened 2025-01-21
feat: add uv plugin

### LINKED ISSUE #789: Add support for the uv plugin
### What needs to get done

craft-parts recently added a uv plugin, which should be adopted into Rockcraft.

### Why it needs to get done

uv is a popular build system and Rockcraft would benefit from supporting it

bepri:

-----
Adding the uv plugin from craft-parts to Rockcraft.

CRAFT-3916
Closes #789.


rockcraft/plugins/uv_plugin.py
```diff
@@ -0,0 +1,69 @@
...
+        """Overridden because for ubuntu bases we must always remove the symlinks."""
+        return python_common.should_remove_symlinks(self._part_info)
+
+    @override
+    def _get_system_python_interpreter(self) -> str | None:
+        """Overridden because Python must always be provided by the parts.
+
+        The uv plugin requires a name to reference Python, so we must depend
+        on a relative python3 being installed."""
+        return "python3"
```
tigarmo, line 41: Can you add the craft-parts bug# here?

rockcraft/plugins/uv_plugin.py
```diff
@@ -0,0 +1,69 @@
...
+
+    @override
+    def _get_script_interpreter(self) -> str:
+        """Overridden because Python is always available in /bin."""
+        return python_common.get_script_interpreter()
+
+    @override
+    def _get_rewrite_shebangs_commands(self) -> list[str]:
+        """Overridden because the original uv plugin does not rewrite shebangs."""
```
tigarmo, line 50: I don't understand why this method is different from the other python-based plugins?

tests/spread/rockcraft/plugin-uv/rockcraft.yaml
```diff
@@ -0,0 +1,16 @@
...
+  amd64:
+
+parts:
+  my-part:
+    plugin: uv
+    source: src/
+    build-snaps:
+      - astral-uv
+    stage-packages:
+      - python3
```
tigarmo, line 16:
```suggestion
      - python3-minimal
```

tigarmo (changes requested):
The docs failure is related:

```

checking consistency... /home/docs/checkouts/readthedocs.org/user_builds/canonical-rockcraft/checkouts/792/docs/common/craft-parts/reference/plugins/go_use_plugin.rst: WARNING: document isn't included in any toctree
/home/docs/checkouts/readthedocs.org/user_builds/canonical-rockcraft/checkouts/792/docs/common/craft-parts/reference/plugins/uv_plugin.rst: WARNING: document isn't included in any toctree
```

rockcraft/plugins/uv_plugin.py
bepri: To my understanding, the uv plugin didn't need to rewrite shebangs since it creates a relocatable venv, but Rockcraft depends on shebangs for different reasons so I had to add this override here

lengau: The uv plugin uses uv's relocatable venvs, which means the shebang is actually `#!/bin/sh`. Specifically it sets this as the first few lines of the Python executable:

```bash
#!/bin/sh
'''exec' "$(dirname -- "$(realpath -- "$0")")"/'python3' "$0" "$@"
' '''
```

If a rock can guarantee those three files (`/bin/sh`, `/usr/bin/realpath` and `/usr/bin/dirname`) we don't need to rewrite shebangs.

tigarmo: I'd rather the shebang the rewritten to `#!/bin/python3`. Is that what this is doing?

bepri: That is what this override should do, yes

tigarmo:
@lengau do you know what's up with this OSV-scanner? I don't know which requirements file it's picking up here:
image

bepri:
I think the failure is coming from this being unpinned: https://github.com/canonical/rockcraft/blob/main/docs/tutorial/code/fastapi/requirements.txt

tigarmo (approved):
great!

tests/spread/rockcraft/plugin-uv/rockcraft.yaml
bepri: Needed to stay as python3:
```
2025-01-28T20:00:22.7921246Z + docker run --rm plugin-uv:0.1 exec black --version
2025-01-28T20:00:22.7922319Z Traceback (most recent call last):
2025-01-28T20:00:22.7923163Z   File "/usr/bin/black", line 6, in <module>
2025-01-28T20:00:22.7924248Z     from black import patched_main
2025-01-28T20:00:22.7925117Z   File "src/black/__init__.py", line 2, in <module>
2025-01-28T20:00:22.7925817Z ModuleNotFoundError: No module named 'json'
```

tests/spread/rockcraft/plugin-uv/src/test_uv/__main__.py
```diff
@@ -0,0 +1,6 @@
+def main():
```
lengau, line 1:
```suggestion
import black  # Test the importing of a non-stdlib module

# "Use" it without using it so linters don't complain
_ = black

def main():
```

lengau (approved):

medubelko (approved):

PR #799, opened 2025-01-30
Update craft parts, minor build and ci fixes

tigarmo:

---

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,30 +1,28 @@
 # .readthedocs.yaml
 # Read the Docs configuration file
 # See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
----
+
 # Required
 version: 2
-
-build:
-  os: ubuntu-22.04
-  tools:
-    python: "3.10"
-  jobs:
-    post_checkout:
-      - git fetch --tags --unshallow # Also fetch tags
-      - git describe # Make sure we get a proper version
 
 # Build documentation in the docs/ directory with Sphinx
 sphinx:
   configuration: docs/conf.py
-  builder: dirhtml
   fail_on_warning: true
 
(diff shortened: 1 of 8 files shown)


uv.lock
```diff
@@ -591,7 +591,7 @@ wheels = [
 
 [[package]]
 name = "craft-parts"
-version = "2.4.1"
+version = "2.5.0"
```
tigarmo, line 594: done with `uv lock --upgrade-package craft-parts`
I've always been a big uv fan, it is known

.readthedocs.yaml
```diff
@@ -24,7 +31,4 @@ sphinx:
 # and requirements required to build your docs
 python:
   install:
-    - method: pip
-      path: .
-      extra_requirements:
-        - docs
+    - requirements: uv-requirements.txt
```
dariuszd21, line 34: It's done a bit differently than in https://github.com/canonical/starbase/blob/main/.readthedocs.yaml 
Any reasons why it is done this way or is there a reason to upstream this way ?

uv.lock
```diff
@@ -2306,6 +2306,7 @@ wheels = [
 
 [[package]]
 name = "rockcraft"
+version = "1.7.0.post39+g6f1048a.d20250130"
```
dariuszd21, line 2309:
```suggestion
```

tigarmo: I don't know how this got here, I'll remove it

.readthedocs.yaml
tigarmo: the reason is: I forgot to see if starbase had already solved this issue, heh

tigarmo: I'll update this here

bepri (approved):

PR #806, opened 2025-02-06
feat(docs): add feedback button

erinecon:

---

Add a "Give feedback" button to the top of the RTD site that links the user to a GitHub issue form to describe any feedback they have about a particular docs page.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -193,10 +193,8 @@ craft_parts_docs_path = pathlib.Path(craft_parts_docs.__file__).parent / "craft-
     craft_parts_docs_path, target_is_directory=True
 )
 
-# By default, the documentation includes a feedback button at the top.
-# You can disable it by setting the following configuration to True.
-disable_feedback_button = False
-
-html_js_files = ['header-nav.js']
-if 'github_issues' in html_context and html_context['github_issues'] and not disable_feedback_button:
-    html_js_files.append('github_issue_links.js')
+# Workaround to enable extra scripts in page template
+html_js_files = [
+    'header-nav.js', # "More links" onclick events from canonical-sphinx-extensions
+    'github_issue_links.js', # Feedback button from canonical-sphinx-extensions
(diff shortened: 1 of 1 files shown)


medubelko (approved):
LGTM! Thanks, @erinecon.

Longer-term we'll replace the composed text in the query with a simple link to a GitHub issue template (CRAFT-4123).

docs/_static/github_issue_links.js
tigarmo, line 1: It seems like this js code should go on [canonical-sphinx](https://github.com/canonical/canonical-sphinx), no?

actually isn't it already there? https://github.com/canonical/canonical-sphinx/blob/main/canonical_sphinx/theme/static/github_issue_links.js

medubelko: It is, but this is a high-priority request. From what I understand, our docs base has diverged and I didn't want to block this task for want of a full reintegration. Do you think it would be trivial to bring it back in sync?

medubelko: Ignore previous – you're right, we only need to enable the button, not redo its code. Rebased with alteration.

docs/conf.py
```diff
@@ -191,3 +192,11 @@ def setup(app):
 (common_docs_path / "craft-parts").symlink_to(
     craft_parts_docs_path, target_is_directory=True
 )
+
+# By default, the documentation includes a feedback button at the top.
+# You can disable it by setting the following configuration to True.
+disable_feedback_button = False
```
tigarmo, line 198: Is this part necessary? I tried it locally with only defining `"github_issues"` above and the feedback button seems to show up and work fine

medubelko: This is boilerplate we've been sharing between repos. Refactored: [bab71e8](https://github.com/canonical/rockcraft/pull/806/commits/bab71e83ee32cddad779d110417e55b82cf5d10b)

PR #818, opened 2025-02-19
docs: fix extensions urls

st3v3nmw:

---

The current URLs raise 404s since they don't have the `.html` extension:. [/reference/extensions/go-framework](https://documentation.ubuntu.com/rockcraft/en/1.8.0/reference/extensions/go-framework) vs [/reference/extensions/go-framework.html](https://documentation.ubuntu.com/rockcraft/en/1.8.0/reference/extensions/go-framework.html)


sergiusens (changes requested):
we should instead fix the root of this, which might be the redirect rules using extensions

PR #822, opened 2025-02-27
build(snap): add sitecustomize to snap package

dariuszd21:

Add sitecustomize.py to `rockcraft` to be able to load plugins dynamically.
Add `craft-store` as an extra dependency.

(CRAFT-4022)


uv.lock
```diff
@@ -516,7 +517,7 @@ toml = [
 
 [[package]]
 name = "craft-application"
-version = "4.8.2"
+version = "4.9.1"
```
dariuszd21, line 520: requires `4.10.0` release

pyproject.toml
```diff
@@ -25,6 +25,7 @@ dependencies = [
     "craft-parts",
     "craft-platforms",
     "craft-providers",
+    "craft-store",
```
lengau, line 28: This one's an interesting one since rockcraft technically doesn't need `craft-store` unless a plugin uses it. Should we put it in an extra that we include in the snap?

lengau (approved):

pyproject.toml
dariuszd21: I think that makes sense to put it in separate extra group

tigarmo (approved):

pyproject.toml
dariuszd21: Done in https://github.com/canonical/rockcraft/pull/822/commits/11c726f0bb85f2fa30962f11fd2095e6c8c4f55d

PR #825, opened 2025-02-28
build(snap): disable armhf

tigarmo:
The `astral-uv` snap on `armhf` needs to be updated/fixed. Until then, we need to stop using it.

---

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -16,7 +16,8 @@ environment:
 platforms:
   amd64:
   arm64:
-  armhf:
+  # Disable armhf until the astral-uv snap is fixed.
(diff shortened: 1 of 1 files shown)


snap/snapcraft.yaml
```diff
@@ -16,7 +16,8 @@ environment:
 platforms:
   amd64:
   arm64:
-  armhf:
+  # Disable armhf until the astral-uv snap is fixed.
```
mr-cal, line 19:
```suggestion
  # Disable armhf until the astral-uv snap is fixed (https://github.com/lengau/uv-snap/issues/4)
```

bepri (approved):

PR #827, opened 2025-03-03
Docs: Tutorial setup maintenance

erinecon:

---

Two maintenance-related changes to the tutorial setup file.

1. Split up the reusable tutorial setup file into two based on whether the tutorial requires the `latest/edge` channel of Rockcraft.

2. Fix a bug in the setup instructions pertaining to Docker. Currently, the Docker setup code block looks like this:

```
sudo addgroup --system docker
sudo adduser $USER docker
newgrp docker
sudo snap disable docker
sudo snap enable docker
```

However, the `newgrp` command will open a new shell session. The `disable` and `enable` commands won't run if the user copies and pastes the entire command block. The fix splits up the command block into two different blocks so copying/pasting the whole block will work. 

Additional reviewers: @jdkandersson


jdkandersson (approved):

docs/reuse/tutorial/setup_edge.rst
```diff
@@ -80,6 +80,15 @@ to be able to use Docker commands as a regular user:
     :end-before: [docs:docker-regular-user-end]
     :dedent: 2
 
+The ``newgrp`` command should open a new shell session. Use ``exit`` to
```
jdkandersson, line 83: if we exit the shell then the newgrp command doesn't take effect until people exit out of multipass and start again. Perhaps we just tell them this open a new shell and to then just keep going with the commands?

jdkandersson: this is definitely an area of linux that is not intuitive

erinecon: Ah okay, I don't know if I'm totally understanding, but here's a proposed change:

```
The ``newgrp`` command should open a new shell session.
Exit from that session using ``exit``, open a new shell using 
``multipass shell rock-dev`` and restart Docker:
```

erinecon: Sure, okay. I will simplify to just `Restart Docker:` and leave it alone. Thanks!

docs/reuse/tutorial/setup_edge.rst
```diff
@@ -80,8 +80,7 @@ to be able to use Docker commands as a regular user:
     :end-before: [docs:docker-regular-user-end]
     :dedent: 2
 
-The ``newgrp`` command should open a new shell session. Use ``exit`` to
-exit from the new shell, and then restart Docker:
+Restart Docker:
```
jdkandersson, line 83: Do we need this in stable as well?

jdkandersson: oh yeh my bad

jdkandersson (approved):

medubelko (approved):

tigarmo (approved):
thanks!

PR #837, opened 2025-03-19
docs: add security policy

tigarmo:
For now it is very close to Snapcraft's, even though Rockcraft itself only has a single major release (1.x). As this documents the current policy, it does not make any references to interim bases - we'll need to update this document when we "oficially" support those.

---


SECURITY.md
```diff
@@ -0,0 +1,54 @@
+# Security policy
+
+## Rockcraft and rocks
+
+Rockcraft is a tool that generates rocks, which are OCI container images. Rocks are
+typically composed of Ubuntu software and application-specific code, and as such they
+are susceptible to security issues and vulnerabilities. Authors should always maintain
```
bepri, line 7:
```suggestion
are susceptible to security issues and vulnerabilities. Authors should always keep
```

SECURITY.md
```diff
@@ -0,0 +1,54 @@
...
+their rocks updated, but a security vulnerability on a rock is not considered an issue
+with Rockcraft itself, and should be reported to the rock's maintainer.
+
+## Build isolation
```
bepri, line 11: This section seems like a good candidate for upstreaming. Thoughts @medubelko @mr-cal ?

bepri (changes requested):

SECURITY.md
medubelko: I agree with the idea but the wording will need to be massaged.

SECURITY.md
```diff
@@ -0,0 +1,54 @@
...
+
+In typical operation, Rockcraft makes use of tools like [LXD] and [Multipass] to create
+isolated build environments. Rockcraft itself provides no extra security, relying on
+the tools themselves to ensure proper sandboxing. Security issues identified on these
+build environments should be reported to the corresponding project.
```
medubelko, line 16:
```suggestion
these tools to provide secure sandboxing. The security of these build environments
are the responsibility of these tools and should be reported to their respective
project maintainers.
```

medubelko (changes requested):

tigarmo:
@bepri @medubelko I applied all of your suggestions (manually), thanks!

medubelko (approved):
:rock: :on: :tea: :a: ご

PR #838, opened 2025-03-19
build(deps): update dependencies

tigarmo:
Updated anyio v4.8.0 -> v4.9.0
Updated astroid v3.3.8 -> v3.3.9
Updated attrs v25.1.0 -> v25.3.0
Updated canonical-sphinx v0.3.0 -> v0.4.0
Updated canonical-sphinx-extensions v0.0.23 -> v0.0.27
Updated coverage v7.6.12 -> v7.7.0
Updated craft-grammar v2.0.1 -> v2.0.3
Updated craft-parts v2.6.2 -> v2.7.0
Updated cryptography v44.0.1 -> v44.0.2
Updated dulwich v0.22.7 -> v0.22.8
Updated filelock v3.17.0 -> v3.18.0
Updated findpython v0.6.2 -> v0.6.3
Updated hypothesis v6.127.2 -> v6.129.4
Updated iniconfig v2.0.0 -> v2.1.0
Updated jeepney v0.8.0 -> v0.9.0
Updated jinja2 v3.1.5 -> v3.1.6
Updated pbs-installer v2025.2.12 -> v2025.3.17
Updated platformdirs v4.3.6 -> v4.3.7
Updated protobuf v5.29.3 -> v6.30.1
Updated pyfakefs v5.7.4 -> v5.8.0
Updated pylint v3.3.4 -> v3.3.5
Updated pytest v8.3.4 -> v8.3.5
Updated pytest-check v2.5.0 -> v2.5.1
Updated python-apt v2.4.0+ubuntu4, v2.7.7+ubuntu3, v2.9.0+ubuntu1 -> v2.4.0+ubuntu4, v2.7.7+ubuntu3, v2.9.9
Updated pywin32 v308 -> v310
Updated rapidfuzz v3.12.1 -> v3.12.2
Updated responses v0.25.6 -> v0.25.7
Updated setuptools v75.8.2 -> v77.0.1
Updated starlette v0.46.0 -> v0.46.1
Updated trove-classifiers v2025.2.18.16 -> v2025.3.19.19
Updated types-setuptools v75.8.0.20250225 -> v76.0.0.20250313
Updated virtualenv v20.29.2 -> v20.29.3
Updated websockets v15.0 -> v15.0.1

### DIFF
--- a/rockcraft/extensions/_constants.py
+++ b/rockcraft/extensions/_constants.py
@@ -1,0 +1,27 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright 2025 Canonical Ltd.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Constants for extensions."""
+
+_PYTHON_3_10 = "3.10"
+_PYTHON_3_12 = "3.12"
+
+UBUNTU_PYTHON_VERSION_MAP = {
+    "ubuntu@22.04": _PYTHON_3_10,
+    "ubuntu:22.04": _PYTHON_3_10,
+    "ubuntu@24.04": _PYTHON_3_12,
+    "ubuntu:24.04": _PYTHON_3_12,
+}
--- a/rockcraft/extensions/_utils.py
+++ b/rockcraft/extensions/_utils.py
@@ -20,6 +20,7 @@ import copy
 from pathlib import Path
 from typing import Any, cast
 
+from ._constants import UBUNTU_PYTHON_VERSION_MAP
 from .extension import Extension
 from .registry import get_extension_class
 
@@ -116,3 +117,12 @@ def _remove_list_duplicates(seq: list[str]) -> list[str]:
             deduped.append(item)
 
     return deduped
+
+
+def find_ubuntu_base_python_version(base: str) -> str:
+    """Find the Python version for the given base."""
+    if base not in UBUNTU_PYTHON_VERSION_MAP:
+        raise NotImplementedError(
+            f"Python version for base {base} is not yet supported."
+        )
+    return UBUNTU_PYTHON_VERSION_MAP[base]
--- a/rockcraft/extensions/gunicorn.py
+++ b/rockcraft/extensions/gunicorn.py
@@ -28,6 +28,7 @@ from packaging.requirements import InvalidRequirement, Requirement
 
 from ..errors import ExtensionError
 from ._python_utils import has_global_variable
+from ._utils import find_ubuntu_base_python_version
 from .extension import Extension, get_extensions_data_dir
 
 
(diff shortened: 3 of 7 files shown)


docs/how-to/index.rst
```diff
@@ -16,6 +16,7 @@ adapt the steps to fit your specific requirements.
    Get started - quick guide <get-started>
    Build a 12-factor app rock <build-a-12-factor-app-rock>
    Outsource rock builds to Launchpad <outsource-rock-builds-to-launchpad>
+   Define the parts in your project file </common/craft-parts/how-to/use_parts>
```
tigarmo, line 19: @medubelko I added you to this review because I'd like your opinion on where we should place this new how-to (looks like [this](https://canonical-ubuntu-documentation-library--838.com.readthedocs.build/rockcraft/en/838/common/craft-parts/how-to/use_parts/). We can a) leave it here, b) put it somewhere else, c) exclude it or d) something else?

Also I think we have a bug in the doc: It has two top-level sections, which I think is what makes it get two entries in this TOC (neither of which is that title I gave it):
image

lengau (approved):

mr-cal (approved):

docs/how-to/index.rst
medubelko: > We can a) leave it here, b) put it somewhere else, c) exclude it or d) something else?

My feeling about it is that it's too generic, clearly flavoured for Craft Parts' needs, but not Rockcraft's. The title doesn't speak to a realistic task in this app – would a user ever want to write exactly one part? Would the dump plugin help them build a rock in isolation?

I think at best it would be better to link to this page in Craft Parts as general guidance if the reader needs a refresher on parts. Otherwise, I'd rather rely on the tutorial and supplementary, specific how-to material that serves Rockcraft with intention. If I had the choice I'd write a how-to about defining parts, multiple, that serve several useful cases _for rocks_.

The base tutorial already covers this task, with more purposeful direction toward a tangible goal – the hello world rock.

> we have a bug in the doc: It has two top-level sections

Oof, that's no good. It would be nice if Sphinx sounded the alarm when that happened. Fixed in https://github.com/canonical/craft-parts/pull/1049.

uv.lock
```diff
@@ -226,23 +226,24 @@ filecache = [
 
 [[package]]
 name = "canonical-sphinx"
-version = "0.3.0"
+version = "0.4.0"
```
medubelko, line 229: You beat me to the punch on this one. Any chance you could incorporate the appropriate [config changes](https://github.com/canonical/canonical-sphinx/pull/74) to enable the edit button?


> * If `html_theme_options.source_edit_link` is set in conf.py and points to a hosted Git repo, the edit button is enabled. The button uses the link in the `github_folder` (old) and `repo_folder` (new) settings.
> ...
> * Rename `github_folder` config setting to `repo_folder`.
> * Rename `github_version` config setting to `repo_default_branch`.

tigarmo: Sure thing. If I'm reading this right, the only thing that was missing was setting `html_theme_options.source_edit_link`.

A caveat here is that the button works fine for pages that are part of rockcraft's docs (e.g. the main page), but cannot work for pages that come from craft-parts (it just 404's)

docs/how-to/index.rst
tigarmo: Thanks! With all of these in mind, I decided to remove this page from rockcraft's docs for now

uv.lock
medubelko: > cannot work for pages that come from craft-parts

Understood. It's an unfortunate problem that we can't really design around because it's a side effect of how we're bringing in content. Other than hiding the button and handling all feedback about it, I don't know of a good UX for pages that are completely transcluded.

medubelko: > If I'm reading this right, the only thing that was missing was setting html_theme_options.source_edit_link.

We'll need to change `github_folder` and `github_version`, too. The build will provide a nag message otherwise.

tigarmo: Those fields were already absent in `conf.py`, am I missing something? It builds cleanly, and the edit button [seems to be working fine](https://canonical-ubuntu-documentation-library--838.com.readthedocs.build/rockcraft/en/838/)

medubelko: My mistake, was looking at the wrong keys. Looks good!

medubelko (approved):

PR #843, opened 2025-03-24
chore(merge): 'starbase/main' into 'main'

tigarmo:
This PR "properly" brings in Starbase as a regular merge. Since we'd done a "manual merge" a short while ago, the merge commit leaves in all conflicts (markers and all), which are then fixed in the next few commits.

Therefore, I think it's most useful to review this per-commit, starting at 001e1215475c88bc7f (`chore: fix conflicts in .github/`). Each commit addresses one broken bit, and most are reasonably small.


README.md
upils, line 1: The content of this file must be adapted based on `README.rst` (that must then be removed).

pyproject.toml
```diff
@@ -104,10 +153,6 @@ build-backend = "setuptools.build_meta"
 [tool.setuptools.dynamic]
 readme = {file = "README.rst"}
 
-[tool.setuptools.packages.find]
```
upils, line 107: question: I am not sure why that was needed but are you sure you do not need that anymore?

pyproject.toml
```diff
@@ -133,18 +193,37 @@ use_parentheses = true
...
+
+[tool.coverage.report]
+skip_empty = true
+exclude_also = [
+    "if (typing\\.)?TYPE_CHECKING:",
+]
 
 [tool.pyright]
-ignore = [
+strict = ["starcraft"]
```
upils, line 219:
```suggestion
strict = ["rockcraft"]
```

pyproject.toml
```diff
@@ -189,73 +268,118 @@ src = ["rockcraft", "tests"]
 extend-exclude = [
     "docs",
     "__pycache__",
-    "rockcraft/_version.py"
+    "rockcraft/_version.py",
+    "tests/spread",
```
upils, line 272: question: what was the issue? Is that because this folder holds python files we do not want to lint?

Makefile
```diff
@@ -36,36 +38,38 @@ publish: publish-pypi  ## Publish packages
...
-ifneq ($(OS),Linux)
+install-build-deps: install-lint-build-deps
+ifeq ($(APT_PACKAGES),)
 else ifeq ($(shell which apt-get),)
-	$(warning apt-get not found. Please install dependencies yourself.)
+	$(warning Cannot install build dependencies without apt.)
+	$(warning Please ensure the equivalents to these packages are installed: $(APT_PACKAGES))
 else
-	sudo $(APT) install libyaml-dev python3-dev python3-pip python3-setuptools \
-	  python3-venv python3-wheel fuse-overlayfs libapt-pkg-dev umoci libgit2-dev
```
upils, line 64: `libgit2-dev` was lost in the change. Is that on purpose?

Makefile
```diff
@@ -36,36 +38,38 @@ publish: publish-pypi  ## Publish packages
...
-endif
-ifneq ($(shell which snap),)
-	sudo snap install lxd
```
upils, line 67: Installing `lxd` was also removed here. Is that on purpose?

HACKING.md
```diff
@@ -0,0 +1,318 @@
+# Starcraft
```
upils, line 1: This file must be adapted. I understand it should contains the content of the existing `CONTRIBUTING.rst`.

.github/CODEOWNERS
upils, line 1: The content looks good but slipping this in this "big" PR to rebase on starbase feels a bit risky and could have been overlooked.

SECURITY.md
upils, line 1: question: What about removing this file for now? (especially since you are working on PR to add a approved one)

rockcraft/services/package.py
```diff
@@ -185,9 +185,6 @@ def _pack(
         datetime.datetime.now(datetime.timezone.utc).isoformat(), base_digest
     )
     rock_metadata["architecture"] = build_for
-    # TODO: add variant to rock_metadata too
```
upils, line 188: `TODO`s in the code were removed. What about creating issues for those that were still making sense?

upils (changes requested):

pyproject.toml
bepri: It is necessary for setuptools to know what folders to include when packaging a wheel, but this is just a bad diff. It made it into the new revision: https://github.com/canonical/rockcraft/pull/843/commits/37ec34a8108d33c323f7b7c00678d2016c4fc192#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R177

.github/CODEOWNERS
bepri: Starbase has a CODEOWNERS template, which is where these changes came from. Do you have any suggestions for how Starbase merges could be done without this problem? I agree that it seems risky - the file says a review should be required by Sergio for any changes to it and yet it doesn't look like that actually triggered.

bepri (approved):
Looks great, which is probably why my only input is picky. Feel free to take it or leave it!

.github/workflows/qa.yaml
```diff
@@ -16,6 +16,4 @@ jobs:
     uses: canonical/starflow/.github/workflows/test-python.yaml@main
     with:
       lowest-python-version: ""
-      # Disable 3.11 and 3.13 while wheels for python-apt and pygit2 are not
```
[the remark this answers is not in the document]
tigarmo: yes; the comment doesn't apply anymore (I think), but the choice not to use 3.11 and 3.13 for the moment is explicit

Makefile
tigarmo: yes; I took the opportunity to do a bit of cleaning up and added back the apt packages as they were needed (they weren't needed apparently)

Makefile
tigarmo: Yes; I don't know why that was there, there's already a target to pack the snap in starbase's makefile

.github/CODEOWNERS
tigarmo: In this case I'm opting-into using CODEOWNERS, this is why it's in the PR
As for the trigger: github uses the CODEOWNERS file from the PR's target branch (in this case `main`), to avoid one being able to get around the owners by removing/editing the file

.gitignore
```diff
@@ -138,20 +140,24 @@ dmypy.json
 .direnv
```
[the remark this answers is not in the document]
tigarmo: thanks for pointing this out; I reorganized both this file and `pyproject.toml` to better follow starbase's ordering

pyproject.toml
tigarmo: Update: I reorganized this file so better follow starbase's ordering

pyproject.toml
tigarmo: thanks for this! Turns out we had a bunch of pyright issues to address

upils (approved):

PR #845, opened 2025-03-26
fix: sanitize summary for OCI annotation

cjdcordeiro:

---

Sanitize the `summary` field before constructing the `"org.opencontainers.image.description"` annotation, such that we avoid causes where the `summary` itself already includes multiple consecutive `\n` chars, thus breaking the existing convention of a double-newline-delimiter for the annotation.

### DIFF
--- a/rockcraft/models/project.py
+++ b/rockcraft/models/project.py
@@ -441,7 +441,8 @@ class Project(BuildPlanner, BaseProject):  # type: ignore[misc]
             "org.opencontainers.image.ref.name": self.name,
             "org.opencontainers.image.created": generation_time,
             "org.opencontainers.image.base.digest": base_digest.hex(),
-            "org.opencontainers.image.description": f"{self.summary}\n\n{self.description}",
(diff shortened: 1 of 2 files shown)


rockcraft/models/project.py
```diff
@@ -441,7 +441,8 @@ def generate_metadata(
             "org.opencontainers.image.ref.name": self.name,
             "org.opencontainers.image.created": generation_time,
             "org.opencontainers.image.base.digest": base_digest.hex(),
-            "org.opencontainers.image.description": f"{self.summary}\n\n{self.description}",
+            "org.opencontainers.image.description": re.sub(r"\n{2,}", " ", self.summary)
```
tigarmo, line 444: shouldn't we replace the sequence of newlines with a single newline? It would preserve the original structure a bit better

cjdcordeiro: sure. I was just doing a `replace` originally, so the " " was to cope with more that 2 consecutive `\n`. but `{2,}` takes care of that. changed

tigarmo (approved):

PR #847, opened 2025-03-26
docs: update pebble page and add references to pebble docs

alesancor1:
- [x] Have you followed the guidelines for contributing?
- [x] Have you signed the [CLA](http://www.ubuntu.com/legal/contributors/)?
- [x] Have you successfully run `make lint && make test`?

---

This PR adds references to [pebble docs](https://documentation.ubuntu.com/pebble/) in the Pebble page inside rockcraft documentation. This page only keeps the information from pebble related to container images and rocks.

Additionally, other pages have been modified so when Pebble is mentioned it links to the pebble page.

---

See built page here: (waiting CI)

PR #855, opened 2025-04-07
docs: refactor tutorial

alesancor1:

---

This PR presents a new tutorial that covers the creation of a rock serving a static HTML file with NGINX via HTTP(s).

## **Rationale**

From @medubelko:

> The tutorial should be a singular, opinionated, end-to-end learning experience with a definite outcome. We’ve already achieved the latter goals in the Hello World page, but by introducing additional tutorial outcomes – for chisel, for Flask, and so forth – we’ve undermined the first two goals. The outcome is to containerise something useful, which means outside software. But the tutorial shouldn’t retread ground and cover multiple similar outcomes, making them compete with each other

Additionally, the toc-tree from how-to guides has been grouped, since the old tutorials have been moved there, and there were many guides, making it hard to navigate.

---

Other PRs affected by these changes:

* https://github.com/canonical/rockcraft/pull/847 - modifies `/tutorial/node-app.rst` which has been moved to `how-to/rocks/pack-rocks`
* https://github.com/canonical/rockcraft/pull/846 - should be closed if this is merged. New tutorial already include this.
* https://github.com/canonical/rockcraft/pull/836 - changes should be made to the tutorials on the new location
* https://github.com/canonical/rockcraft/pull/828 - changes should be made to the tutorials on the new location

---

See [built-docs](awaiting CI)


varshigupta12:
@alesancor1 While I understand the reasoning behind consolidating into a single tutorial experience, the 12-factor charm tutorials are not traditional how-to guides. Instead, they walk users through steps to leverage native framework support in Rockcraft for auto-generating rocks. This is a distinct objective compared to the general Rockcraft tutorial, which focuses on creating a rock from scratch.

The framework-based approach offers a faster path for users to generate rocks using Rockcraft functionality and serves a different purpose. If we’re moving toward a unified tutorial experience, we should ensure there's a clear path within the main tutorial that links to the framework-specific guidance, so users can easily find what they need when, for example, building a rock for a Flask app.

Alternatively, we could still retain the framework tutorials under the main tutorials section, but organize them under a new heading—something like “Native Framework Support in Rockcraft”—to reflect their focus and intent.

sergiusens (changes requested):
I don't know exactly where this is coming from, but this is indeed in our plans, perhaps not in this way.

Before making such changes I would advise to first gather stakeholders related to the product as it is coming as a surprise, at least in timing.

Yes we do eventually want one tutorial, so the spirit of this is there, but we have an experience to cater for (12 factor) and we are going to be discussing plans for that during our Engineering Sprint in Frankfurt.

cjdcordeiro:
> I don't know exactly where this is coming from, but this is indeed in our plans, perhaps not in this way.
> 
> Before making such changes I would advise to first gather stakeholders related to the product as it is coming as a surprise, at least in timing.
> 
> Yes we do eventually want one tutorial, so the spirit of this is there, but we have an experience to cater for (12 factor) and we are going to be discussing plans for that during our Engineering Sprint in Frankfurt.

Just for context, this was one of the [items](https://warthogs.atlassian.net/browse/ROCKS-1275) agreed upon during the UX session in our last sprint.

This proposal is, however, just that - a proposal. We should definitely decide how to preserve the 12-factor tutorial, especially if they have already been referenced elsewhere (blogs, etc.) without pinning a specific version of the docs.

Feel free to hold this PR until the next sprint, and pls do invite us to take part in such discussions.

The work here should also be preserved, regardless of the 12-factor tutorials, since the latter provide a more automated UX and thus may not cover some of the more basic and fundamental steps of building a rock, as show in this PR's new tutorial.

PR #856, opened 2025-04-07
refactor: use craft-application 5

bepri:

---


mr-cal (commented):
Looks good, I have no concerns with the direction of this.  I'm curious to see the spread tests run.

rockcraft/extensions/gunicorn.py
```diff
@@ -367,7 +367,7 @@ def _check_wsgi_path(self) -> None:
             )
         if not has_global_variable(wsgi_file, "application"):
             raise ExtensionError(
-                "django application can not be imported from {self.default_wsgi_path}, "
```
tigarmo, line 370: hehe

rockcraft/services/project.py
```diff
@@ -0,0 +1,38 @@
...
+class RockcraftProjectService(ProjectService):
+    """Rockcraft-specific project service."""
+
+    @staticmethod
+    @override
+    def _app_preprocess_project(
+        project: dict[str, Any], *, build_on: str, build_for: str, platform: str
+    ) -> None:
+        """Apply preprocessing for Rockcraft projects."""
+        project_root = Path.cwd()
```
tigarmo, line 36: should you use `self._project_dir` here instead?

snap/snapcraft.yaml
```diff
@@ -127,7 +127,7 @@ parts:
       # Extensions currently expect to find data files in share/rockcraft/
       "**/site-packages/extensions": share/rockcraft/extensions
     override-build: |
-      uv export --no-dev --extra store --no-emit-workspace --no-emit-package pywin32 --output-file uv-requirements.txt
+      uv export --no-dev --no-hashes --extra store --no-emit-workspace --no-emit-package pywin32 --output-file uv-requirements.txt
```
tigarmo, line 130: reminder to drop this

tests/conftest.py
```diff
@@ -285,3 +191,202 @@ def project_main_module() -> types.ModuleType:
...
+def fake_project_file(in_project_path, fake_project_yaml) -> Path:
+    project_file = in_project_path / "rockcraft.yaml"
+    project_file.write_text(fake_project_yaml)
+
+    return project_file
+
+
+@pytest.fixture
+def project_keys(request: pytest.FixtureRequest) -> dict[str, Any]:
+    """Fixture to modify the default project as needed.
```
tigarmo, line 357: 👍🏾

tests/integration/services/test_lifecycle.py
```diff
@@ -20,20 +20,24 @@
...
 from craft_application.util import repositories
 from craft_parts import overlays
+from rockcraft.services.image import ImageInfo
 
 from tests.testing.project import create_project
 from tests.util import jammy_only
 
 pytestmark = [
-    jammy_only,
+    # jammy_only,
```
tigarmo, line 29:
```suggestion
```

tigarmo (commented):
looks great :)
thanks a lot

rockcraft/services/image.py
```diff
@@ -65,25 +67,28 @@ def _create_image_info(self) -> ImageInfo:
...
+
+        if len(build_plan) != 1:
             raise errors.MultipleBuildsError
 
-        build_for = self._build_plan[0].build_for
-        project = cast(models.Project, self._project)
-        if project.base == "bare":
+        build_for = build_plan[0].build_for
+        project = self._services.get("project").get()
+        base = cast(str, project.effective_base)
```
tigarmo, line 77: this must be the project's actual `base`, and *not* the "build-base if build-base else base"

rockcraft/services/service_factory.py
```diff
@@ -16,38 +16,36 @@
...
+    "package": "RockcraftPackageService",
+    "provider": "RockcraftProviderService",
+    "project": "RockcraftProjectService",
+    "remote_build": "RockcraftRemoteBuildService",
+    "image": "RockcraftImageService",
+}
 
 
-@dataclass
 class RockcraftServiceFactory(ServiceFactory):
```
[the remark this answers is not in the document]
bepri: Only the `TYPE_CHECKING` box is helpful in any way, but even that is gonna disappear once we fully deprecate `__getattr__` in favor of `get` on the service factory. Do you want to keep this until then, or just rip off the bandaid now?

tigarmo (approved):

medubelko (approved):
Approving, but I'm not involved in testing this...

rockcraft/services/project.py
lengau: FWIW this is a potential reason to de-static this method or provide another way to access the project directory.

tests/conftest.py
lengau: Hol'up really?

I ask because when @bepri and I were writing it I thought you would have the most/best ideas for improvements.

Context: I really want to add a fixture like this to craft-application's pytest plugin so we can have a uniform way to generate projects, and this was a first attempt. (The craft-application one would come with some serious docs about setting up your app's tests)

rockcraft/services/package.py
```diff
@@ -57,24 +42,22 @@ def pack(self, prime_dir: pathlib.Path, dest: pathlib.Path) -> list[pathlib.Path
...
-        image_service = services.image
+        image_service = cast(RockcraftImageService, self._services.get("image"))
         image_info = image_service.obtain_image()
 
-        if not self._build_plan:
-            raise errors.EmptyBuildPlanError
+        build_plan = self._services.get("build_plan").plan()
 
-        if len(self._build_plan) > 1:
+        if len(build_plan) > 1:
```
sergiusens, line 52: minor consistency issue, but above you do `!= 1` (https://github.com/canonical/rockcraft/pull/856/files#diff-1c50437519bc7c33b649e77b5197022dadb1baad316acc13d1822c1a7cb6f327R72)

I am not certain about the behavior of len being 0 here though, for which this raised error would be incorrect

rockcraft/services/image.py
```diff
@@ -65,25 +67,28 @@ def _create_image_info(self) -> ImageInfo:
         image_dir = self._work_dir / "images"
         bundle_dir = self._work_dir / "bundles"
 
-        if len(self._build_plan) != 1:
+        build_plan = self._services.get("build_plan").plan()
+
+        if len(build_plan) != 1:
```
sergiusens, line 72: > 1 for the raised error to make sense, not sure what happens when this is 0

sergiusens (approved):
Large, looks good :fist_right: :fist_left: , did some fly by comments.

> [!CAUTION]
> But this was **not** an 8 :man_facepalming:

tests/conftest.py
tigarmo: I'm totally game for doing a craft-app design of such a fixture, but this is plenty good for what we need now (which is to merge this PR)

rockcraft/services/package.py
bepri: I'll change the other one to be `> 1` as it's more semantically clear what we're checking and why, imo.

PR #861, opened 2025-04-15
chore: set UV_USE_IO_URING=0 on npm builds

tigarmo:
I should note for the rustaceans in the crowd that this refers to [libuv](https://github.com/libuv/libuv), which node uses to do async io

### DIFF
--- a/docs/tutorial/code/expressjs/task.yaml
+++ b/docs/tutorial/code/expressjs/task.yaml
@@ -12,6 +12,7 @@ summary: Getting started with ExpressJS tutorial
 priority: -200
 
 environment:
+  UV_USE_IO_URING: "0"
 
 execute: |
   # [docs:install-deps]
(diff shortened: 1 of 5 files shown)


medubelko (approved):

mr-cal (approved):

docs/tutorial/code/expressjs/task.yaml
```diff
@@ -12,6 +12,7 @@ summary: Getting started with ExpressJS tutorial
 priority: -200
 
 environment:
+  UV_USE_IO_URING: "0"
```
bepri, line 15: Should this get the same "NOTE" comment as the other places?

tests/spread/rockcraft/extension-expressjs/task.yaml
```diff
@@ -3,6 +3,7 @@ environment:
   SCENARIO/bare: bare
   SCENARIO/base_2404: ubuntu-24.04
   ROCKCRAFT_ENABLE_EXPERIMENTAL_EXTENSIONS: "true"
+  UV_USE_IO_URING: "0"
```
bepri, line 6: Same question here

bepri (approved):
LGTM, just a small question. I would also like to note that "uv" in that program's name stands for "unicorn velociraptors" and I think that's fantastic.

docs/tutorial/code/expressjs/task.yaml
tigarmo: I think we have enough warnings; ideally when this gets fixed we'll just revert this commit wholesale, or at the very least we can search for `UV_USE_IO_URING`

PR #863, opened 2025-04-17
build(deps): update craft-parts and craft-platforms

tigarmo:
The craft-parts update brings in the 'cargo-use' plugin, and the craft-platforms
update allows us to use "ubuntu:22.04" deprecated notation again.

### DIFF
--- a/docs/reference/plugins.rst
+++ b/docs/reference/plugins.rst
@@ -13,6 +13,7 @@ Rockcraft.
    /common/craft-parts/reference/plugins/dotnet_plugin
    plugins/ant_plugin
    /common/craft-parts/reference/plugins/autotools_plugin
+   /common/craft-parts/reference/plugins/cargo_use_plugin
    /common/craft-parts/reference/plugins/cmake_plugin
    /common/craft-parts/reference/plugins/dump_plugin
    /common/craft-parts/reference/plugins/go_plugin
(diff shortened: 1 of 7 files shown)


docs/reference/plugins.rst
```diff
@@ -13,6 +13,7 @@ Rockcraft.
    /common/craft-parts/reference/plugins/dotnet_plugin
    plugins/ant_plugin
    /common/craft-parts/reference/plugins/autotools_plugin
+   /common/craft-parts/reference/plugins/cargo_use_plugin
```
tigarmo, line 16: @medubelko I suppose you got added because of this line

bepri (approved):

medubelko (approved):

rockcraft/extensions/_utils.py
```diff
@@ -64,7 +64,7 @@ def _apply_extension(
     if "parts" not in yaml_data:
         yaml_data["parts"] = {}
 
-    parts: dict[str, Any] = yaml_data["parts"]
+    parts = cast(dict[str, Any], yaml_data["parts"])
```
lengau, line 67: It doesn't really matter, but quoting the first value passed to `cast` will let us keep any related imports in `if CHECKING:` blocks

```suggestion
    parts = cast("dict[str, Any]", yaml_data["parts"])
```

lengau (approved):

PR #862, opened 2025-04-16
feat(project): use the newest craft-app schema

lengau:
NOT YET READY TO MERGE (craft-application needs a release for it).

This uses the latest platforms schema changes in review in craft-application (https://github.com/canonical/craft-application/pull/735) to update the rockcraft schema.

- [ ] Have you followed the guidelines for contributing?
- [ ] Have you signed the [CLA](http://www.ubuntu.com/legal/contributors/)?
- [ ] Have you successfully run `make lint && make test`?

---

PR #867, opened 2025-04-23
feat(project): update Platform schema  

tigarmo:
Supersedes #862

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,6 +43,17 @@ publish: publish-pypi  ## Publish packages
 publish-pypi: clean package-pip lint-twine  ##- Publish Python packages to pypi
 	uv tool run twine upload dist/*
 
+schema: install-uv  ## Generate the schema file.
+	mkdir -p schema
+	uv run python tools/schema/schema.py > schema/rockcraft.json
+
+validate-schema: install-ajv-cli
+	# Find all the rockcraft.yaml files that don't contain "# pragma: no-schema-validate"
+	find . -type f -name rockcraft.yaml -exec grep -HL '# pragma: no-schema-validate' '{}' '+' | \
+	xargs -I {} ajv validate --strict=false --spec=draft2020 -s schema/rockcraft.json -d {}
+    # The line below can be used to use the go-based jv command instead.
+    # xargs -I {} sh -c 'echo "\e[32mValidating {}\e[0m"; jv schema/rockcraft.json {}'
+
 # Find dependencies that need installing
 APT_PACKAGES :=
 ifeq ($(wildcard /usr/include/libxml2/libxml/xpath.h),)
@@ -78,3 +89,11 @@ endif
 # If additional build dependencies need installing in order to build the linting env.
 .PHONY: install-lint-build-deps
 install-lint-build-deps:
+
+
+.PHONY: install-ajv-cli
+install-ajv-cli: install-npm
+ifneq ($(shell which ajv),)
+else
+	npm install -g ajv-cli
+endif
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,7 +3,7 @@ name = "rockcraft"
 description = "Create rocks"
 dynamic = ["version", "readme"]
 dependencies = [
-    "craft-application>=5.0.1",
+    "craft-application~=5.1",
     "craft-archives",
     "craft-cli",
     "craft-parts",
--- a/rockcraft/models/project.py
+++ b/rockcraft/models/project.py
@@ -99,7 +99,6 @@ class Project(BaseProject):
     services: dict[str, Service] | None = None
     checks: dict[str, Check] | None = None
     entrypoint_service: str | None = None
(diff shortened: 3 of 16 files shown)


pyproject.toml
```diff
@@ -3,7 +3,9 @@ name = "rockcraft"
 description = "Create rocks"
 dynamic = ["version", "readme"]
 dependencies = [
-    "craft-application>=5.0.1",
+    # "craft-application>=5.0.1",
+    # Undo before merge
+    "craft-application @ git+https://github.com/canonical/craft-application.git@main",
```
tigarmo, line 8: This will be fixed before merging

Makefile
```diff
@@ -43,6 +43,17 @@ publish: publish-pypi  ## Publish packages
...
+	mkdir -p schema
+	uv run python tools/schema/schema.py > schema/rockcraft.json
+
+validate-schema: install-ajv-cli
+	# Find all the rockcraft.yaml files that don't contain "# pragma: no-schema-validate"
+	find . -type f -name rockcraft.yaml -exec grep -HL '# pragma: no-schema-validate' '{}' '+' | \
+	xargs -I {} ajv validate --strict=false --spec=draft2020 -s schema/rockcraft.json -d {}
+#	The line below can be used to use the go-based jv command instead.
+# 	xargs -I {} sh -c 'echo "\e[32mValidating {}\e[0m"; jv schema/rockcraft.json {}'
+
```
bepri, line 56: A similar instance of this exists in Charmcraft too - should we upstream this next? cc @lengau

snap/snapcraft.yaml
```diff
@@ -127,7 +127,7 @@ parts:
       # Extensions currently expect to find data files in share/rockcraft/
       "**/site-packages/extensions": share/rockcraft/extensions
     override-build: |
-      uv export --no-dev --extra store --no-emit-workspace --no-emit-package pywin32 --output-file uv-requirements.txt
+      uv export --no-hashes --no-dev --extra store --no-emit-workspace --no-emit-package pywin32 --output-file uv-requirements.txt
```
bepri, line 130: reminder to fix this as well

tests/spread/rockcraft/destructive/rockcraft.yaml
```diff
@@ -1,3 +1,4 @@
+# pragma: no-schema-validate
```
bepri, line 1: Why is this one not being validated? I could be missing something but it seems validate-able to me

Makefile
```diff
@@ -43,6 +43,17 @@ publish: publish-pypi  ## Publish packages
...
+schema: install-uv  ## Generate the schema file.
+	mkdir -p schema
+	uv run python tools/schema/schema.py > schema/rockcraft.json
+
+validate-schema: install-ajv-cli
+	# Find all the rockcraft.yaml files that don't contain "# pragma: no-schema-validate"
+	find . -type f -name rockcraft.yaml -exec grep -HL '# pragma: no-schema-validate' '{}' '+' | \
+	xargs -I {} ajv validate --strict=false --spec=draft2020 -s schema/rockcraft.json -d {}
+#	The line below can be used to use the go-based jv command instead.
+# 	xargs -I {} sh -c 'echo "\e[32mValidating {}\e[0m"; jv schema/rockcraft.json {}'
```
bepri, line 55:
```suggestion
  #	The line below can be used to use the go-based jv command instead.
  # xargs -I {} sh -c 'echo "\e[32mValidating {}\e[0m"; jv schema/rockcraft.json {}'
```
I fear the whitespace sensitivity of makefiles

rockcraft/models/project.py
```diff
@@ -99,7 +99,6 @@ class Project(BaseProject):
     services: dict[str, Service] | None = None
     checks: dict[str, Check] | None = None
     entrypoint_service: str | None = None
-    platforms: dict[str, Platform | None]  # type: ignore[assignment]
```
bepri, line 102: This is surprising, was it overriding a better definition from a parent?

bepri (approved):
LGTM, just two questions for you (everything else is a nit or a general comment that can be ignored)

tests/spread/rockcraft/destructive/rockcraft.yaml
tigarmo: the `ubuntu:22.04` deprecated base is not part of the schema today (it gets converted into `ubuntu@22.04` early in validation)

rockcraft/models/project.py
tigarmo: it used to be `platforms: dict[str, Platform]`, which meant that
```
platforms:
  amd64:
```

... failed validation

Makefile
lengau: Yes, but it's kinda low priority. If we don't do it until 26.04 is out, we can use the go-based jv command instead and save the headache of more npm packages :-)

lengau (approved):
LGTM, thanks!

PR #878, opened 2025-05-06
feat(oci): inject mediatype in manifest

zhijie-yang:

---

This PR make `rockcraft` add the field `mediaType` in the image manifest, which is missing in the default manifest created by `umoci`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,6 +74,9 @@ endif
 ifeq ($(wildcard /usr/share/doc/umoci/copyright),)
 APT_PACKAGES += umoci
 endif
+ifeq ($(wildcard /usr/share/doc/skopeo/copyright),)
+APT_PACKAGES += skopeo
+endif
 
 # Used for installing build dependencies in CI.
 .PHONY: install-build-deps
--- a/rockcraft/oci.py
+++ b/rockcraft/oci.py
@@ -49,6 +49,8 @@ REGISTRY_URL = ECR_URL
 # The number of times to try downloading an image from `REGISTRY_URL`.
 MAX_DOWNLOAD_RETRIES = 5
 
+MANIFEST_MEDIA_TYPE = "application/vnd.oci.image.manifest.v1+json"
+
 
 @dataclass(frozen=True)
 class Image:
@@ -147,8 +149,10 @@ class Image:
         # umoci config, but not the variant. Need to do it manually
         _config_image(image_target, ["--architecture", mapping.go_arch, "--no-history"])
 
-        if mapping.go_variant:
-            _inject_architecture_variant(Path(image_target_no_tag), mapping.go_variant)
+        _inject_oci_fields(
+            image_target,
+            arch_variant=mapping.go_variant,
+        )
 
         # for new OCI images, the source image corresponds to the newly generated image
         return (
@@ -301,6 +305,15 @@ class Image:
         image_path = self.path / self.image_name
         output: bytes = _process_run(
             ["umoci", "stat", "--json", "--image", str(image_path)]
+        ).stdout
+        result: dict[str, Any] = json.loads(output)
+        return result
+
+    def get_manifest(self) -> dict[str, Any]:
+        """Obtain the image manifest, as reported by "skopeo inspect --raw"."""
+        image_path = self.path / self.image_name
+        output: bytes = _process_run(
+            ["skopeo", "inspect", "--raw", f"oci:{str(image_path)}"]
         ).stdout
         result: dict[str, Any] = json.loads(output)
         return result
@@ -520,6 +533,16 @@ class Image:
         _config_image(image_path, annotation_params)
         emit.progress(f"Labels and annotations set to {labels_list}")
 
+    def set_media_type(
+        self,
+    ) -> None:
+        """Set the media type in the target image's manifest."""
+        image_path = self.path / self.image_name
+        _inject_oci_fields(
+            image_path,
+            arch_variant=None,
+        )
+
 
 def _copy_image(
     source: str,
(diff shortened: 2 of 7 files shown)


cjdcordeiro:
For reference, this is the original bug report: https://bugs.launchpad.net/ubuntu-docker-images/+bug/2107660

Makefile
```diff
@@ -74,6 +74,9 @@ endif
 ifeq ($(wildcard /usr/share/doc/umoci/copyright),)
 APT_PACKAGES += umoci
 endif
+ifeq ($(wildcard /usr/share/doc/skopeo/copyright),)
```
tigarmo, line 77: 👍🏾

tigarmo (approved):
thanks!

rockcraft/oci.py
```diff
@@ -305,6 +308,15 @@ def stat(self) -> dict[str, Any]:
         result: dict[str, Any] = json.loads(output)
         return result
 
+    def manifest(self) -> dict[str, Any]:
+        """Obtain the image manifest, as reported by "skopeo inspect --raw"."""
+        image_path = self.path / self.image_name
+        output: bytes = _process_run(
+            ["skopeo", "inspect", "--raw", f"oci:{str(image_path)}"]
+        ).stdout
```
cjdcordeiro, line 316: I know it's used in a couple more places in the code, but it begs the question: what happens if you're running in destructive mode and `skopeo` isn't installed on the host? I think it will fail.

If https://github.com/canonical/rockcraft/blob/bbeac3aaf984903564562f409e43ba1f220bd240/rockcraft/utils.py#L95 addresses this problem, then let's use that path resolver. Otherwise, I'd revert back to the old logic and find the manifest manually.

P.S. this might be applicable to https://github.com/canonical/rockcraft/blob/main/rockcraft/oci.py#L524 too.

rockcraft/oci.py
```diff
@@ -305,6 +308,15 @@ def stat(self) -> dict[str, Any]:
         result: dict[str, Any] = json.loads(output)
         return result
 
+    def manifest(self) -> dict[str, Any]:
```
cjdcordeiro, line 311:
```suggestion
    def get_manifest(self) -> dict[str, Any]:
```

cjdcordeiro, line 311: but wait, this is not used anywhere but the tests

rockcraft/oci.py
```diff
@@ -570,11 +582,14 @@ def _add_layer_into_image(
     _process_run([*cmd, "--history.created_by", " ".join(cmd)])
 
 
-def _inject_architecture_variant(image_path: Path, variant: str) -> None:
-    """Inject architecture variant into existing OCI Image config.
+def _inject_manifest(
```
cjdcordeiro, line 585:
```suggestion
def _inject_oci_fields(
```

Probably not the best name, but this function is touching more than just the manifest - it also changes the OCI config.

rockcraft/oci.py
```diff
@@ -147,8 +147,11 @@ def new_oci_image(
         # umoci config, but not the variant. Need to do it manually
         _config_image(image_target, ["--architecture", mapping.go_arch, "--no-history"])
 
-        if mapping.go_variant:
-            _inject_architecture_variant(Path(image_target_no_tag), mapping.go_variant)
+        _inject_manifest(
+            Path(image_target_no_tag),
+            arch_variant=mapping.go_variant,
+            add_media_type=True,
```
cjdcordeiro, line 153: hum, this feels unnecessary. When would it be False? As it stands, this will never change, so you can either:
 1. assume it is always True and avoid this unnecessary input, or
 2. actually check if `media_type` is already set, before injecting one

The latter is preferred IMO.

cjdcordeiro (changes requested):

rockcraft/oci.py
zhijie-yang: Nice suggestion, let's do the latter pattern: We always check if `mediaType` is there; if so, we skipping injecting it, and if not, we set it as `application/vnd.oci.image.manifest.v1+json`, wdyt?

rockcraft/oci.py
zhijie-yang: Nice catch.

rockcraft/oci.py
zhijie-yang: It seems we already have a similar implementation here[https://github.com/canonical/rockcraft/blob/22d0b2ae86f8a5a1cff66e26d55bab054db76bf4/rockcraft/oci.py#L317]. I'll test with destructive mode to make sure.

zhijie-yang: In the [implementation](https://github.com/canonical/rockcraft/blob/main/rockcraft/oci.py#L622) of `_process_run`, the command is already wrapped with `get_snap_command_path`, so there is no need to wrap with this path resolver.

rockcraft/oci.py
zhijie-yang: Fixed in 9cc530b.

rockcraft/oci.py
zhijie-yang: Fixed in 9cc530b.

rockcraft/oci.py
zhijie-yang: Fixed in 9cc530b.

zhijie-yang:
Updates in the last commit:

1. After examining the build image, I found adding another step of "set_media_type" in `services/package` is necessary either. Setting the media type during `oci.Image.new_oci_images` only sets the `mediaType` field for the base image's (i.e. `rockcraft_base`) manifest while keeping the manifest of the target image untouched.

2. When walking through the target image's manifest during `_pack`, the assumption that the manifest list contains only one item is violated, since we have the manifests for the base image and the target image in the target image's path (file system path) simultaneously. We therefore need to parse the manifest list and find the one with the matching tag as the target image to correctly find and update the manifest.

Remaining caveats:

- The `mediaType` is missing in the image's `index.json`. Currently, the Assemble Rock workflow in the OCI Factory adds this field to the index when assembling the multi-arch rock (which can also be a single-arch rock) using `buildah`. In the future, if the images are released using the direct artifact of `rockcraft`, we will be missing the `mediaType` in the image's index again.
- The unit and integration tests don't cover the logic in `services/package.py` well. They weren't able to tell that even with the changes in the initial three commits of this PR, it was not sufficient to fully fix this issue. I'd suggest adding a spread test to cover this feature, but I'd use some help on where to put it.

rockcraft/oci.py
cjdcordeiro: I assume that means it works well in destructive mode, so lets resolve this

rockcraft/oci.py
cjdcordeiro: is this intentional @tigarmo ?

cjdcordeiro:
> The mediaType is missing in the image's index.json. Currently, the Assemble Rock workflow in the OCI Factory adds this field to the index when assembling the multi-arch rock (which can also be a single-arch rock) using buildah. In the future, if the images are released using the direct artifact of rockcraft, we will be missing the mediaType in the image's index again.

I'd say let's fix this too, in a following PR

> The unit and integration tests don't cover the logic in services/package.py well. They weren't able to tell that even with the changes in the initial three commits of this PR, it was not sufficient to fully fix this issue. I'd suggest adding a spread test to cover this feature, but I'd use some help on where to put it. I've added an integration test for the feature introduced in this PR in https://github.com/canonical/rockcraft/commit/08c46fc2feecf056a168172001c422efcde47982. @cjdcordeiro Do we assume that the arch variant should appear in the config of the target image, and should we also add an integration test for arch variants here if so?

We should test that, but it should have its own commit, since the fix is unrelated with this PR, so if done in this PR, the existing git log should be cleaned. Otherwise, raise another PR

rockcraft/oci.py
```diff
@@ -570,49 +593,71 @@ def _add_layer_into_image(
...
+        and a.get("org.opencontainers.image.ref.name") == image_tag
+    ]
+    if not manifest_digests:
+        raise errors.RockcraftError(
+            f"Cannot find manifest for {image_tag} in {tl_index_path}"
+        )
+    if len(manifest_digests) > 1:
+        raise errors.RockcraftError(
+            f"Found multiple manifests for {image_tag} in {tl_index_path}"
+        )
```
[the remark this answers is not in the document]
zhijie-yang: It will only happen when the index is not valid. Since we are using `umoci` to create the image index, we are unable to guarantee the correctness of an image index that is not created by `rockcraft` directly.

cjdcordeiro (approved):
LGTM. Thanks.

Just asking @tigarmo to weigh in on https://github.com/canonical/rockcraft/pull/878#discussion_r2095302113

PR #909, opened 2025-06-29
[PROTOTYPE]: Interim bases

lengau:

---


rockcraft/commands/lifecycle.py
```diff
@@ -0,0 +1,101 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
```
lengau, line 1: This is duplicated in utils.py due to some import issues when generating the snap. I can delete this file

rockcraft/models/project.py
```diff
@@ -80,9 +80,30 @@
     ),
 ]
 
-BaseT = Literal["bare", "ubuntu@20.04", "ubuntu@22.04", "ubuntu@24.04"]
+BaseT = Literal[
+    "bare",
+    "ubuntu@20.04",
+    "ubuntu@22.04",
+    "ubuntu@24.04",
+    "ubuntu@25.04",
```
tigarmo, line 88: I don't think we need to support `25.04` (but I'm fine with leaving it in for testing purposes while it's a prototype)

tigarmo (commented):
Thanks! I think there are two pieces missing:
1) How do we update archive urls in destructive-mode?
2) We also need to update the archive urls in the overlay system, like we do for [package-repositories](https://github.com/canonical/craft-application/blob/main/craft_application/util/repositories.py#L590)

lengau:
@tigarmo 

>  How do we update archive urls in destructive-mode?

Good point. I've added both this and the overlay installer to the lifecycle service.

For production I think we should modify craft-archives to be able to modify the existing `/etc/apt/sources.list` or `/etc/apt/sources.list.d/ubuntu.sources` files - that sed command is just a quick hack.

> As per bullet point 2 in R4 on the spec, we need to tell craft-parts to pass --ack-unsupported-release when cutting chisel slices.

Thanks, I missed that. Working on it.

cjdcordeiro:
Cross-referencing chisel's complementary implementation for releases that are EOSS: https://github.com/canonical/chisel/pull/238/

lengau:
Superseded by #946

PR #924, opened 2025-07-22
docs: exclude reuse dir from search results

medubelko:
Removing it entirely from the node graph seems to do the trick.

---


jahn-junior (approved):
Ah, that explains why Charmcraft doesn't have the same issue. Looks like we already exclude them over there. Thanks for fixing this!

tigarmo (approved):

PR #927, opened 2025-07-23
ci: clean up OSV workflows

bepri:

---

Fixes up the OSV scanner workflows moving forward by using the same arguments [Snapcraft and Charmcraft use](https://github.com/canonical/snapcraft/blob/932bd92e510f1217a23551c9dd454bebdb771c2c/.github/workflows/policy.yaml#L16-L22) and removing a now-outdated config.

### DIFF
--- a/.github/workflows/policy.yaml
+++ b/.github/workflows/policy.yaml
@@ -15,7 +15,10 @@ jobs:
     uses: canonical/starflow/.github/workflows/scan-python.yaml@main
     with:
       packages: python-apt-dev
-      # requirements-noble.txt can't build on jammy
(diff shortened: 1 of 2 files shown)


upils (approved):
LGTM.

Note: Looking at the logs, it seems that osv is **still** looking at `requirements-noble.txt` (see https://github.com/canonical/rockcraft/actions/runs/16480559838/job/46593493067?pr=927#step:6:11).
Also it appears we could maybe make the scanning faster (it is not really slow so that is a nitpick) by also excluding both `tools/external` and `docs/sphinx-starter-pack` submodules.

PR #944, opened 2025-08-27
chore: pin gunicorn package in flask and django extensions

javierdelapuente:

---

Pin gunicorn library in Django and Flask extensions to the major 23. 

We could break users if a new Gunicorn revision comes up that introduce breaking changes.


rockcraft/extensions/gunicorn.py
```diff
@@ -97,7 +97,7 @@ def _gen_parts(self) -> dict[str, Any]:
                 "plugin": "python",
                 "stage-packages": stage_packages,
                 "source": ".",
-                "python-packages": ["gunicorn"],
+                "python-packages": ["gunicorn~=23.0"],
```
tigarmo, line 100: The PR description mentions pinning gunicorn to 24, was that a typo?

tigarmo, line 100: @javierdelapuente since Django and Flask are not experimental extensions, we need to be careful to not break existing users. What if a project requires an older version of gunicorn? Say, the project has a dependency that only works if gunicorn <23.

One alernative to would be to solve this with documentation instead. We can tell users that they should pin gunicorn to a major version that is suitable to their project in their requirements.txt file.

Out of curiosity, what was the initial motivation for this PR?

javierdelapuente: Yeah, sorry, updated.

PR #946, opened 2025-08-27
feat: enable interim bases

lengau:
This does the following:

1. Enable's craft-application's check_supported_base metadata
2. Adds plucky and questing as supported bases
3. Adds tests for plucky and questing.

---

### DIFF
--- a/rockcraft/models/project.py
+++ b/rockcraft/models/project.py
@@ -87,9 +87,24 @@ ProjectName = Annotated[
     ),
 ]
 
-BaseT = Literal["bare", "ubuntu@20.04", "ubuntu@22.04", "ubuntu@24.04"]
+BaseT = Literal[
+    "bare",
+    "ubuntu@20.04",
+    "ubuntu@22.04",
+    "ubuntu@24.04",
+    "ubuntu@25.04",
+    "ubuntu@25.10",
+]
 BuildBaseT = typing.Annotated[
-    Literal["ubuntu@20.04", "ubuntu@22.04", "ubuntu@24.04", "devel"] | None,
+    Literal[
+        "ubuntu@20.04",
+        "ubuntu@22.04",
+        "ubuntu@24.04",
+        "ubuntu@25.04",
+        "ubuntu@25.10",
+        "devel",
+    ]
+    | None,
     pydantic.Field(validate_default=True),
 ]
 
--- a/rockcraft/models/project.py
+++ b/rockcraft/models/project.py
@@ -92,7 +92,6 @@ BaseT = Literal[
     "ubuntu@20.04",
     "ubuntu@22.04",
     "ubuntu@24.04",
-    "ubuntu@25.04",
     "ubuntu@25.10",
 ]
 BuildBaseT = typing.Annotated[
@@ -100,7 +99,6 @@ BuildBaseT = typing.Annotated[
         "ubuntu@20.04",
         "ubuntu@22.04",
         "ubuntu@24.04",
-        "ubuntu@25.04",
         "ubuntu@25.10",
         "devel",
     ]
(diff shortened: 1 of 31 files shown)


rockcraft/models/project.py
```diff
@@ -87,9 +87,24 @@
     ),
 ]
 
-BaseT = Literal["bare", "ubuntu@20.04", "ubuntu@22.04", "ubuntu@24.04"]
+BaseT = Literal[
+    "bare",
+    "ubuntu@20.04",
+    "ubuntu@22.04",
+    "ubuntu@24.04",
+    "ubuntu@25.04",
```
tigarmo, line 95: What's the reasoning for adding plucky? Can we skip it?

rockcraft/services/lifecycle.py
```diff
@@ -69,9 +70,9 @@ def post_prime(self, step_info: StepInfo) -> bool:
 
 
 def _python_usrmerge_fix(step_info: StepInfo) -> None:
-    """Fix 'lib64' symlinks created by the Python plugin on ubuntu@24.04 projects."""
-    if step_info.project_info.base != "ubuntu@24.04":
-        # The issue only affects rocks with 24.04 bases.
+    """Fix 'lib64' symlinks created by the Python plugin on ubuntu@24.04+ projects."""
```
tigarmo, line 73: 👍🏾

spread.yaml
```diff
@@ -31,6 +31,9 @@ backends:
       - ubuntu-22.04-64:
           workers: 4
           storage: 40G
+      - ubuntu-24.04-64:
```
tigarmo, line 34: hmm do we need this?

rockcraft/models/project.py
```diff
@@ -87,9 +87,24 @@
...
+    "ubuntu@25.10",
```
tigarmo, line 96: I'm blanking on where we implement the behavior of devel bases: specifically here, if `base: ubuntu@25.10` then `build-base` **must** be `devel`

tigarmo: do we still support that via craft-application? [reference](https://github.com/canonical/rockcraft/commit/97e0481790e018d7eee390d65f371c8ad5cebb59)

tigarmo (changes requested):
thanks!

spread.yaml
lengau: Oops! No, not right now. That was related to me trying some stuff out and forgetting to remove it.

rockcraft/models/project.py
lengau: The maven spread test uses chisel, which does not yet have a questing branch.

Per separate conversation I'll remove plucky, turn off that test for plucky and add a task to turn it on for questing.

rockcraft/models/project.py
lengau: My answer to you is the worst of all possible answers!

1. Yes, we [do still support it](https://github.com/canonical/craft-application/blob/c5fdcf0521a7c026cace5883bd7590fc841536ed/craft_application/models/project.py#L273-L303) in craft-application
2. It's [currently a no-op 😬](https://github.com/canonical/craft-application/blob/c5fdcf0521a7c026cace5883bd7590fc841536ed/craft_application/models/project.py#L58-L64)
3. Now that we're getting up-to-date distro support data files, we should really rewrite it to use that. (`if today < base.release_date...`)

- [Short term](https://github.com/canonical/craft-application/issues/880)
- [Long term](https://github.com/canonical/craft-application/issues/881)

rockcraft/models/project.py
lengau: https://github.com/canonical/rockcraft/issues/953

rockcraft/models/project.py
lengau: Hmm... that's a good point, but it makes things complicated since different apps may want to do that at different times. I've put your comment in the issue so we can discuss further.

PR #947, opened 2025-08-27
build(deps): update Python to >=3.12

lengau:
Since we're targeting core24, this drops Python 3.10 and 3.11 support.

All changes in `.py` files were automatically made by running `make format` after the update.

---


lengau:
@mr-cal IIRC the Snapcraft thing was for the core22 snapcraft rock (we run it using the system Python), not because of destructive mode. If someone uses the snap with destructive mode they'll still get a working rockcraft, but we are limiting ourselves from making a rockcraft rock that runs on Ubuntu 22.04.

PR #957, opened 2025-09-05
Rocks 2419/allow single char names

alesancor1:

---

This PR allows rockcraft to produce Rocks with less than three characters in the name by removing such restriction in `rockcraft` and using the one from parent class in `craft-application`. To prevent the resulting pebble layer to fail the validation, a `-rockcraft-` prefix is automatically added to the layer name.

### Related PR

https://github.com/canonical/craft-application/pull/894

Context: Although rockcraft itself will allow names with a single character, it will fail the validation for `title`. This is because when `title` is not set, Rockcraft will set the `title` value to be equal to the `name`. However, `title` has a `min_length` of `2` while `name` has a min length of `1`. The PR above aims to fix this situation.

### DIFF
--- a/tests/unit/test_project.py
+++ b/tests/unit/test_project.py
@@ -24,9 +24,10 @@ import pydantic
 import pytest
 import yaml
 from craft_application.errors import CraftValidationError
+from craft_application.models.constraints import MESSAGE_INVALID_NAME
 from craft_providers.bases import ubuntu
 from rockcraft.models import Project
-from rockcraft.models.project import MESSAGE_INVALID_NAME, Platform
+from rockcraft.models.project import Platform
 from rockcraft.pebble import Service
 from rockcraft.services.project import RockcraftProjectService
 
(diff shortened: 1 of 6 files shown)


tests/unit/test_project.py
```diff
@@ -517,7 +520,7 @@ def test_project_name_valid(yaml_loaded_data, valid_name):
 
 
 @pytest.mark.parametrize(
-    "invalid_name", ["AAA", "0aaa", "a", "a--a", "aa-", "a:a", "a/a", "a@a", "a_a"]
+    "invalid_name", ["", "AAA", "a--a", "aa-", "a:a", "a/a", "a@a", "a_a"]
```
lengau, line 523: If the answer to my question about about the 40-character name limit is okay, do we want to test it here? (Maybe not if it's incidental and we're happy for that requirement to change upstream and simply adopt it.)

Likewise, should we add the 40-char limit to [the reference docs](https://canonical-ubuntu-documentation-library--957.com.readthedocs.build/rockcraft/957/reference/rockcraft.yaml/#name)?

PR #969, opened 2025-09-16
feat: usrmerged install dirs in 25.10+

tigarmo:
Probably easier to review per-commit


tests/spread/rockcraft/usrmerged-install/rockcraft.yaml
bepri, line 1: Is there any need to have a test ensuring the pre-25.10 behavior is kept, or can we expect other existing tests to fail if that breaks?

rockcraft/services/lifecycle.py
```diff
@@ -44,13 +44,21 @@ def setup(self) -> None:
         image_service = services.image
         image_info = image_service.obtain_image()
 
+        base = project.effective_base
+        usrmerged_by_default = True
+
+        # Bases older than 25.10 do not get usermerged install dirs by default
+        if base in ("ubuntu@20.04", "ubuntu@22.04", "ubuntu@24.04"):
```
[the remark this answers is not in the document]
lengau: `craft_platforms.DistroBase` is comparable, so one could write it as

```suggestion
        if DistroBase.from_str(base) < DistroBase("ubuntu", "25.10"):
```

but I'm not sure we need to. The `is_core24` piece is a problem because it's special-casing the new standard, whereas here it's special-casing the old cases.

The existing version is also perhaps less surprising, since if we add a non-ubuntu distribution the comparison would raise an error.

tests/spread/rockcraft/usrmerged-install/rockcraft.yaml
lengau: https://github.com/canonical/rockcraft/blob/main/tests/spread/rockcraft/usrmerge/rockcraft.yaml

I guess that could get variants to declare the different pre-25.10 bases, but I'm not sure there'd be utility to it.

lengau (approved):
Very nice! I'm thrilled to see just how little logic is required in the application to implement this change.

PR #974, opened 2025-09-19
refactor: consolidate base-specific knowledge

tigarmo:
Instead of having the information about which base behaves which way with regards to support, pebble, plugins, etc, have a single "bases" module that serves as an authoritative source for all base-related knowledge. The idea is that someone can look at this single module and have a full view of how Rockcraft's behavior changes depending on the project's base.

---

### DIFF
--- a/rockcraft/models/project.py
+++ b/rockcraft/models/project.py
@@ -29,10 +29,11 @@ from craft_application.models import (
 )
 from craft_application.models import Project as BaseProject
 from craft_application.models.base import alias_generator
-from craft_providers import bases
+from craft_providers import bases as providers_base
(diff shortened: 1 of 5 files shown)


rockcraft/models/project.py
```diff
@@ -56,22 +57,9 @@
...
 BuildBaseT = typing.Annotated[
-    Literal[
-        "ubuntu@20.04",
-        "ubuntu@22.04",
-        "ubuntu@24.04",
-        "ubuntu@25.10",
-        "devel",
-    ]
-    | None,
+    Literal[(*bases.SUPPORTED, "devel")] | None,
```
lengau, line 62: WHOA THAT WORKS?

PR #977, opened 2025-09-24
feat: use the Python plugin v2 for 25.10+

tigarmo:
The changeset is big but a lot of it is tests and glue code. I split the changes over commits:

- First commit is updating craft-parts to a dev version, pending a release;
- Second one adds a way to register plugins depending on the base. It's a bit awkward because the plugin registration happens very early, *before* we have a project (since we need to know the plugins in order to validate the plugin-specific keys);
- Third one adds a function to fix the shebangs after the lifecycle is done;
- Four one adds a few spread tests for the new plugin.


rockcraft/application.py
```diff
@@ -67,3 +69,14 @@ def _enable_craft_parts_features(self) -> None:
         # enable the craft-parts Features that we use here, right before
         # loading the project and validating its parts.
         Features(enable_overlay=True)
+
+    def _get_build_base(self) -> str | None:
+        """Get the project's build-base, if the project file exists."""
+        project_file = self.project_dir / "rockcraft.yaml"
```
lengau, line 75: You should be able to use the `ProjectService`'s `get_raw` here.

Since we'll be extending this functionality to other crafts, we should probably add a `get_build_base` to the `ProjectService` that does this for us.

rockcraft/services/lifecycle.py
```diff
@@ -73,22 +74,24 @@ def post_prime(self, step_info: StepInfo) -> bool:
         layers.prune_prime_files(prime_dir, files, base_layer_dir)
 
         _python_usrmerge_fix(step_info)
+        _python_v2_shebang_fix(step_info)
 
         return True
 
 
 def _python_usrmerge_fix(step_info: StepInfo) -> None:
     """Fix 'lib64' symlinks created by the Python plugin on ubuntu@24.04+ projects."""
```
lengau, line 83:
```suggestion
    """Fix 'lib64' symlinks created by the Python plugin on ubuntu@24.04 projects."""
```

bepri (approved):

rockcraft/application.py
tigarmo: oh good call. I thought that the loading of plugins happened early enough that the project service hadn't even been 'early configured', but I was wrong! Updated

medubelko (approved):

PR #982, opened 2025-10-01
feat(models): add documentation to the Project model

lengau:
This adds docstrings, examples, etc. to the Project model and its child models.

---

### DIFF
--- a/rockcraft/models/project.py
+++ b/rockcraft/models/project.py
@@ -81,15 +81,76 @@ class Project(BaseProject):
 
     # Type of summary is Optional[str] in BaseProject
     summary: str  # type: ignore[reportIncompatibleVariableOverride]
+    """A short summary describing the rock."""
     description: str  # type: ignore[reportIncompatibleVariableOverride]
+    """A longer, possibly multi-line description of the rock."""
     environment: dict[str, str] | None = None
+    """A set of key-value pairs specifying the environment variables to be added
+    to the base image's OCI environment.
+    """
     run_user: _RunUser = None
+    """The default OCI user. It must be a supported shared user. Currently, the only
+    supported shared user is ``_daemon_`` (with UID/GID 584792).
+    If not specified, the user ``root`` (UID/GID 0) will be used.
+    """
     services: dict[str, Service] | None = None
-    checks: dict[str, Check] | None = None
-    entrypoint_service: str | None = None
-    entrypoint_command: str | None = None
+    """Follows the Pebble :external+pebble:doc:`reference/layer-specification`.
+
+    A mapping of ``name: service`` for the Pebble entrypoint. It uses Pebble's layer
+    specification syntax, with each entry defining a Pebble service. For
+    each service, the ``override`` and ``command`` keys are mandatory, but all
+    others are optional.
+    """
+    checks: dict[str, Check] | None = pydantic.Field(
+        default=None,
+        description="Health checks for this service.",
+    )
+    """A mapping of health checks that can be configured to restart Pebble services
+    when they fail. It uses Pebble's layer specification syntax, with each
+    entry corresponding to a check.
+    """
+    entrypoint_service: str | None = pydantic.Field(
+        default=None,
+        description="The optional name of the Pebble service to serve as the entrypoint.",
+        examples=["my-service"],
+    )
+    """The optional name of the Pebble service to serve as the `OCI entrypoint
+    <https://specs.opencontainers.org/image-spec/config/?v=v1.0.1#properties>`_.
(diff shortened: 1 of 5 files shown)


docs/reference/rockcraft.yaml.rst
lengau, line 1: I added this in a separate commit so it's easy to revert, but I used kitbash and modified this page to test what the resulting text would look like.

[Rendered version](https://canonical-ubuntu-documentation-library--982.com.readthedocs.build/rockcraft/982/reference/rockcraft.yaml/)

rockcraft/pebble.py
```diff
@@ -16,81 +16,168 @@
...
+            raise CraftValidationError(
+                f"Multiple check types specified ({', '.join(sorted(check_types))}). "
+                "Each check must have exactly one type."
             )
-        else:
-            return values
 
-        raise CraftValidationError(err)
+
+Check = Annotated[
```
lengau, line 174: This change means that the schema is now aware that these fields are mutually exclusive. It's a tagged union, so we still get a nice error message out of it. No new tests are needed, as `_get_check_tag` is tested with the same tests that previously tested `_validates_check_type`. I did add some test cases though.

rockcraft/pebble.py
```diff
@@ -101,26 +188,112 @@ class Service(CraftBaseModel):
...
+    If both group and group-id are specified, the group's GID must match group-id.
+    """
     working_dir: str | None = None
-    on_success: Literal["restart", "shutdown", "ignore"] | None = None
-    on_failure: Literal["restart", "shutdown", "ignore"] | None = None
-    on_check_failure: dict[str, Literal["restart", "shutdown", "ignore"]] | None = None
-    backoff_delay: str | None = None
-    backoff_factor: float | None = None
+    """Working directory for the service command."""
+    on_success: Literal["restart", "shutdown", "failure-shutdown", "ignore"] | None = (
```
lengau, line 241: This as well as `success-shutdown` below I got from the pebble docs. I'm not entirely sure these are relevant values for rocks, though. LMK if this should be excluded.

lengau:
Linting issue is fixed in https://github.com/canonical/rockcraft/pull/984

docs/reference/rockcraft.yaml.rst
medubelko: You've done half the work of the followup PR! :P

Should probably save this for later...

lengau: I've made a copy of this branch here: https://github.com/canonical/rockcraft/tree/work/ROCKCRAFT-263-additional

I'm guessing we should still revert the rst file before merge here, but I'll leave it for now to help with review.

medubelko (commented):
As discussed earlier, there are some principles of key/setting descriptions I'd like to adhere to.

In the first paragraph of the description, the goal is to in as short a space as possible describe what the setting _represents_ or what it _does for the user_. There are a few guidelines

- Focus on what a setting achieves, rather than its data type; for a `recipes` key in a coffee making script, it shouldn't be "A dictionary of key-value pairs for coffee types" but "The names and descriptions of the available coffee brews". Iff necessary, we can delve into the data structures later.
- Don't create circular definitions; not "The temperature knob sets the coffee's temperature."

I've left some examples using these rules.

tigarmo (changes requested):
thanks!

rockcraft/models/project.py
```diff
@@ -81,15 +81,76 @@ class Project(BaseProject):
 
     # Type of summary is Optional[str] in BaseProject
     summary: str  # type: ignore[reportIncompatibleVariableOverride]
+    """A short summary describing the rock."""
```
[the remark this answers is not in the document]
lengau: Rockcraft has intentionally overridden the 78 character summary limit, though I'm not sure why.

rockcraft/pebble.py
```diff
@@ -16,81 +16,168 @@
...
-
-
-class Check(CraftBaseModel):
+    command: str = pydantic.Field(
+        description="The command line to execute for this health check.",
+        examples=["/usr/bin/health-check"],
+    )
+    service_context: str | None = pydantic.Field(
+        default=None,
+        description=textwrap.dedent(
```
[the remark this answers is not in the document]
lengau: Looking through a lot of these, I haven't done a great job at that either, so I'm fixing that.

rockcraft/models/project.py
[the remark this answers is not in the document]
lengau: @tigarmo this might be a relevant question for you, especially if we want to reintroduce the length limit in future.

tigarmo (approved):
thanks!

PR #994, opened 2025-10-07
docs: add a reference doc for the new python plugin

bepri:

---

Adds reference docs for the new Python plugin. 

ROCKCRAFT-261.

### DIFF
--- a/docs/reference/plugins/python_plugin.rst
+++ b/docs/reference/plugins/python_plugin.rst
@@ -15,7 +15,7 @@ Python plugin (ubuntu\@20.04)
 
     For base ``ubuntu@25.10`` and newer, Python parts build with
     :doc:`python_v2_plugin`. For an explanation of why there are two different versions
-    of this plugin, see <LINK-TO-USRMERGE-EXPLANATION>.
+    of this plugin, see :ref:`explanation-usrmerge-implementation`.
 
 .. include:: /common/craft-parts/reference/plugins/python_plugin.rst
     :start-after: .. _craft_parts_python_plugin-keywords:
--- a/docs/reference/plugins/python_v2_plugin.rst
+++ b/docs/reference/plugins/python_v2_plugin.rst
@@ -15,7 +15,7 @@ Python plugin (ubuntu\@25.10)
 
     For base ``ubuntu@24.04`` and lower, Python parts build with the
     :doc:`python_plugin`. For an explanation of why there are two different versions
-    of this plugin, see <LINK-TO-USRMERGE-EXPLANATION>.
+    of this plugin, see :ref:`explanation-usrmerge-implementation`.
 
 .. include:: /common/craft-parts/reference/plugins/python_v2_plugin.rst
     :start-after: .. _python_v2_plugin-keywords:


docs/reference/plugins/python_plugin.rst
```diff
@@ -1,8 +1,19 @@
+.. include:: /common/craft-parts/reference/plugins/python_plugin.rst
+    :end-before: .. _craft_parts_python_plugin-keywords:
+
+.. admonition:: Legacy plugin
+
+    This plugin is superseded by the modern :doc:`python_v2_plugin`.
+
+    It is selected by default when building with the ``ubuntu@25.04`` base or older.
+
+    For more information, see <LINK-TO-USRMERGE-EXPLANATION>.
```
bepri, line 10: Blocking comment so this doesn't get merged until the other document exists to be referenced.

medubelko, line 10:
```suggestion
    This reference is for the **Python (v1)** plugin, which is available for base 
    ``ubuntu@24.04`` or lower.
    
    For the updated version, see :doc:`python_v2_plugin`. For an explanation of why 
    there are different versions of this plugin, see <LINK-TO-USRMERGE-EXPLANATION>.
```

docs/reference/plugins/python_v2_plugin.rst
```diff
@@ -0,0 +1,13 @@
+.. include:: /common/craft-parts/reference/plugins/python_v2_plugin.rst
+    :end-before: .. _python_v2_plugin-keywords:
+
+.. admonition:: Updated plugin
+
+    This plugin supersedes the legacy :doc:`python_plugin`.
+
+    It is selected by default when building with the ``ubuntu@25.10`` base or newer.
+
+    For more information, see <LINK-TO-USRMERGE-EXPLANATION>.
```
medubelko, line 10:
```suggestion
    This reference is for the **Python (v2)** plugin, which is available for base 
    ``ubuntu@25.10`` or higher.
    
    For the legacy version, see :doc:`python_plugin`. For an explanation of why 
    there are different versions of this plugin, see <LINK-TO-USRMERGE-EXPLANATION>.
```

medubelko (approved):
What do you think about this text?

docs/reference/plugins/python_plugin.rst
bepri: I've tried to incorporate both feedbacks: how about this? https://github.com/canonical/rockcraft/pull/994/commits/8fe8018a803d72a6a6466ee799b878b4a852114d

docs/reference/plugins.rst
```diff
@@ -29,6 +29,7 @@ Rockcraft.
    /common/craft-parts/reference/plugins/npm_plugin
    plugins/poetry_plugin
    plugins/python_plugin
+   plugins/python_v2_plugin
```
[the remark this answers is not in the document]
bepri: I believe that the near-term is to keep the status quo with the .NET plugin, but I agree it doesn't look good.

I like your proposal as a potential start, but this is something I think should be discussed with the TAs and perhaps some other members of the team. I don't think it should be solved here, though.

PR #998, opened 2025-10-08
docs: add 'specify a base' how-to

medubelko:

---


docs/how-to/specify-a-base.rst
```diff
@@ -0,0 +1,58 @@
+.. _how-to-specify-a-base:
+
+Specify a base
+==============
+
+All rocks are layered on top of a software base, which determines compatibility with
```
tigarmo, line 6: what about `bare` rocks? They are built using a certain software base, but I would not say they are 'layered' on top of that. Layering has a very specific meaning for container images

docs/how-to/specify-a-base.rst
```diff
@@ -0,0 +1,58 @@
...
+----------------------------
+
+Specify an Ubuntu base with:
+
+.. code-block:: yaml
+    :caption: rockcraft.yaml
+
+    base: ubuntu@<version>
+
+For most rocks, select the latest Ubuntu LTS base, as your rock will benefit from that
```
tigarmo, line 25:
```suggestion
For most Ubuntu-based rocks, select the latest Ubuntu LTS base, as your rock will benefit from that
```

This is to emphasize that this sentence refers specifically to ubuntu rocks. "For most rocks" makes me think it includes bare-based too (even though this is a specific section)

docs/how-to/specify-a-base.rst
```diff
@@ -0,0 +1,58 @@
...
+release's 10-year support window.
+
+If instead you need access to the features and packages of an interim Ubuntu base,
+declare the interim release.
+
+
+Unmaintained bases
```
tigarmo, line 32: I think this section should come after "Bare base". The rationale is that the introduction talks about how rocks can be either 'ubuntu based' or have a 'bare base', so keeping the two sections together makes sense (to me)

docs/how-to/specify-a-base.rst
```diff
@@ -0,0 +1,58 @@
...
+------------------
+
+Over time, a rock's base will reach its end-of-life. It could be that the rock was
+short-lived, meant for testing software during an interim base's comparatively short
+nine-month lifecycle, or that it was extremely long-lived, and persisted past the
+support window of its LTS base.
+
+If you try and build a rock on an unsupported base, even if it first built when the base
```
tigarmo, line 40:
```suggestion
If you try and build a rock on an unmaintained base, even if it first built when the base
```

docs/how-to/specify-a-base.rst
```diff
@@ -0,0 +1,58 @@
...
+was supported, Rockcraft will by default halt the build. If you know the risks of
+unsupported bases, you can force the build to continue with:
```
tigarmo, line 42:
```suggestion
unmaintained bases, you can force the build to continue with:
```

docs/how-to/specify-a-base.rst
```diff
@@ -0,0 +1,58 @@
...
+Bare base
+---------
+
+A bare base is a special base for when you need a very lightweight rock. Declare the
+bare base with:
+
+.. code-block:: yaml
+    :caption: snapcraft.yaml
+
+    base: bare
```
tigarmo, line 58: need to add a build-base here

docs/how-to/specify-a-base.rst
```diff
@@ -0,0 +1,58 @@
...
+
+.. code-block:: bash
+
+    rockcraft pack --ignore=unmaintained
+
+
+Bare base
+---------
+
+A bare base is a special base for when you need a very lightweight rock. Declare the
```
tigarmo, line 52: I think we can augment this a bit with two pieces of info:

a) bare bases have no base software so your rockcraft project must declare everything that the rock will need;
b) bare bases must have a build-base, which specifies the version of Ubuntu system that is used for the build itself.

tigarmo (changes requested):
thanks!

tigarmo (approved):
thanks!

jahn-junior (approved):
Looks good! A few comments but nothing blocking.

tigarmo (approved):
@medubelko looks good to me, thanks!

PR #1014, opened 2025-10-24
fix: convert build-for to a string before serializing to metadata

### LINKED ISSUE #992: Architectures field in .rock/metadata.yaml is serialized incorrectly
### Bug Description

The architectures field in the .rock/metadata.yaml under the rock metadata layer is being serialized incorrectly:
```
architecture: !!python/object/apply:craft_platforms._architectures.DebianArchitecture
- amd64
base: bare
base-digest: e7c9bd1ccd1407f5eaf349b83d9cce7fb4837b5b5abd9e35218bf408e4cab14b
created: '2025-10-06T13:31:44.633401+00:00'
name: moaaz-bld-rock1
summary: A summary of your Go application
title: moaaz-bld-rock1
version: '0.2'
```
I would expect the result to just be `architecture: amd64`

### To Reproduce

Run rockcraft pack

### Environment

Ubuntu 24.04 LTS, AMD64

### rockcraft.yaml

```yaml
name: moaaz-bld-rock1
base: bare
build-base: ubuntu@24.04
version: '0.2'
summary: A summary of your Go application
description: Rock with fake bld part
platforms:
    amd64:
parts:
  fake-bld-part:
    plugin: dump
    source: bld:starcraft-test
    source-channel: latest/edge
```

### Relevant log output

```shell
None
```

bepri:

---

The `build-for` field was not being passed appropriately to `yaml.dump`. Since `DebianArchitecture` inherits from `str`, linters did not complain that `DebianArchitecture != str`, but `yaml.dump` was selecting `DebianArchitecture` as the object type to serialize. This led to yaml tags being dumped in, instead of plain strings.

Fixes #992.
ROCKCRAFT-271

### DIFF
--- a/tests/unit/services/test_package.py
+++ b/tests/unit/services/test_package.py
@@ -220,9 +220,7 @@ def test_inner_pack(
     )
     image.set_environment.assert_called_once_with(project.environment)
     image.set_annotations.assert_called_once_with(annotations)
-    image.set_control_data.assert_called_once_with(
-        {**metadata, "architecture": "amd64"}
-    )
+    image.set_control_data.assert_called_once_with(metadata)
     image.set_media_type.assert_called_once_with()
     image.to_oci_archive.assert_called_once_with(
         tag=project.version, filename=f"{project.name}_{project.version}_test-rock.rock"


tests/unit/services/test_package.py
```diff
@@ -227,3 +228,36 @@ def test_inner_pack(
...
+    )
+    mock_inner_pack = mocker.patch.object(package, "_pack")
+
+    fake_services.get("package").pack(prime_dir=new_dir / "prime", dest=new_dir)
+
+    mock_inner_pack.assert_called_once()
+    call = mock_inner_pack.call_args
+    # Assert that a plain string was given as a "build_for"
+    # Incorrect behavior would include something like DebianArchitecture in this list
+    assert call.kwargs.get("build_for").__class__.mro() == [str, object]
```
tigarmo: Looking at [this code](https://github.com/canonical/rockcraft/blob/main/rockcraft/services/package.py#L181) now it feels 'dirty' that the dict that is returned by `Project.generate_metadata()` gets further modified to add the architecture. It feels like that method in Project should do the whole thing.

So my suggestion is that we add an `architecture` parameter to `Project.generate_metadata()`, and that method ensures that the parameter is correctly stringified. Then we can test by just looking at the returned dict, no mocks necessary.

(also why use `mro()`?)

bepri: > (also why use `mro()`?)

I didn't want to do `assert not isinstance` because it felt possible that it could accidentally become some other class in the unknowable future. `mro()` let me just check that it was a string and nothing else

Issue #1022, opened 2025-11-17
entrypoint-service rejects empty argument list in 1.15 even though this previously worked


bschimke95:
### Bug Description

Rockcraft 1.15 introduces a stricter validation on `entrypoint-service` that rejects services whose command has zero default arguments. This breaks Rocks that rely on a bare command invocation.
Previous versions allowed an empty argument list. The new behavior appears to be an unintended side-effect of recent validation changes rather than a design requirement (as discussed [here](https://matrix.to/#/!AveAluboPXIRinIodA:ubuntu.com/$qgJvq8VTHoK5xzIJXtmHn2-4Qd-cRJZ8qix4mpslKGQ?via=ubuntu.com&via=matrix.org)).
The validation now asserts [len(args) > 0](https://github.com/canonical/rockcraft/pull/939/files#diff-54f38854c7e2d7d0fb5ab8160a493b4cf8f6f7da496a149eeff2680ec7ae6dbbR263-R270), making [] invalid even though CMD is optional in OCI and earlier Rockcraft versions accepted it.

### To Reproduce

1. Use Rockcraft 1.15.
3. Define a Pebble service with a bare command and no arguments.
5. Set entrypoint-service to that service.
7. Run rockcraft pack.
9. Observe the validation error:
```
the Pebble service '<name>' has a command <binary> without default arguments and thus cannot be used as the entrypoint-service
```

### Environment

Rockcraft 1.15 (tested on Ubuntu 22.04)

### rockcraft.yaml

```yaml
name: metrics-server
summary: ROCK for the metrics-server Project.
description: >
  Metrics Server is a scalable, efficient source of container resource metrics
  for Kubernetes built-in autoscaling pipelines.
version: "0.7.2"
license: Apache-2.0

base: bare
build-base: ubuntu@22.04
platforms:
  amd64:
  arm64:

services:
  metrics-server:
    override: replace
    summary: "metrics-server service"
    startup: enabled
    command: "/metrics-server [ ]"
    on-failure: shutdown

entrypoint-service: metrics-server

parts:
  openssl:
    plugin: nil
    stage-packages:
      - openssl-fips-module-3 # a Pro only FIPS package for openssl
      - openssl # now automatically selected from fips-preview
  metrics-server:
    plugin: nil
    source: https://github.com/kubernetes-sigs/metrics-server.git
    source-type: git
    source-tag: v0.7.2
    source-depth: 1
    build-packages:
      - build-essential
    build-snaps:
[...]

PR #1032, opened 2025-11-20
fix: preserve previous entrypoint-service behaviour

alesancor1:

---

This PR removes the restriction for specifying default args in `entrypoint-service` command. The reason behind this change is detailed in [this comment](https://github.com/canonical/rockcraft/issues/1022#issuecomment-3551984739)

Closes https://github.com/canonical/rockcraft/issues/1022

### DIFF
--- a/rockcraft/utils.py
+++ b/rockcraft/utils.py
@@ -128,7 +128,7 @@ def parse_command(command: str) -> tuple[list[str], list[str] | None]:
     """Parse command using shlex and return the command and its arguments split in lists inside a tuple.
 
     :param command: the command string to be parsed
-    :return: Tuple containing the command and its arguments split in lists
+    :return: Tuple containing ([command], [args]) where args is None if no bracket syntax was used, or a list (possibly empty) if bracket syntax was used
 
     :raises ValueError: if command is invalid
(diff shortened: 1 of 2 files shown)


rockcraft/utils.py
```diff
@@ -144,12 +144,12 @@ def parse_command(command: str) -> tuple[list[str], list[str]]:
             if arg == "]":
                 in_brackets = False
                 continue
-            args.append(arg)
+            args.append(arg)  # type: ignore[union-attr]
```
tigarmo: Instead of disabling the linting I think we should instead add an assertion `assert args is not None` or similar
@lengau this is in line with what we discussed during the sprint right?

tests/unit/test_utils.py
```diff
@@ -63,14 +63,14 @@ def test_get_managed_environment_snap_channel_none(monkeypatch):
 @pytest.mark.parametrize(
     ("command", "expected"),
     [
-        ("echo foo", (["echo", "foo"], [])),
+        ("echo foo", (["echo", "foo"], None)),
```
tigarmo: can you please also add the following cases: `"echo"`, `"echo [ ]"`

tigarmo (approved):

PR #1061, opened 2026-01-05
docs: Update LXD/Docker warning block in 12-factor tutorials

erinecon:
During a UX session, a participant required the workaround in the 12-factor Flask tutorial to fix the connectivity issue between Docker and LXD. The experience had some tension that can be resolved with documentation changes. 

The following updates were made to the warning blocks in all 12-factor tutorials:
* The user needs to know the name of the network bridge before running the commands, so this instruction was moved earlier in the block.
* The code block contains four commands that all require the name of the network bridge, so the block now instructs the user to save the name of their network bridge to an environment variable before running the fix. The environment variable is now used in the code block instead of a placeholder variable.
* Added `sudo` to all commands (required to run the commands, and should be safe since the tutorial takes place within a Multipass VM).

Ideally, a code block that can dynamically update with the name of the user's network bridge would be best. However, this solution would require custom JS code, and I didn't want to make this RTD project more complex when an environment variable can achieve the same result.

---

- [X] I've followed the [contribution guidelines](https://github.com/canonical/rockcraft/blob/main/CONTRIBUTING.md).
- [X] I've signed the [CLA](http://www.ubuntu.com/legal/contributors/).
- [ ] I've successfully run `make lint && make test`.
- [X] I've added or updated any relevant documentation.

I ran `make lint-docs` locally and fixed the issues.

### DIFF
--- a/docs/explanation/cryptography.rst
+++ b/docs/explanation/cryptography.rst
@@ -30,3 +30,4 @@ process. Rockcraft only uses the internal, pre-packaged copy of skopeo for this 
 .. _Craft Application: https://canonical-craft-application.readthedocs-hosted.com/en/latest/
 .. _Cryptographic technology in Craft Application: https://canonical-craft-application.readthedocs-hosted.com/en/latest/explanation/cryptography/
 .. _umoci: https://umo.ci/
+.. _skopeo: https://github.com/containers/skopeo


docs/tutorial/express.rst
```diff
@@ -181,20 +181,28 @@ Pack the rock:
...
    <lxd-docker-connectivity-issue_>`_. If we see a
    networking issue such as "*A network related operation failed in a context
-   of no network access*" or ``Client.Timeout``, allow egress network traffic
-   to flow from the LXD managed bridge using:
+   of no network access*" or ``Client.Timeout``, we need to allow egress network
+   traffic to flow from the LXD managed bridge.
+
+   First, run ``lxc network list`` to show the existing LXD managed bridges. The
+   bridge will be named something like ``lxdbr0``. Save the name to an
+   environment variable:
```
jahn-junior: The first sentence leads me to believe I'll see multiple bridges, while the second sentence refers to a single bridge. Will the output only show one? If not, which one is "the bridge"?

Then, because I'm feeling _really_ nitpicky today...

> LXD managed bridges

I think this should be either "LXD-managed bridges" or "managed LXD bridges". The latter seems to have [precedent in the LXD docs](https://documentation.ubuntu.com/lxd/stable-5.21/howto/network_bridge_firewalld/). "Existing" also feels redundant. How would it show nonexistent ones?

```suggestion
   First, run ``lxc network list`` to show the managed LXD bridges. The
   bridge will be named something like ``lxdbr0``. Save the name to an
   environment variable:
```

PR #1108, opened 2026-02-25
docs: make package-slices.svg readable in dark mode

### LINKED ISSUE #840: Graphs in Chisel Documentation not very readable in dark mode
### Request type

Fix

### What needs to get done

The SVG showing the slicing contains black arrows on a transparent background. This means that when reading in dark mode, this is barely visible, thanks to the dark background of the page.
The graphic there should be made more readable in dark mode.

### Documentation location

https://documentation.ubuntu.com/rockcraft/en/latest/explanation/chisel

### Additional information

Here's a screenshot from Firefox 137.0b3: Image

gcomneno:
Fixes #840

The package slicing SVG used black strokes/fills on a transparent background, making arrows/text hard to see in dark mode. Use currentColor for black stroke/fill so the diagram follows the documentation theme.

### DIFF
--- a/docs/_static/package-slices.svg
+++ b/docs/_static/package-slices.svg
@@ -6,61 +6,61 @@
 <rect x="0" y="0" width="480" height="360" fill="none" />
 
 <g transform="translate(0, 0)">
-<rect x="10" y="10" width="420" height="150" stroke="currentColor" fill="#c0c0c0" />
+<rect x="10" y="10" width="420" height="150" stroke="#777777" fill="#c0c0c0" />
 
 <text x="220" y="35" font-family="Ubuntu" font-size="18"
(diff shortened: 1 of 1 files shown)


gcomneno:
Good catch — I've pushed an update that replaces `currentColor` with consistent neutral greys for both strokes and labels in `package-slices.svg`, so it should render clearly in both light and dark themes. The preview should now reflect the change.

gcomneno:
Noticed that one slow test job (jammy, amd64, 3.10) is failing, while the rest of the checks are passing. Since this PR only adjusts an SVG in the docs, this might be unrelated. Could this be rerun to confirm whether it's a flake?

asanvaq (approved):

PR #1110, opened 2026-02-25
docs: reuse LXD/Docker networking troubleshooting snippet

### LINKED ISSUE #598: in the common case of networking issues between LXD and Docker, the error's doc_url isn't useful
### Bug Description

When hitting the issue described here: https://documentation.ubuntu.com/lxd/en/latest/howto/network_bridge_firewalld/#prevent-connectivity-issues-with-lxd-and-docker, Rockcraft will raise an error similar to the following: 

```
A network related operation failed in a context of no network access.                                                                                                                                                                                       
Recommended resolution: Verify that the environment has internet connectivity; see https://craft-providers.readthedocs.io/ for further reference.  
```

But the given `doc_url` doesn't really have any useful information about how to fix this issue. It is too generic.

I understand networking issues can vary, but we should at least have a troubleshooting section as a "catch all", where this specific LXD vs Docker conflict is described as a potential solution to the error.

### To Reproduce

Start from a fresh installation, install Docker, setup LXD and then `rockcraft pack`.

### Environment

Any

### rockcraft.yaml

```shell
Any
```

### Relevant log output

```shell
A network related operation failed in a context of no network access.                                                                                                                                                                                       
Recommended resolution: Verify that the environment has internet connectivity; see https://craft-providers.readthedocs.io/ for further reference.
```

gcomneno:
Fixes #598

Move the LXD/Docker connectivity troubleshooting into a reusable RST snippet and include it across tutorials to avoid duplication and keep the guidance consistent.

### DIFF
--- a/docs/reuse/troubleshooting/lxd-docker-networking.rst
+++ b/docs/reuse/troubleshooting/lxd-docker-networking.rst
@@ -4,12 +4,24 @@
    <https://documentation.ubuntu.com/lxd/en/latest/howto/network_bridge_firewalld/#prevent-connectivity-issues-with-lxd-and-docker>`_. If we see a
    networking issue such as "*A network related operation failed in a context of no
    network access*" or ``Client.Timeout``, we need to allow egress network traffic
-   to flow from the managed LXD bridge:
+   to flow from the managed LXD bridge.
+
+   First, list the available networks:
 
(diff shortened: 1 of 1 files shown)


medubelko:
Pinging @erinecon as the primary author. :)

PR #1179, opened 2026-04-09
docs: update content on bare vs non-bare bases based on UR

asanvaq:
Recent UR feedback requested documentation updates to clearly explain that both bare and non-bare Rockcraft images are production ready and to provide guidance on the advantages and trade-offs of each, including security and user-friendliness. This PR updates docs/explanation/bases.rst and makes sure the following content is present:

-  both images will be production ready whilst one is more secure though less user friendly
-  non-bare base is more intended for development purposes, make chiseled recommended for production

---


docs/explanation/bases.rst
```diff
@@ -23,6 +23,7 @@ The built rock contains the software and utilities from the Ubuntu image.
 
 Setting an Ubuntu base for a rock is especially useful when the goal is to build a rock
 that can serve as a general-purpose environment, such as a development workspace.
+Ubuntu base rocks are also more user-friendly compared to bare base ones.
```
medubelko: I take it we mean to say they're user friendly because they contain a mostly complete Ubuntu environment, but saying "**also** more user-friendly" suggests that they're user friendly for a different reason. Should we connect the ideas together, or did you mean there was another reason?

docs/explanation/bases.rst
```diff
@@ -53,8 +54,10 @@ and some :ref:`additional metadata <what-sets-rocks-apart>`.
 The bare base is especially useful when the goal is to build a lean chiselled rock.
 Such rocks are typically preferred for production environments due to their optimized
 deployment efficiency and reduced attack surface. The combination of the bare base with
-:ref:`Chisel <explanation-chisel>` slices will result in a minimalist container image
-that meets the production environment's criteria while retaining its functionality.
+:ref:`Chisel <explanation-chisel>` slices will result in a minimalist, secure
+container image with enhanced security that meets the production environment's
+criteria while retaining its functionality. Compared to Ubuntu base rocks,
+bare base rocks are less user-friendly.
```
medubelko: Same question here.

tigarmo: I agree. Would something like this connect the two sentences in a good way? 

> Compared to Ubuntu base rocks, bare base rocks are less use-friendly because authors need to actively include *all* necessary software

medubelko: I think I would put it like this:

> Compared to Ubuntu base rocks, bare base rocks are less user-friendly because they lack many common packages that users depend on.

tigarmo: Semi-related but who is the "user" in this context? The rock author, or the person making use of the rock?

medubelko: I was going to ask the same question. :)

asanvaq: I would say that if we take Michael's suggestion, the user is the person making use of the rock.

docs/explanation/bases.rst
asanvaq: I removed it here and mention it only for bare bases. But no, there was no other reason.

asanvaq:
@medubelko @tigarmo I applied the suggestions, what do you think?

medubelko (approved):

tigarmo (approved):
I ag

PR #1191, opened 2026-04-19
docs: put run-the-app instructions in their own code block

### LINKED ISSUE #1145: docs: Running Flask, Django and FastAPI should be in a code block
### Check existing issues

- [x] I've verified that this request isn't described by any existing issues.

### Request

This concerns the Flask, Django and FastAPI tutorials. 

After the app is created, there are in-line instructions on how to run the app. This does not match the rest of the instructions and can lead to users potenitally missing it. 

These instructions should have their own code block and the text above it should be updated accordingly.

### Document location

https://documentation.ubuntu.com/rockcraft/latest/tutorial/flask/#tutorial-build-a-rock-for-a-flask-app 
https://documentation.ubuntu.com/rockcraft/latest/tutorial/django/#tutorial-build-a-rock-for-a-django-app 
https://documentation.ubuntu.com/rockcraft/latest/tutorial/fastapi/#tutorial-build-a-rock-for-a-fastapi-app

Pawansingh3889:
## What changed

The "run the app" instructions on the Flask, Django, and FastAPI tutorials were inlined in the prose as RST inline code. Every other command on the same pages uses a proper code block (either `literalinclude` or `.. code-block:: bash`), so the inlined versions read as inconsistent and, per [#1145](https://github.com/canonical/rockcraft/issues/1145), users were scrolling past them.

Each of the three tutorials now promotes the inlined command into its own `.. code-block:: bash` block, and the lead-in sentence has been tightened so it reads naturally before the block. The change shape is identical across all three files.

## Files touched

- `docs/tutorial/flask.rst`, line 52 area: `flask run -p 8000`.
- `docs/tutorial/django.rst`, line 53-54 area: `python manage.py runserver`. The two sentences ("Enter the directory. Then run the Django app...") were fused into one so the block flows from it cleanly.
- `docs/tutorial/fastapi.rst`, line 53-54 area: `fastapi dev app.py --port 8000`.

## Not touched

- The `docs/tutorial/code/<framework>/task.yaml` files keep the existing background `&` form plus the `retry` + `kill` logic, since those are what the spread tests actually use. The user-facing code block intentionally shows the foreground command without `&`, matching what a learner would type.

## Resolves

Closes #1145.


medubelko (approved):
LGTM, thanks @Pawansingh3889!

medubelko:
@erinecon Do we want Spread to test these commands?

erinecon:
> Do we want Spread to test these commands?

Hi @medubelko , the Spread tests currently test the apps in a similar way to the tutorial, but not in an identical way.

I'll use Flask as an example. To test the app we use a [similar command in task.yaml](https://github.com/canonical/rockcraft/blob/main/docs/tutorial/code/flask/task.yaml#L20) but the command is run in the background so Spread can move onto the next command. Then we need to kill the app, similar to the tutorial's Ctrl + C, but we do this [in a different way in task.yaml](https://github.com/canonical/rockcraft/blob/main/docs/tutorial/code/flask/task.yaml#L27).

All of that to say, the Spread tests do test that the app can run and is reachable, but in a slightly different way. I don't think we need to make any Spread test changes in this PR (although I haven't reviewed this PR in great detail).

medubelko:
> All of that to say, the Spread tests do test that the app can run and is reachable, but in a slightly different way. I don't think we need to make any Spread test changes in this PR (although I haven't reviewed this PR in great detail).

If there's uncertainty, this probably needs your review. :)

erinecon (approved):
Thanks so much! Upon a more thorough review, I agree with my initial assessment that the Spread tests don't require any updates based on this PR.

asanvaq (approved):
Thanks!

tigarmo (approved):

docs/tutorial/django.rst
```diff
@@ -56,12 +56,12 @@ Create a new project using ``django-admin``:
     :end-before: [docs:create-project-end]
     :dedent: 2
 
-Enter the ``~/django-tutorial/django_hello_world/`` directory.
-Then run the Django app to verify that it works:
+Enter the ``~/django-hello-world/django_hello_world/`` directory, then run
```
tigarmo:
```suggestion
Enter the ``~/django-tutorial/django_hello_world/`` directory, then run
```

docs/tutorial/django.rst
```diff
@@ -56,12 +56,12 @@ Create a new project using ``django-admin``:
...
+the Django app to verify that it works:
 
 .. code-block:: bash
 
-    python manage.py runserver
+   python manage.py runserver
```
tigarmo:
```suggestion
    python manage.py runserver
```

docs/tutorial/fastapi.rst
```diff
@@ -54,7 +54,7 @@ Run the FastAPI app to verify that it works:
 
 .. code-block:: bash
 
-    fastapi dev app.py --port 8000
+   fastapi dev app.py --port 8000
```
tigarmo:
```suggestion
    fastapi dev app.py --port 8000
```

tigarmo (approved):

docs/tutorial/flask.rst
```diff
@@ -53,7 +53,7 @@ Run the Flask app to verify that it works:
 
 .. code-block:: bash
 
-    flask run -p 8000
+   flask run -p 8000
```
tigarmo:
```suggestion
    flask run -p 8000
```

PR #1205, opened 2026-05-05
ci: make workflows conditional, add docs-only spread tests

medubelko:
Make workflows conditional:

- The full lint job only runs if source files are changed.
- The Spread and test workflows only run if source files are changed.
- The Spread workflow only runs the `test/**` Spread tests.

Add a docs Spread workflow that runs the `docs/**` Spread tests.

With the QA workflow, if there's a change to the source, all linters run in one runner (`make lint`). But for the Spread tests, there's a hard split between the docs and source Spread tests, meaning each will spawn their own runner. This isn't bad, but it adds complexity and inconsistency between the different workflow types. If Starflow's workflows were completely split, we could properly partition the checks in both collections (QA source, QA docs, Spread source, Spread docs).

---

### DIFF
--- a/.github/workflows/spread-docs.yaml
+++ b/.github/workflows/spread-docs.yaml
@@ -1,0 +1,80 @@
+name: Spread tests for docs
+on:
+  pull_request:
+    paths:
+      - spread.yaml
+      - docs/**/code/**
+  push:
+    branches:
+      - main
+    paths:
+      - spread.yaml
+      - docs/**/code/**
+
+jobs:
+  snap-build:
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+      - name: Build snap
+        uses: canonical/action-build@v1
+        id: rockcraft
+      - name: Upload snap artifact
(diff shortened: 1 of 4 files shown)


.github/workflows/spread-docs.yaml
```diff
@@ -0,0 +1,80 @@
+name: Spread tests for docs
+on:
+  pull_request:
+    paths:
+      - spread.yaml
+      - docs/**/code/**
+  push:
+    branches:
+      - main
```
medubelko, line 9: Should I schedule this one for main?

.github/workflows/spread-docs.yaml
```diff
@@ -0,0 +1,80 @@
...
+        uses: canonical/action-build@v1
+        id: rockcraft
+      - name: Upload snap artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: snap
+          path: ${{ steps.rockcraft.outputs.snap }}
+
+  docs-tests:
+    runs-on: ubuntu-24.04
```
medubelko, line 32: Should this be one of the other runners so it's bypassed in forks? Once this reaches the Spread tests this will try and fail, which triggers notifications for contributors.

PR #1206, opened 2026-05-05
ci: make workflows conditional, add docs-only spread tests

medubelko:
Make workflows and jobs conditional:

- Full QA / lint runs if source files change.
- QA / Test runs if source files change.
- The Spread workflow calls the `test/**` Spread tests if source files change.

Add a Docs Spread workflow that runs the `docs/**` Spread tests if docs files change.

With the QA workflow, if there's a change to the source, all linters run in one runner (`make lint`). But for the Spread tests, there's a hard split between the docs and source Spread tests, meaning each will spawn their own runner. This isn't bad, but it adds complexity and inconsistency between the different workflow types. If Starflow's QA workflow were completely split, we could properly partition the checks in both collections – QA source, QA docs, Spread source, Spread docs.

---

### DIFF
--- a/.github/workflows/spread-docs.yaml
+++ b/.github/workflows/spread-docs.yaml
@@ -4,15 +4,17 @@ on:
     paths:
       - .github/workflows/spread-docs.yaml
       - spread.yaml
+      - .readthedocs.yaml
       - docs/**/code/**
   push:
     branches:
       - main
     paths:
       - spread.yaml
+      - .readthedocs.yaml
       - docs/**/code/**
   schedule:
-    - cron: "0 2 * * 0"
+    - cron: "0 2 * * 0" # Run every Sunday at 02:00
 
 jobs:
   docs-snap-build:
--- a/.github/workflows/spread-docs.yaml
+++ b/.github/workflows/spread-docs.yaml
@@ -15,7 +15,7 @@ on:
     - cron: "0 2 * * 0"
 
 jobs:
-  snap-build:
+  docs-snap-build:
     runs-on: ubuntu-24.04
     steps:
       - name: Checkout code
@@ -31,7 +31,7 @@ jobs:
           name: snap
           path: ${{ steps.rockcraft.outputs.snap }}
 
-  docs-tests:
+  docs-spread-tests:
     runs-on: [self-hosted, spread-installed]
     needs: [snap-build]
     timeout-minutes: 90
--- a/.github/workflows/spread-docs.yaml
+++ b/.github/workflows/spread-docs.yaml
@@ -11,7 +11,7 @@ on:
       - spread.yaml
       - docs/**/code/**
   schedule:
-    - cron: "0 0 */2 * *"
+    - cron: "0 2 * * 0"
 
 jobs:
   snap-build:
--- a/.github/workflows/spread-docs.yaml
+++ b/.github/workflows/spread-docs.yaml
@@ -28,7 +28,7 @@ jobs:
         uses: canonical/action-build@v1
         id: rockcraft
       - name: Upload snap artifact
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v7
         with:
           name: snap
           path: ${{ steps.rockcraft.outputs.snap }}
@@ -45,13 +45,13 @@ jobs:
           mkdir "${{ github.workspace }}"
 
       - name: Checkout rockcraft
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
         with:
           fetch-depth: 0
           submodules: true
 
       - name: Download snap artifact
-        uses: actions/download-artifact@v6
+        uses: actions/download-artifact@v8
         with:
           name: snap
           path: tests
--- a/.github/workflows/spread-docs.yaml
+++ b/.github/workflows/spread-docs.yaml
@@ -41,11 +41,6 @@ jobs:
           rm -rf "${{ github.workspace }}"
           mkdir "${{ github.workspace }}"
 
-      - name: Fix network issues related to docker and lxd
-        run: |
-          sudo iptables -P FORWARD ACCEPT
-          sudo ip6tables -P FORWARD ACCEPT
-
       - name: Checkout rockcraft
         uses: actions/checkout@v5
         with:
--- a/.github/workflows/spread-docs.yaml
+++ b/.github/workflows/spread-docs.yaml
@@ -10,6 +10,8 @@ on:
     paths:
       - spread.yaml
       - docs/**/code/**
+  schedule:
+    - cron: "0 0 */2 * *"
 
 jobs:
   snap-build:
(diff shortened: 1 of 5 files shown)


.github/workflows/spread-docs.yaml
```diff
@@ -0,0 +1,80 @@
+name: Spread tests for docs
+on:
+  pull_request:
+    paths:
+      - spread.yaml
+      - docs/**/code/**
+  push:
+    branches:
+      - main
```
medubelko, line 11: Should I schedule this for weekly runs?

.github/workflows/spread-docs.yaml
```diff
@@ -0,0 +1,80 @@
...
+        uses: canonical/action-build@v1
+        id: rockcraft
+      - name: Upload snap artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: snap
+          path: ${{ steps.rockcraft.outputs.snap }}
+
+  docs-tests:
+    runs-on: ubuntu-24.04
```
medubelko: Just to confirm, this is fine being a non-labelled runner?

tigarmo: It needs to run on a host that has spread installed
`runs-on: [self-hosted, spread-installed]`

.github/workflows/spread-docs.yaml
```diff
@@ -0,0 +1,80 @@
...
+
+jobs:
+  snap-build:
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+      - name: Build snap
```
tigarmo, line 27: I suppose that as an optimization we could have an step before this that disables building the wheels from [source](https://github.com/canonical/rockcraft/blob/main/snap/snapcraft.yaml#L141). It would speed up the snap build considerably and should be ok for doc tests

But it would also be fine to do this at a later date.

.github/workflows/spread-docs.yaml
```diff
@@ -0,0 +1,80 @@
+name: Spread tests for docs
+on:
+  pull_request:
+    paths:
+      - spread.yaml
```
tigarmo, line 6: I think you should do a spurious change on `spread.yaml` to trigger this workflow on this PR

.github/workflows/spread-docs.yaml
tigarmo: sure

.github/workflows/spread-docs.yaml
```diff
@@ -0,0 +1,80 @@
...
+        uses: canonical/setup-lxd@d492474f6c8ceed1b716bb34c8f575a428871c70
+
+      - name: Set up Go
+        uses: actions/setup-go@v6
+        with:
+          go-version: 1.26
+
+      - name: Install spread
+        run: |
+          go install github.com/canonical/spread/cmd/spread@latest
```
tigarmo: when you fix the runner declaration you don't need these steps anymore

.github/workflows/spread-docs.yaml
```diff
@@ -0,0 +1,80 @@
...
+        with:
+          fetch-depth: 0
+          submodules: true
+
+      - name: Download snap artifact
+        uses: actions/download-artifact@v6
+        with:
+          name: snap
+
+      - name: Install LXD
```
tigarmo: why do we need lxd here?

.github/workflows/spread-docs.yaml
medubelko: Let's table that for later.

.github/workflows/spread-docs.yaml
medubelko: f4eb7b3

.github/workflows/spread-docs.yaml
medubelko: Done. 8e18629

I was confused about why the Charmcraft workflow installed Spread manually.

.github/workflows/spread-docs.yaml
medubelko: Leftover from another test, oops. 8e18629

.github/workflows/spread-docs.yaml
medubelko: 87529c8 8e18629

.github/workflows/spread-docs.yaml
medubelko, line 1: @lengau Would there be a performance impact (good or bad) if I used a matrix to partition each Spread test into a different runner?

medubelko:
Both spread tests run. The docs one fails predictably.

.github/workflows/spread-docs.yaml
```diff
@@ -0,0 +1,66 @@
+name: Spread tests for docs
+on:
+  pull_request:
+    paths:
```
lengau, line 4:
```suggestion
    paths:
      - .github/workflows/spread-docs.yaml
```

Changes to this file will trigger its run too.

.github/workflows/spread-docs.yaml
lengau: Overall negative impact since it would need to spin up a machine for each test rather than running multiple tests on the same machine. Breaking it up by suite would probably be the safest way to go.

lengau (approved):

.github/workflows/spread-docs.yaml
medubelko: @tigarmo How would you feel about slicing by doc Spread suites here? I noticed we didn't do it in the original Spread workflow.

tigarmo: Well, we have to be careful with the split because we are ultimately constrained by the number of runners we can run simultaneously. So if you have (say) 5 runners, the best way to use them is the tell spread to run all tests and let it schedule the tests between the runners. In that sense, I don't immediately see how splitting the runs per-suite would help unless we *also* bump the number of runners.

I think we should discuss this separately.

.github/workflows/spread-docs.yaml
```diff
@@ -0,0 +1,67 @@
...
+    branches:
+      - main
+    paths:
+      - spread.yaml
+      - docs/**/code/**
+  schedule:
+    - cron: "0 2 * * 0"
+
+jobs:
+  snap-build:
```
tigarmo: Looking at the runs on the PR we have two "required snap-build" workflows:
Image

I think this is because this job here is also called "snap-build". I think we should rename it so we have finer-grained control on the required overflows: doc-only PRs will have (say) `docs-snap-build` and `docs-spread-test` as required, and "full" PRs will require the existing workflows.

tigarmo (changes requested):

.github/workflows/spread-docs.yaml
medubelko: Good catch. I didn't notice that GH pools all the job names together. :no_mouth: 9db2de7

medubelko: Docs spread is no longer required: 
image

tigarmo (approved):

PR #1207, opened 2026-05-14
style: replace mypy and pyright with ty

bepri:
Move to only using ty for linting. The targets to stop setting up mypy and pyright will come in from Starbase once all of our repos are ready.

Best for per-commit reviewing.

---

- [ ] I've followed the [contribution guidelines](https://github.com/canonical/rockcraft/blob/main/CONTRIBUTING.md).
- [ ] I've signed the [CLA](http://www.ubuntu.com/legal/contributors/).
- [ ] I've successfully run `make lint && make test`.


rockcraft/models/project.py
```diff
@@ -41,11 +43,13 @@
...
 
-# pyright workaround
-if TYPE_CHECKING:
-    _RunUser = str | None
-else:
-    _RunUser = Literal[tuple(SUPPORTED_GLOBAL_USERNAMES)] | None
+
+class _RunUser(str):
+    @classmethod
+    def __get_pydantic_core_schema__(
```
bepri, line 49: This effectively is a run-time type definition, as opposed to a static `Literal` definition. As of this PR, this basically just becomes:

```
_RunUser = Literal["_daemon_"]
```

This is necessary so that we can just list the supported names in that one mapping, rather than enumerating them both in a `Literal` definition and in the mapping.

[Relevant docs](https://pydantic.dev/docs/validation/latest/concepts/types/#customizing-validation-with-__get_pydantic_core_schema__), for the curious

rockcraft/oci.py
```diff
@@ -291,7 +291,10 @@ def add_user(
                 groupf.write(user_files["group"])
 
             if user_files["shadow"]:
-                days_since_epoch = (datetime.utcnow() - datetime(1970, 1, 1)).days  # type: ignore[reportDeprecated]
+                days_since_epoch = (
+                    datetime.now(timezone.utc)
+                    - datetime(1970, 1, 1, tzinfo=timezone.utc)
```
bepri, line 296: I tested this by comparing the output of these two lines:
```
datetime.now(timezone.utc) - datetime(1970, 1, 1, tzinfo=timezone.utc)
# against
datetime.utcnow() - datetime(1970, 1, 1)
```
It seems neither line gives a deterministic number (varies by ~1-2 seconds each run), but the two lines always generate a result within ~1-2 seconds of each other, so they should be identical.

In an attempt to harden this, I left `test_oci.py` to use timezone-unaware time objects to mix the two formats, and this still works.

(fwiw, the test was never correct because `utcnow() - datetime(year, month, day)` is wrong because it's comparing a timezone-offset time object against a "raw" time object)

tests/spread/rockcraft/extension-django/example_django/example_django/asgi.py
bepri, line 1: This directory (`tests/spread`) is ignored in the config moving forward, but I still didn't like type ignores hanging around for a checker that we didn't even use

mr-cal (approved):

lengau (approved):

rockcraft/services/service_factory.py
```diff
@@ -39,7 +39,7 @@ class RockcraftServiceFactory(ServiceFactory):
     if TYPE_CHECKING:
         from rockcraft.services import RockcraftImageService
 
-        image: RockcraftImageService = None  # type: ignore[assignment]
+        image: RockcraftImageService = None  # ty: ignore[invalid-assignment]
```
tigarmo: should we drop the `= None` here? Like craft-app does: https://github.com/canonical/craft-application/blob/main/craft_application/services/service_factory.py#L80

tests/integration/plugins/test_python_plugin.py
```diff
@@ -61,7 +63,7 @@ def create_python_project(base: str, extra_part_props=None, *, plugin: str) -> P
 
     build_base = None
     if base != "bare":
-        base_alias = get_base_alias(base_name=tuple(base.split("@")))  # pyright: ignore[reportCallIssue,reportArgumentType]
+        base_alias = get_base_alias(base_name=tuple(base.split("@")))  # ty: ignore[no-matching-overload] ty does not understand the output of splitting these is predictable
```
tigarmo: do you want to split this into
```python
name, version = base.split("@")
base_alias = get_base_alias(base_name=(name, version))
```
to drop the annotation?

tigarmo (approved):
thanks for sensibly splitting the commits, made the review much easier

tests/integration/plugins/test_python_plugin.py
bepri: It still complained about not being a `Literal["ubuntu"]` for the name, but an extra import fixed that. Thanks!

rockcraft/services/service_factory.py
bepri: Ooo, didn't know I could. Done

PR #1214, opened 2026-05-26
fix: use fips channel when fips services enabled

smethnani:
Uses FIPS channel for the pebble snap when FIPS pro services are enabled

(ROCKCRAFT-306)

---


lengau:
Does it matter that the build still won't be happening with a FIPS-enabled image/kernel?

lengau (approved):

rockcraft/pebble.py
```diff
@@ -496,7 +498,17 @@ def add_pebble_part(project: dict[str, Any]) -> None:
         # don't try to add the relevant pebble part if base is missing
         # (friendly error message will be presented later from pydantic)
         return
-    pebble_part = Pebble.get_part_spec(build_base)
+    pebble_part = deepcopy(Pebble.get_part_spec(build_base))
+
+    if ProServices._pro_client_exists() and (  # noqa: SLF001 # pyright: ignore[reportPrivateUsage]
```
tigarmo: If we're going to use craft-app here I think the right thing to do is make those two methods public
otherwise we're going to eventually change that API in craft-application and break rockcraft

tests/spread/pro/fips-pebble/task.yaml
tigarmo, line 1: Image

tests/spread/pro/fips-pebble/task.yaml
```diff
@@ -0,0 +1,19 @@
...
+  UNPACK_DIR="unpacked"
+
+  rockcraft pack --pro=fips
+
+  mkdir -p $IMAGE_DIR
+  tar -xvf $ROCK_FILE -C $IMAGE_DIR
+
+  /snap/rockcraft/current/bin/umoci unpack --rootless --image $IMAGE_DIR:$TAG $UNPACK_DIR
+
+  # verify pebble version
```
tigarmo: the test is on the right track but I think we should run pebble in the container itself to ensure its execution in the target environment (the rock) is working

something like `docker run --rm -it fips-pebble-test:latest exec pebble -v | MATCH ...` after loading the rock

tests/spread/pro/fips-pebble/task.yaml
```diff
@@ -0,0 +1,19 @@
...
+  $UNPACK_DIR/rootfs/bin/pebble version | MATCH "\-fips"
+
+  rockcraft clean
```
tigarmo: this line isn't necessary (we [clean](https://github.com/canonical/rockcraft/blob/main/spread.yaml#L103) the build instances after every test)

tests/spread/pro/fips-pebble/task.yaml
```diff
@@ -0,0 +1,19 @@
...
+  rm -rf $ROCK_FILE $IMAGE_DIR $UNPACK_DIR
```
tigarmo: this line should be in a `restore:` declaration, to run even if the test fails

tests/unit/test_pebble.py
```diff
@@ -515,3 +515,26 @@ def test_project_unmarshal_existing_pebble_same():
...
+        ({"esm-apps", "fips-preview"}, "pebble/fips/stable"),
+        ({"esm-apps"}, "pebble/latest/stable"),
+    ],
+    ids=["no_pro_services", "fips", "fips_updates", "mixed", "not_fips"],
+)
+def test_add_pebble_part_snap(mocker, enabled_services, expected_channel):
+    """Test that FIPS pebble snap is used when FIPS services are enabled."""
+
+    mocker.patch(
+        "rockcraft.pebble.ProServices._get_pro_services", return_value=enabled_services
```
tigarmo:
```suggestion
        "craft_application.util.ProServices._get_pro_services", return_value=enabled_services
```

tigarmo (changes requested):

smethnani:
Linting fixed in #1215

rockcraft/pebble.py
smethnani: craft-application [PR here](https://github.com/canonical/craft-application/pull/1083)

tests/spread/pro/fips-pebble/task.yaml
```diff
@@ -0,0 +1,13 @@
+summary: Test packing rock with FIPS pebble snap
+prepare: |
+  sudo pro enable --assume-yes fips-updates
+execute: |
+  ROCK_FILE="fips-pebble-test_latest_amd64.rock"
+
+  rockcraft pack --destructive --pro=fips-updates
+  sudo rockcraft.skopeo --insecure-policy copy oci-archive:$ROCK_FILE docker-daemon:fips-pebble-test:latest
+
+  docker run --rm fips-pebble-test:latest exec pebble version | MATCH "\-fips"
```
tigarmo, line 8: 👍🏾

rockcraft/pebble.py
tigarmo: thanks! Approving the PR but let's leave this comment up until this particular part is fixed

tests/spread/pro/fips-pebble/task.yaml
```diff
@@ -1,19 +1,13 @@
 summary: Test packing rock with FIPS pebble snap
+prepare: |
+  sudo pro enable --assume-yes fips-updates
```
tigarmo: wait why is this necessary?

smethnani: Without it, the spread test fails with the `failed to enable fips` error.

lengau: Ohhh because we're running in destructive mode.

tigarmo: > Ohhh because we're running in destructive mode.

ah I see; why are we packing in destructive mode? I'd rather test 'regular' mode

tigarmo (approved):

PR #1246, opened 2026-06-08
docs: add metadata to all remaining pages

asanvaq:
This PR adds metadata that follows the [Starflow guidelines](https://documentation.ubuntu.com/starflow/latest/how-to/add-a-page-meta-description/#write-a-meta-description).
Pages not included are part of other PRs currently under review or incoming.

---

- [x] I've followed the [contribution guidelines](https://github.com/canonical/rockcraft/blob/main/CONTRIBUTING.md).
- [x] I've signed the [CLA](http://www.ubuntu.com/legal/contributors/).
- [x] I've successfully run `make lint && make test`.
- [ ] I've added or updated any relevant documentation.
- [x] In documents I changed, I [added a meta description](https://canonical-starflow.readthedocs-hosted.com/how-to/add-a-page-meta-description/) if one was missing.
- [ ] I've updated the relevant release notes.

PR #1253, opened 2026-06-16
docs: Incorporate user feedback in the 12-factor reference pages

erinecon:
This PR updates the 12-factor reference pages to incorporate the following feedback received during UXR sessions:

* Update the headers so that they’re human-readable and indicate the purpose or goal of the section
* Clarify that the `parts` snippets are technically optional, but depending on the complexity of the project, could be required. Clarify when the user needs to add the snippet versus when they can safely ignore the snippet
* Clarify the parts snippets that include or exclude files. Add this type of snippet overrides the default, and you exclude a file by including the snippet in your project file without the file. Add a reference for more complex ways of including or excluding files (e.g., with glob patterns)
* Add more links (start script in the Express framework page, bare base, rocks, etc.)
* Add captions to all code snippets so it’s clear that they touch the rockcraft.yaml project file.
* Double check and clarify any undefined acronyms (e.g. WSGI in the Flask reference page).

I successfully ran `make docs-lint` before opening the PR.

I also did not add meta descriptions to avoid conflicts with https://github.com/canonical/rockcraft/pull/1246

---

- [x] I've added or updated any relevant documentation.


docs/reference/extensions/django-framework.rst
```diff
@@ -9,7 +9,8 @@ rocks.
 It facilitates the installation of Django application dependencies, including
 Gunicorn, inside the rock. Additionally, it transfers your project files to
 ``/django/app`` within the rock.
-By default, the base ``bare`` is used to generate a lightweight image.
+By default, the :ref:`base <explanation-bases>` ``bare`` is used to generate a
```
medubelko: I think it makes more sense grammatically to say "`bare` base".

What's your take on the value of the added links? Were subjects confused about the terms?

docs/reference/extensions/django-framework.rst
```diff
@@ -55,6 +56,8 @@ application. In the following example we use it to specify ``libpq-dev``:
          # list required packages or slices for your Django application below.
          - libpq-dev
 
+Adding this key is optional unless your application requires additional dependencies.
```
medubelko: Could we combine this with the earlier paragraph? We start with an instruction, then walk it back to a conditional.

docs/reference/extensions/fastapi-framework.rst
```diff
@@ -24,8 +25,9 @@ There are 2 requirements to be able to use the ``fastapi-framework`` extension:
 
 1. There must be a ``requirements.txt`` file in the root of the project with
    ``fastapi`` declared as a dependency
-2. The project must include a ASGI app in a variable called ``app`` in one of
-   the following files relative to the project root (in order of priority):
+2. The project must include an `ASGI <https://asgi.readthedocs.io/en/latest/>`_ app
```
medubelko: If we're expanding WSGI, we should do the same for ASGI.

docs/reference/extensions/fastapi-framework.rst
```diff
@@ -24,8 +25,9 @@ There are 2 requirements to be able to use the ``fastapi-framework`` extension:
...
+   in a variable called ``app`` in one of the following files relative to the project
+   root (in order of priority):
```
medubelko: Out of scope for this PR, but IMO "precedence" fits better than "priority"

docs/reference/extensions/fastapi-framework.rst
```diff
@@ -71,10 +79,12 @@ example:
         - app/templates
         - app/static
 
-Some files, if they exist, are included by default. These include:
-``app``, ``src``, ``<rock name>``, ``app.py``, ``migrate``, ``migrate.sh``,
-``migrate.py``, ``static``, ``templates``.
+You can use glob patterns to define your list of files; see :ref:`filesets_explanation`
+for more details.
```
medubelko: - I'd rather we avoid semicolons.
- Can we be a bit more helpful than "more details"? What useful thing will the user [find at the other end of the link](https://documentation.ubuntu.com/starflow/latest/how-to/starcraft-style-guide/#links)?

docs/reference/extensions/fastapi-framework.rst
```diff
@@ -71,10 +79,12 @@ example:
...
 
+Adding the ``prime`` field to your project file overrides the default files to be included.
```
medubelko:
```suggestion
Adding the ``prime`` key to your project file overrides the default files to be included.
```

docs/reference/extensions/fastapi-framework.rst
medubelko:
```suggestion
2. The project must include an `ASGI <https://asgi.readthedocs.io/en/latest/>`__ app
```

docs/reference/extensions/go-framework.rst
```diff
@@ -23,14 +24,14 @@ in the root directory of the project.
...
--------------------------------------------------------
+Specifying a different binary for your app
+------------------------------------------
 
 If the main package is in the base directory and the rockcraft name
 attribute is equal to the go module name, the name of the binary will
 be selected correctly, otherwise you will need to adjust it.
 
-You can use this field to specify a different binary to be used as the
+Use the ``organize`` field to specify a different binary to be used as the
```
medubelko:
```suggestion
Use the ``organize`` key to specify a different binary to be used as the
```

docs/reference/extensions/django-framework.rst
```diff
@@ -40,10 +41,10 @@ For the project to make use of asynchronous Gunicorn workers:
 
 .. _reference-django-framework-stage-packages:
 
-``parts`` > ``django-framework/dependencies:`` > ``stage-packages``
--------------------------------------------------------------------
+Specifying required dependencies
```
medubelko: Sounds a bit like a procedure to me. Could we nominalise these new headings, so they're less action-oriented?

docs/reference/extensions/django-framework.rst
erinecon: Yes, nearly everyone I talked to about the reference pages said "I don't know what a base is" or "I don't know what bare means". A couple people explicitly requested a link to read more about bases. The value of this particular link is that it provides an immediate path toward clarity and understanding if someone isn't familiar with bases.

Admittedly I was potentially dropping these users into the middle of the user journey; I don't expect that this will be the first page that they will read, but it was in the UXR sessions. On the other hand, it is possible that this will be the first page that a developer could read (for example: it's the second Rockcraft docs link after the tutorial when you search for "django support rockcraft"). 

However, I'm not opposed to moving this link down to the "Useful links" section if you have concerns over including this link so early in the page. Then I can test the page with more users to see whether the initial concerns have been addressed.

Another option is that we move this information lower in the page -- part of the feedback was that this information appears early in the page and immediately introduces a point of question or confusion.

WDYT?

docs/reference/extensions/django-framework.rst
erinecon: Absolutely, I can move this line into the earlier paragraph!

docs/reference/extensions/fastapi-framework.rst
erinecon: Sure, can do. Thanks!

docs/reference/extensions/fastapi-framework.rst
erinecon: Since we're here and it's a relatively minor comment, I don't mind making the change

docs/reference/extensions/fastapi-framework.rst
erinecon: Ah right, thanks for the reminder about semicolons! I'll also expand a bit on what users will find.

docs/reference/extensions/django-framework.rst
erinecon: For more context, I was trying to avoid action-oriented headers by using gerunds. My thought process was that a heading like "Specify required dependencies" would fit more with a how-to guide -- it's more authoritative, more action-y, more procedural. While "Specifying required dependencies" is more conceptual, more about the abstract topic rather than the action itself.

WDYT?

docs/reference/extensions/django-framework.rst
asanvaq: We are running into a similar problem in the home page. We were trying to find what to call the domain of concern related to base. Based on the user feedback, I think it would be interesting to explore briefly defining with the same term what a base is.

What do you think of  "system foundation, or base, is set as bare..."?

medubelko: I like Anna's approach. I prefer sprinkling in light explanation so we don't need to send the user on a journey elsewhere. The user should stay on the tutorial page. They're already splitting their attention between it and their terminal/IDE. Links risk more browser tabs and distractions.

We recently added a succinct definition as a meta description: https://github.com/canonical/rockcraft/blob/f452750648b1b56730bba296a1c6d14b61a6b4c1/docs/explanation/bases.rst?plain=1#L2

docs/reference/extensions/django-framework.rst
medubelko: IMO it still implies action, because of the verb root. For a nominal, I was thinking:

```rst
App dependencies
```

Then, in the section, we'd make the machinery the subject:

```rst
The ``stage-packages`` key specifies all additional dependencies. If the Django app 
has its own special dependencies, this key must declare them.

In the following example...
```

This is a bikeshedding suggestion, though, because it would require rejiggering the whole reference to be about the machine, not the user. Feel free to ignore. ;)

medubelko (approved):
Some followup discussion for consideration, but nothing blocking. Approving.

docs/reference/extensions/django-framework.rst
erinecon: Sure, I'm willing to try another approach instead of adding a link in the reference pages. I'll opt for Anna's approach, thanks for the discussion and feedback!

docs/reference/extensions/django-framework.rst
erinecon: Well the biggest pain point with the reference pages are the headers, so I'm willing to play around  with them for a while longer if it means we're adequately addressing the concerns. Let me give it a try

erinecon:
Hi @medubelko and @asanvaq , thanks for your reviews and feedback. I have updated the PR to resolve the remaining open discussions. Let me know if there's anything else that catches your eyes!

medubelko:
@erinecon I think this is good to go. Happy to merge when you're ready.

asanvaq (approved):
Looks great, thank you so much for the improvements!

PR #1259, opened 2026-06-17
feat(extensions): extension dispatch for 12-factor extensions.

Thanhphan1147:
Add V2 extension classes and factory dispatch for all 12-factor frameworks (Flask, Django, FastAPI, ExpressJS, SpringBoot).

## What this PR does
- `get_project_bases(yaml_data)`: extracts base strings from project YAML
- `_FrameworkFactory`: routes between V1/V2 extension classes based on the project's target bases
- `FlaskFrameworkV2` / `FlaskFrameworkFactory` — targets `ubuntu@26.04`, experimental
- `DjangoFrameworkV2` / `DjangoFrameworkFactory` — targets `ubuntu@26.04`, experimental
- `FastAPIFrameworkV2` / `FastAPIFrameworkFactory` — targets `ubuntu@26.04`, experimental
- `ExpressJSFrameworkV2` / `ExpressJSFrameworkFactory` — targets `ubuntu@26.04`, experimental
- `SpringBootFrameworkV2` / `SpringBootFrameworkFactory` — targets `ubuntu@26.04`, experimental

Each V2 class is behaviourally identical to its V1 counterpart for now; it exists so that a future paas-charm 2.0 implementation can be dispatched transparently based on the rock's target base.


Thanhphan1147:
Hi @lengau, this PR is now ready for review which sets the base for other framework-specific PRs for the extension dispatch. I've also aligned it with @alithethird 's PR on charmcraft. Can I ask you to take a look when you have the time? I'll mark the framework PRs ready for review once this is reviewed and merged.

rockcraft/extensions/extension.py
```diff
@@ -116,6 +128,32 @@ def validate(self, extension_name: str) -> None:
...
+    """Route to V1 or V2 extension based on project bases."""
+
+    def __init__(self, v1_cls: type[Extension], v2_cls: type[Extension]) -> None:
+        self._v1_cls = v1_cls
+        self._v2_cls = v2_cls
+
+    def __call__(self, *, project_root: Path, yaml_data: dict[str, Any]) -> Extension:
+        bases = get_project_bases(yaml_data)
+        if any(base in self._v2_cls.get_supported_bases() for base in bases):
+            return self._v2_cls(project_root=project_root, yaml_data=yaml_data)
```
Thanhphan1147: Update this to defaults to v2 if condition for dispatch doesn't match

rockcraft/extensions/__init__.py
```diff
@@ -32,11 +39,23 @@
     "register",
     "unregister",
     "gen_logging_part",
+    "DjangoFramework",
```
Thanhphan1147: Don't need these

rockcraft/extensions/extension.py
```diff
@@ -28,6 +28,25 @@
...
+
+    if base_str := yaml_data.get("base"):
+        bases.add(base_str)
+
+    if platforms := yaml_data.get("platforms", {}):
+        for label in platforms:
+            if "@" in label:
+                bases.add(label)
+
+    return bases
```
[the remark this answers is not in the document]
Thanhphan1147: updated this with the logic used in charmcraft

rockcraft/extensions/extension.py
Thanhphan1147: done

rockcraft/extensions/extension.py
```diff
@@ -23,11 +23,58 @@
...
 from typing import Any, final
 
+import craft_platforms
 from craft_cli import emit
 
 from rockcraft import errors
 
 
+# Similar logic in charmcraft but we return the set of base_str instead of a tuple
+def get_project_bases(yaml_data: dict[str, Any]) -> set[str]:
```
lengau: I don't believe we need this in Rockcraft since Rockcraft requires the `base` key. Charmcraft is the special snowflake here.

lengau (approved):
LGTM, but with the caveat about pulling bases from the `platforms` and `bases` keys.

rockcraft/extensions/extension.py
Thanhphan1147: Thanks, I removed the code to support  `platform` and `bases`

rockcraft/extensions/extension.py
```diff
@@ -23,11 +23,25 @@
...
```
tigarmo: Hrm currently the base can only ever be a single string; shouldn't this return `str | None` instead?

rockcraft/extensions/extension.py
```diff
@@ -23,11 +23,25 @@
...
+    """Extract and normalize all bases used in the project."""
+    bases: set[str] = set()
+    if base_str := yaml_data.get("base"):
```
tigarmo: Some comments about this:

1) I believe the 12f extensions in general promote the use of `bare` bases, like this:

```yaml
base: bare
build-base: ubuntu@26.04
```

In this common scenario, the base that you're interested in is the `build-base`. In general, the 'algorithm' to get a project's base is: try to get the `build-base`, if it fails get the `base` (which will then not be `bare` in a well-formed project).

rockcraft/extensions/extension.py
```diff
@@ -23,11 +23,25 @@
...
+        if parsed := craft_platforms.parse_base_and_name(base_str)[0]:
+            bases.add(f"{parsed.distribution}@{parsed.series}")
+        else:
+            name, _, channel = base_str.partition("@")
```
tigarmo: 2) I think it's useful to be a bit proactive with error handling here, since this code is (probably) called before the project has been validated. For example, if `base: invalid`:

```python
>>> "invalid".partition("@")
('invalid', '', '')
```

rockcraft/extensions/extension.py
```diff
@@ -116,6 +130,32 @@ def validate(self, extension_name: str) -> None:
...
+
+    def get_supported_bases(self) -> tuple[str, ...]:
+        return tuple(
+            dict.fromkeys(
+                self._v1_cls.get_supported_bases() + self._v2_cls.get_supported_bases()
+            )
+        )
+
+    def is_experimental(self, base: str | None) -> bool:
+        if base in self._v2_cls.get_supported_bases():
```
tigarmo: like in `__call__`, I think this method should actively check against `v1` bases, and then default to returning the `v2` class method

tigarmo (changes requested):
Some of the spread test failures are related to this change, please take a look.
Example:

```
2026-07-07T12:39:46.6790357Z 2026-07-07 12:39:46 Error executing google:ubuntu-22.04-64:tests/spread/rockcraft/extension-django:bare_2204 (jul071227-938563) : 
2026-07-07T12:39:46.6792673Z -----
2026-07-07T12:39:46.6793288Z + run_rockcraft init --name example-django --profile django-framework
2026-07-07T12:39:46.6794022Z Checking project directory.
2026-07-07T12:39:46.6794480Z Initializing project.
2026-07-07T12:39:46.6795349Z Rendering project.
2026-07-07T12:39:46.6796069Z Rendered project.
2026-07-07T12:39:46.6797401Z Go to https://documentation.ubuntu.com/rockcraft/latest/reference/extensions/django-framework to read more about the 'django-framework' profile.
2026-07-07T12:39:46.6798702Z Successfully initialised project.
2026-07-07T12:39:46.6799390Z + sed -i 's/^base: .*/base: bare/g' rockcraft.yaml
2026-07-07T12:39:46.6800229Z + sed -i 's/^build-base: .*/build-base: ubuntu@22.04/g' rockcraft.yaml
2026-07-07T12:39:46.6800822Z + cat
2026-07-07T12:39:46.6801209Z + run_rockcraft pack
2026-07-07T12:39:46.6801705Z Extension is experimental: 'django-framework'
2026-07-07T12:39:46.6803339Z Recommended resolution: Run with ROCKCRAFT_ENABLE_EXPERIMENTAL_EXTENSIONS=True to enable experimental extensions.
2026-07-07T12:39:46.6804886Z For more information, check out: https://documentation.ubuntu.com/rockcraft/latest/reference/extensions/
2026-07-07T12:39:46.6806146Z Full execution log: '/root/.local/state/rockcraft/log/rockcraft-20260707-123945.759460.log'
2026-07-07T12:39:46.6807198Z -----
```

rockcraft/extensions/extension.py
Thanhphan1147: Oops, forgot to change this as well, I updated this, thanks!

rockcraft/extensions/extension.py
Thanhphan1147: Thanks, I added the same validation as in craft_platforms.parse_base_and_name

rockcraft/extensions/extension.py
Thanhphan1147: I refactored the parsing logic here

rockcraft/extensions/extension.py
Thanhphan1147: Thanks, I updated this part.

rockcraft/extensions/extension.py
```diff
@@ -21,26 +21,31 @@
...
+    ):
         if parsed := craft_platforms.parse_base_and_name(base_str)[0]:
-            bases.add(f"{parsed.distribution}@{parsed.series}")
-        else:
-            name, _, channel = base_str.partition("@")
-            bases.add(f"{name}@{channel}")
+            return f"{parsed.distribution}@{parsed.series}"
+
+        if base_str.count("@") != 1:
+            raise ValueError(
```
tigarmo: I think raising an error here is the wrong approach, for two reasons:

1) The main way to validate the project is through the pydantic model that gets created a bit after this. This produces a consistent error message with references to the yaml, etc. Raising an error here bypasses this
2) `devel` is a valid `build-base`, so this error would be incorrect in that case anyway.

I think you should check for `@` but just return `None` if it's missing

tigarmo (changes requested):
Thanks. The spread tests look OK now (well, they are failing for reasons unrelated to this PR) but ruff is complaining about the code on the linting workflow.

rockcraft/extensions/extension.py
Thanhphan1147: Makes sense, I updated this part to simply returns none and handle `devel` as a base. As a side note, `craft_platforms.parse_base_and_name` is still raising `ValueError` if the base_str is not properly formatted, do you want me to also catch that error and return none?

tigarmo: yeap

tigarmo:
@Thanhphan1147 there are failing linters and unit tests

rockcraft/extensions/extension.py
Thanhphan1147: updated!

rockcraft/extensions/extension.py
```diff
@@ -21,13 +21,35 @@
...
+        if base_str == "devel":
+            return "ubuntu@devel"
+
+        if base_str.count("@") != 1:
+            return None
+
+        name, _, channel = base_str.partition("@")
+        return f"{name}@{channel}"
+
+    return None
```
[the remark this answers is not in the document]
Thanhphan1147: we want to get build-base and defaults back to `base` as mentioned here: https://github.com/canonical/rockcraft/pull/1259#discussion_r3543653932

tests/unit/extensions/test_fastapi.py
```diff
@@ -332,3 +332,130 @@ def test_fastapi_extension_incorrect_prime_prefix_error(tmp_path, fastapi_input_
...
+
+def test_factory_supported_bases():
+    """fastapi_framework_factory includes both V1 and V2 bases."""
+    factory_bases = extensions.FastAPIFrameworkFactory.get_supported_bases()
+    assert "ubuntu@26.04" in factory_bases
+    for base in extensions.FastAPIFramework.get_supported_bases():
+        assert base in factory_bases
+
+
+def test_FastAPIFrameworkFactory_is_experimental():
```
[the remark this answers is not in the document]
Thanhphan1147: updated

Thanhphan1147:
@tigarmo I realized that in trying to match what's implemented in charmcraft I overcomplicated things, so I refactor the `get_project_base` to simplify it given that rockcraft is more strict about what's accepted for `base` and `build-base`. I also fixed the lint and unit test failures

rockcraft/extensions/extension.py
```diff
@@ -116,6 +138,32 @@ def validate(self, extension_name: str) -> None:
...
+    def __init__(self, v1_cls: type[Extension], v2_cls: type[Extension]) -> None:
+        self._v1_cls = v1_cls
+        self._v2_cls = v2_cls
+
+    def __call__(self, *, project_root: Path, yaml_data: dict[str, Any]) -> Extension:
+        base = get_project_base(yaml_data)
+        if base in self._v1_cls.get_supported_bases():
+            return self._v1_cls(project_root=project_root, yaml_data=yaml_data)
+        return self._v2_cls(project_root=project_root, yaml_data=yaml_data)
+
```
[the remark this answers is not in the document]
Thanhphan1147: we want to default to V2 for non-v1-supported bases

tigarmo (approved):
thanks!

PR #1267, opened 2026-06-22
docs: Remove Chisel-related duplicated content

asanvaq:
Chiselling how-to section is outdated. Some of the pages contain Chisel-only content that is already covered in the Chisel docs themselves. As these pages are not related to Rockcraft and exist elsewhere.

Installing a custom slice was based on previous steps from the deleted guides. Instead of giving a specific package as an example, I added a placeholder that solves this problem.

I also updated the explanation page to focus on how Chisel relates to Rockcraft. 

---

- [x] In documents I changed, I [added a meta description](https://canonical-starflow.readthedocs-hosted.com/how-to/add-a-page-meta-description/) if one was missing.


docs/how-to/chiselling/install-slice.rst
```diff
@@ -13,10 +13,10 @@ likely end up creating your own slice definition.
 Once you have it though, the most obvious question is: **how can I install
 this custom slice with Chisel?**
 
-Let's assume you want to install the OpenSSL binaries slice created in
-:doc:`this guide <create-slice>`.
+Let's assume you want to install the <package> slice. Follow this guide
+and replace any mention of <package> with your desired sliced program.
```
medubelko: Recommended:

I have some thoughts about this in view of all the removals.

These two paragraphs have ideas that are joined at the hip. I think the beginning will go a lot smoother than before.

Since this whole guide depends on the user already having made a slice, I would say it's critical to keep a link to https://ubuntu.com/chisel/docs/latest/how-to/slice-a-package, in text (an unusual request from me!). That way, if they haven't, they know which guide to follow first.

I'd much rather we don't use `<package>` for the main subject of a guide. The purpose of angle bracket text is for disposable placeholders in code or templates. In the main text itself, it has the effect of making the document look incomplete. We could continue with the example of vim-tiny, and even notify the user that we're continuing where we left off in the Chisel guide.

docs/how-to/chiselling/install-slice.rst
```diff
@@ -25,17 +25,17 @@ Let's assume you want to install the OpenSSL binaries slice created in
     :dedent: 2
 
 This repository acts as the database of slice definitions files for each
-Chisel release (Chisel releases are named analogously to Ubuntu releases, and
-mapped into Git branches within the repository).
+Chisel release. Chisel releases are named analogously to Ubuntu releases, and
+mapped into Git branches within the repository.
```
medubelko: Recommended:

When initially developing a local slice, the user will already have cloned chisel-releases, so I suggest removing lines 19 to 29.

docs/how-to/chiselling/create-slice.rst
medubelko, line 1: Required:

All removed pages should be rerouted in `redirects.txt`.

docs/how-to/code/install-slice/task.yaml
```diff
@@ -12,23 +12,23 @@ execute: |
   # chisel snap to be able to access them.
   mv rockcraft.yaml "${HOME}/rockcraft.yaml.backup"
   pushd $HOME
-  rm -fr chisel-releases/ my-custom-openssl-fs/
+  rm -fr chisel-releases/ my-custom-<package>-fs/
```
medubelko: As per the Copilot review, this code won't run properly in Bash or Shell.

This is a tricky situation. If we want to test this properly, we'd need an SDF in the doc assets that Spread can read to run the Chisel commands.

Was the test working prior to this?

medubelko (changes requested):

medubelko:
@asanvaq After further conversation with @tigarmo, it probably makes the most sense to remove the Spread test from _Install a custom package slice_. Chisel is causing intermittent CI failures in the test, and we're only testing a handful of commands, anyway.

asanvaq:
Okay! Thanks for the feedback, I will look into it.

medubelko:
@asanvaq Do you think we'll land this before the end of the week?

docs/how-to/chiselling/install-slice.rst
asanvaq: Done, I edited the text that follows slightly.

docs/how-to/chiselling/install-slice.rst
asanvaq: Done!

docs/how-to/chiselling/create-slice.rst
asanvaq: Done!

asanvaq:
@medubelko I applied the feedback and pushed the changes!

asanvaq:
Docs are building locally, unsure on why the RTD build is failing here. This is also the case. It passes the lint tests locally.

medubelko (approved):

tigarmo (approved):

PR #1270, opened 2026-06-25
fix: allow downgrades in overlay apt upgrade for FIPS builds

ktsakalozos-canonical:
## Summary

- Add `--allow-downgrades` to `apt-get -y upgrade` in the overlay step when Pro/FIPS sources are detected
- Fixes FIPS rock builds failing because the `fips-updates` repository pins `libgcrypt20` to a version older than the noble GA archive, causing apt to refuse the downgrade without `--allow-downgrades`

## Problem

When building FIPS rocks on noble, the `esm.ubuntu.com/fips-updates` repository pins `libgcrypt20` to the FIPS-certified version which is older than the GA version in the standard noble archive. During the overlay `apt-get -y upgrade` step, apt detects this as a downgrade and aborts with:
```
E: Packages were downgraded and -y was used without --allow-downgrades.
```

This started affecting builds recently (e.g. https://github.com/canonical/cilium-rocks/actions/runs/28017207023/job/83118786893).

## Fix

The `--allow-downgrades` flag is safe here because:

1. This code path only executes when Pro/FIPS sources are present (guarded by the `find` check for `ubuntu-esm-*.sources` or `ubuntu-fips-*.sources`)
2. The downgrade is intentional - the FIPS pin sets the preferred version and apt should honor it
3. No other packages are affected since only pinned packages will be downgraded

It is also future safe because it does not break existing build pipelines and allows the rockcraft team to fix future occurances of the issue when on their own pace.

### DIFF
--- a/rockcraft/services/project.py
+++ b/rockcraft/services/project.py
@@ -39,7 +39,7 @@ APT_UPGRADE_PART = {
         if find /etc/apt/sources.list.d/ -iname "ubuntu-esm-*.sources" -or -iname "ubuntu-fips-*.sources" | grep -q "."; then
             echo "Upgrading base overlay system packages"
             apt-get update
-            apt-get -y upgrade
+            apt-get -y --allow-downgrades upgrade
             apt-get clean
             rm -rf /var/cache/apt
         else
--- a/tests/spread/pro/fips/rockcraft.yaml
+++ b/tests/spread/pro/fips/rockcraft.yaml
@@ -1,0 +1,14 @@
+name: fips-test
+summary: Rock to test FIPS Pro overlay upgrade.
+description: Test that --pro=fips-updates packs a rock with a full base.
+version: "latest"
+license: Apache-2.0
+
+base: ubuntu@24.04
+
+platforms:
+  amd64:
+
+parts:
+  my-part:
+    plugin: nil
--- a/tests/spread/pro/fips/task.yaml
+++ b/tests/spread/pro/fips/task.yaml
@@ -1,0 +1,19 @@
+summary: |
+  Test that packing a rock with FIPS Pro sources succeeds.
+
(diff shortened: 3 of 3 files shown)


tigarmo:
Hi @ktsakalozos-canonical , thanks a lot for the report and fix. I agree with your assessment of the risk here and in fact that seems like the correct behavior in general.

I do have two requests for this PR, though:

1) Can you re-sit the commit on `hotfix/1.19`? That's the branch for bugfixes on 1.19, `main` has a whole bunch of other stuff that we can't really publish on a patch release.
2) Can you update the base on [this fips spread test](https://github.com/canonical/rockcraft/blob/main/tests/spread/pro/fips-pebble/rockcraft.yaml#L7) from

```yaml
base: bare
build-base: ubuntu@22.04
```

to

```yaml
# Use a full base to catch potential issues when upgrading 
# the base overlay system (e.g. fips downgrades)
base: ubuntu@24.04
```

(with no build-base). I confirmed that this change makes the test fail, so this is a good double-check that the fix is good.

ktsakalozos-canonical:
@tigarmo I am slightly confused. The [fips spread test ](https://github.com/canonical/rockcraft/blob/main/tests/spread/pro/fips-pebble/rockcraft.yaml#L7)does not exist in the hotfix/1.19.

tigarmo:
Oh you're right! That test got added a few weeks ago for [something else](https://github.com/canonical/rockcraft/commit/f6bebb5cc72dcce5700900bcdd337a51a77f7167). What a coincidence

In this case, I think we need a new test, that just checks that packing with fips works. It can be called `fips`.

ktsakalozos-canonical:
Added what should be a minimal test. Should fail the packing on noble without the suggested patch.

tigarmo:
This fix was merged on #1272 . Thanks!

PR #1289, opened 2026-07-07
docs: make tutorials architecture-agnostic

### LINKED ISSUE #1280: Docs: Tutorials assume an `amd64` host which causes `platforms:` and `.rock` filenames break for `arm64` readers
### Check existing issues

- [x] I've verified that this request isn't described by any existing issues.

## Summary

Several Rockcraft tutorials hardcode `amd64` in the `rockcraft.yaml` `platforms:` block and in the resulting `.rock` filenames used by later commands. A reader on an `arm64` host (e.g. Apple Silicon macOS, which is the majority of current Macs) packs a `..._arm64.rock` file, so the copy-paste `skopeo` import commands which reference `..._amd64.rock`, and this fails with a file-not-found error.

### Environment
- Host architecture: `arm64` (Apple Silicon)
- Build VM: Multipass (Ubuntu 24.04)
- Docs version: `1.19`
 
### Reproduction:
Using "Pack a Python package" as the clearest example:
 
1. In "Setup your environment," select the **macOS** tab (as an Apple Silicon user would).
2. Follow through to the `rockcraft.yaml` snippet:
```yaml
   platforms:
     arm64:
```
3. Run `rockcraft pack` on the `arm64` host/VM.
4. Observe the produced file is `pyfiglet_0.7.6_arm64.rock`.
5. Run the import command exactly as written:
```
   sudo rockcraft.skopeo --insecure-policy copy oci-archive:pyfiglet_0.7.6_amd64.rock docker-daemon:pyfiglet:0.7.6
```
   → fails, because the file on disk is `..._arm64.rock`.
 
The Hello World and Chisel tutorials reproduce the same way with their respective filenames.

## Suggested fix

Apply one consistent approach across the affected tutorials, changing from 
```
sudo rockcraft.skopeo --insecure-policy copy oci-archive:hello_latest_amd64.rock docker-daemon:hello:latest
```
to something that use $(dpkg --print-architecture) like
```
sudo rockcraft.skopeo copy \
  --insecure-policy \
  oci-archive:hello_latest_$(dpkg --print-architecture).rock \
  docker-daemon:hello:latest
```

### Document location

https://documentation.ubuntu.com/rockcraft/1.19/tutorial/hello-world/

minulo:
#### Summary

Several tutorials hardcoded `amd64` in the `platforms:` block and in the
resulting `.rock` filenames used by later commands. A reader on an arm64 host
packs a `..._arm64.rock` file, so the copy-paste `skopeo` import commands that
reference `..._amd64.rock` fail with a file-not-found error.

This applies the approach already used by the Flask tutorial across the
remaining affected tutorials.

#### Changes

Tutorials updated: **Hello World**, **Pack a Python package** (pyfiglet), and
**Bundle a Node.js app**. For each:

- `rockcraft.yaml`: commented-out extra architectures + a hint to check the
  host arch with `dpkg --print-architecture`.
- `task.yaml`: hardcoded `..._amd64.rock` replaced with
  `..._$(dpkg --print-architecture).rock`; `skopeo copy` reformatted to
  multi-line.
- `.rst`: added a note after `rockcraft pack` clarifying the `.rock` filename
  depends on name, version, and platform.

#### Testing

- `make docs-lint` — passes (0 errors/warnings, linkcheck clean)
- `make test-fast` — 620 passed, 1 skipped

The `task.yaml` files run in CI as Spread tests; the `dpkg` substitution keeps
them passing on any host architecture.

Fixes #1280

---

- [ x] I've followed the [contribution guidelines](https://github.com/canonical/rockcraft/blob/main/CONTRIBUTING.md).
- [ x] I've signed the [CLA](http://www.ubuntu.com/legal/contributors/).
- [ x] I've successfully run `make lint && make test-fast`.
- [x ] I've added or updated any relevant documentation.
- [ x] In documents I changed, I [added a meta description](https://canonical-starflow.readthedocs-hosted.com/how-to/add-a-page-meta-description/) if one was missing.


minulo:
The Security scan failure is unrelated to this docs-only change. I can rebase once that dependency update lands on main.

docs/tutorial/node-app.rst
```diff
@@ -1,3 +1,6 @@
+.. meta::
+    :description: Learn to package apps with Rockcraft. In this tutorial, we bundle a typical Node.js application into an OCI-compliant container and run it with Docker.
```
asanvaq, line 2: Thanks for adding this!

docs/tutorial/code/hello-world/rockcraft.yaml
```diff
@@ -7,8 +7,13 @@ version: "latest"
 license: Apache-2.0
 
 base: ubuntu@26.04
+# the platforms this rock should be built on and run on.
+# you can check your architecture with `dpkg --print-architecture`
 platforms:
   amd64: # Make sure this value matches your computer's architecture
```
asanvaq:
```suggestion
  amd64:
```

asanvaq: If we keep this, we are duplicating the text above.

asanvaq:
Please include this snippet in the tutorials right before the "pack" step. It should be included in the following places based on the tutorial:
- Hello world: line 31, after "describe the rock."
- Node app: after line 47
- Pypi: after line 52.

```
The ``platforms`` key must match the architecture of your host.
Edit the ``platforms`` key in ``rockcraft.yaml`` if required.

.. note::
    For this tutorial, we'll use the ``name`` ``<name of rock in each tutorial>`` and assume
    we're running on the  ``amd64`` platform. Check the architecture of the
    system using ``dpkg --print-architecture``.

    The ``name``, ``version`` and ``platform`` all influence the name of the
    generated ``.rock`` file.

File: docs/tutorial/hello-world.rst [hunk 3]
```diff
@@ -46,6 +46,10 @@ At the end of the process, a file named ``hello_latest_amd64.rock`` should be
 present in the current directory. That's your rock, in oci-archive format
 (a tarball).
```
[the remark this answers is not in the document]
asanvaq: Not applicable, we are specifying we are using amd64 in this tutorial.

asanvaq (changes requested):
Apologies for the delay! We have been thinking on how to best solve this problem once more. There are some small changes in the comments.
I also proposed adding an extra note before the packing step, similar to the Flask tutorial. This is outside the scope of what we agreed on during the initial issue discussion, so if you want to just improve the work that is in scope of what was originally agreed, and leave out this addition, please let me know and we can discuss further.

docs/tutorial/node-app.rst
asanvaq: This needs to be updated so all the other architectures are also displayed,

asanvaq:
Removed @erinecon as this PR is not touching 12-factor content.

minulo:
> Apologies for the delay! We have been thinking on how to best solve this problem once more. There are some small changes in the comments. I also proposed adding an extra note before the packing step, similar to the Flask tutorial. This is outside the scope of what we agreed on during the initial issue discussion, so if you want to just improve the work that is in scope of what was originally agreed, and leave out this addition, please let me know and we can discuss further.

No worries! I have implemented all the suggestions (even the out-of-scope). Current lint check fails because they can't reach packages.ubuntu.com, I will push another change at a later date to check for connection.

asanvaq (approved):
Looks good, thanks!

docs/tutorial/hello-world.rst
```diff
@@ -31,6 +31,17 @@ This file instructs Rockcraft to build a rock that **only** has the ``hello`` pa
...
+The ``platforms`` key must match the architecture of your host.
+Edit the ``platforms`` key in ``rockcraft.yaml`` if required.
+
+.. note::
+    For this tutorial, we'll use the ``name`` ``hello`` and assume
+    we're running on the  ``amd64`` platform. Check the architecture of the
+    system using ``dpkg --print-architecture``.
+
+    The ``name``, ``version`` and ``platform`` all influence the name of the
+    generated ``.rock`` file.
```
medubelko: @asanvaq @minulo Notes are for optional but nice-to-know information. Should this be a note, or an essential part of the tutorial?

docs/tutorial/hello-world.rst
```diff
@@ -46,6 +57,10 @@ At the end of the process, a file named ``hello_latest_amd64.rock`` should be
 present in the current directory. That's your rock, in oci-archive format
 (a tarball).
 
+.. note::
```
medubelko: Same question about notes here.

docs/tutorial/node-app.rst
medubelko:
```suggestion
    :description: Learn to package whole apps with Rockcraft. In this tutorial, we bundle a Node.js app into an OCI-compliant container.
```

I wouldn't include a competitor's product in the description.

docs/tutorial/code/node-app/task.yaml
```diff
@@ -10,14 +10,17 @@ execute: |
   # [docs:build-rock-end]
 
   # [docs:check-rock]
-  ls my-node-app_1.0_amd64.rock
+  ls my-node-app_1.0_$(dpkg --print-architecture).rock
```
medubelko, line 13: I have an alternative take. IMO, this serves the integrity of the tutorial's tests more than the experience of reading the tutorial.

The reader is always going to be able to tab-autocomplete this path, and won't be giving a second thought to the arch string in the file, which actually undermines our earlier goal of making them aware of the architecture. By asking them to type (or copy) this out, it teaches an anti-pattern they'll never use in regular work, and lets them ignore this subtlety.

For the code we actually show on the page, this would serve better:

```bash
ls my-node-app-1.0_<architecture>.rock
```

This teaches the reader that the arch string will vary, and points it out in a way that they _can't_ simply skip by copy-pasting. But, if they know auto-complete, they can easily fill without slowing down their keypresses.

minulo: Yes, it is to preserve the integrity of the tests. The reason why we can't `ls my-node-app-1.0_<architecture>.rock` is because the test cases in task.yaml requires all CLI commands to be able to be run by the users directly. Hence it is either specifically and/arm or `$(dpkg --print-architecture)`. 

Also, from my opinion, the tutorial is the entry points for new users. Personally, I believe it is best to remove as much friction as possible. Also, I believe that not many people would type it by hand and most likely will do a click copy button into ctrl+v combo. Also, part of why I think  `$(dpkg --print-architecture)` is okay is because other tutorials use the same pattern.

docs/tutorial/hello-world.rst
minulo: Since the must-do step ("the platforms key must match your host") is already in the text above, and the later commands use $(dpkg --print-architecture), I think what's left in these notes is nice-to-know as a reader who skips them can still completes the tutorial. 

The wording also matches the Flask/Django/Go tutorials, as @asanvaq suggested. This is probably a maintainer opinion piece. If we do change it, it probably wants a follow-up across all tutorials for consistency.

erinecon: IMO we should remove this note everywhere, including the Flask/Django/Go tutorials. I've received feedback in UX research sessions that participants treat the note as duplicate or redundant information. In addition, since the tutorials are architecture-agnostic, we're not assuming any architecture. So `...assume we're running on amd64` is not strictly true.

asanvaq: Thanks for the feedback @medubelko @erinecon. It makes sense to remove the note in all the tutorials then. The last part is important for users to know, so I propose leaving it as part of the text.

 ```suggestion
    The ``name``, ``version`` and ``platform`` all influence the name of the
    generated ``.rock`` file.

docs/tutorial/hello-world.rst
asanvaq: As we already specify this above, I propose dropping this note @minulo.

docs/tutorial/hello-world.rst
minulo: Thanks for the feedback as well. That makes sense.

docs/tutorial/hello-world.rst
minulo: Will do.

docs/tutorial/hello-world.rst
minulo: I will create another PR for Flask/Django/Go tutorials, I think this PR will be scoped specifically to the Hello World, Pack a Python package (pyfiglet), and Bundle a Node.js app tutorials.

asanvaq: @minulo could you please wait until we have agreed on this: https://github.com/canonical/rockcraft/pull/1289#discussion_r4064925177 ? I will get back to you as soon as possible.

minulo: Yep, I will wait.  The fix for notes is in. Please take your time on the <architecture> discussion.

PR #1317, opened 2026-08-03
fix: set sticky bit on the default directory

### LINKED ISSUE #1170: missing sticky bit in PEBBLE_PATH
### Check existing issues

- [x] I've verified that this bug isn't described by any existing issues.

### Bug description

May not be a bug, but definitely a security-related consideration that may need to be revisited, to ensure it was either done on purpose, and if so, to document why, or if not, to fix it.

The original thread comes from https://github.com/dotnet/dotnet-docker/issues/7054.

Once addressed, please provide feedback in that thread.

### Steps to reproduce

`rockcraft pack` any rock

### Environment

Should be the case on any rockcraft version and underlying environment, since the culprit is the implicit Pebble part.

### rockcraft.yaml

```yaml

```

### Log output

```shell

```

HeRaNO:
Describe your changes.

---

Fixes #1170.

`/var/lib/pebble/default` is intentionally world-writable so that Pebble can create its socket and state files when a rock uses a non-root `run-user`. However, mode `0777` allows users to remove or rename entries owned by other users and is reported by insecure world-writable-directory checks.

Set the directory mode to `01777`, preserving non-root write access while protecting directory entries with the sticky bit.

The mode is updated in both paths that create the directory:

- the implicit Pebble part, including rocks without services or checks;
- the layer generated when services or checks are present.

Tests now verify the complete mode with `stat.S_IMODE` instead of inspecting only the final three octal digits. Spread coverage checks the resulting image both with and without a generated Pebble layer, while retaining the existing non-root file-creation check.

After this is released, affected base images must be rebuilt and republished; the change cannot alter permissions in already-published OCI layers.


tigarmo (approved):
Thanks a lot! I looked at the spread tests and confirm that the failures are unrelated (the new spread test changes passed)

bepri (approved):

PR #1322, opened 2026-08-13
style: resolve linting issues

lengau:
This PR fixes several linting issues including ISC004, INP001, and PTH118/120.

### DIFF
--- a/rockcraft/extensions/springboot.py
+++ b/rockcraft/extensions/springboot.py
@@ -223,8 +223,10 @@ class SpringBootFramework(Extension):
     # discovered at build time from build/libs/ (bootJar's output directory).
     GRADLE_OVERRIDE_BUILD_COMMANDS = [
         "craftctl default",
-        "SPRING_FAT_JAR=$(find ${CRAFT_PART_BUILD}/build/libs"
-        " -name '*.jar' ! -name '*-plain.jar' -type f -printf '%f\\n' | head -1)",
+        (
+            "SPRING_FAT_JAR=$(find ${CRAFT_PART_BUILD}/build/libs"
+            " -name '*.jar' ! -name '*-plain.jar' -type f -printf '%f\\n' | head -1)"
+        ),
         '[ -n "${SPRING_FAT_JAR}" ] || (echo "ERROR: could not find Spring Boot fat JAR in build/libs" && exit 1)',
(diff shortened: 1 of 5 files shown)


rockcraft/extensions/springboot.py
```diff
@@ -223,7 +223,7 @@ def _user_install_app_build_packages_override(self) -> list[str]:
     # discovered at build time from build/libs/ (bootJar's output directory).
     GRADLE_OVERRIDE_BUILD_COMMANDS = [
         "craftctl default",
-        "SPRING_FAT_JAR=$(find ${CRAFT_PART_BUILD}/build/libs"
+        "SPRING_FAT_JAR=$(find ${CRAFT_PART_BUILD}/build/libs",
```
lengau: This is incorrect. This line and the next are supposed to do string concatenation.

rockcraft/plugins/java_common.py
```diff
@@ -24,7 +24,7 @@ def create_usr_bin_java() -> list[str]:
     return [
         '# Find the "java" executable and make a link to it in $CRAFT_PART_INSTALL/bin/java',
         "java_bin=$(find ${CRAFT_PART_INSTALL} -name java -type f -executable)",
-        "test -d ${CRAFT_PART_INSTALL}/usr/bin && "
-        "(test -f ${CRAFT_PART_INSTALL}/usr/bin/java || "
+        "test -d ${CRAFT_PART_INSTALL}/usr/bin && ",
```
lengau: This is also supposed to be string concatenation.

snap/local/__init__.py
lengau, line 1: We shouldn't be creating these extra files.

tests/unit/test_pebble.py
```diff
@@ -81,20 +81,7 @@ def test_attributes(self):
...
-                    f"{os.linesep}"
-                    "  mockServiceOne:"
-                    f"{os.linesep}"
-                    "    override: replace"
-                    f"{os.linesep}"
-                    "    command: foo"
-                    f"{os.linesep}"
-                    "    on-success: shutdown"
-                    f"{os.linesep}"
+                    f"summary: mock summary{os.linesep}description: mock description{os.linesep}services:{os.linesep}  mockServiceOne:{os.linesep}    override: replace{os.linesep}    command: foo{os.linesep}    on-success: shutdown{os.linesep}",
```
lengau: @copilot undo this change

PR #1366, opened 2026-09-16
build(deps): bump craft-application to 7.2.1

zhijie-yang:
## Summary

- require craft-application 7.2.1 or newer
- update the lockfile to use craft-application 7.2.1
- receive the armv8l architecture fix

## Testing

- [x] uv lock --check
- [x] git diff --check
- [x] verified the frozen environment installs craft-application 7.2.1


tigarmo (approved):
thanks!

